PR #24, opened 2025-04-09
feat: run the code in daytona

idagelic:
This PR adds a prompt for running the code that the coder generated - the code is run in a secure and isolated Daytona Sandbox that spins up in a fraction of a second and runs the arbitrary Python/TypeScript code

How to use it:

1. First terminal window:
`export GEMINI_API_KEY=<your-gemini-api-key>`
`npm run agents:coder`

2. Second terminal window:
`export DAYTONA_API_KEY=<your-daytona-api-key`
`npm run a2a:cli:daytona`

image

This is a simple addition to the Coder sample as a proof of work implementation of an AI runtime. We are working on an A2A implementation with our sandbox orchestration agent that will be released soon. Interested in the feedback from the team.

### DIFF
--- a/samples/js/package-lock.json
+++ b/samples/js/package-lock.json
@@ -8,6 +8,7 @@
       "name": "a2a-samples-js",
       "version": "0.1.0",
       "dependencies": {
+        "@daytonaio/sdk": "^0.12.1",
         "@genkit-ai/googleai": "^1.4.0",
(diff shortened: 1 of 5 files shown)


zeroasterisk:
I believe this PR is now deprecated by #170 / reopen if untrue

PR #26, opened 2025-04-09
fix: update directory path in README for clarity

ekadetov:


kthota-g:
The readme has been fixed with the right path in main.

PR #32, opened 2025-04-09
docs: Update README for Google ADK agent to correct run command path

syllogismos:
from the `samples/python` directory, the right uv command is `uv run agents/google_adk` for it to run the `__main__.py` file in `google_adk` directory


kthota-g:
These changes are also addressed part of https://github.com/google/A2A/pull/79

PR #39, opened 2025-04-10
Instructions for multiagent to work with GOOGLE_API_KEY, for the demo UI to run as expected

syllogismos:
- Added instructions for creating a .env file with GOOGLE_API_KEY and why its needed in the hosts README.
- Updated host_agent.py to check for the GOOGLE_API_KEY environment variable and raise an error if not set.


kthota-g:
GOOGLE_API_KEY check is added to all sample agent/client invocations part of https://github.com/google/A2A/pull/79. Closing this PR.

PR #41, opened 2025-04-10
load environment variables in CrewAI main module

syllogismos:
Even though the env variables are loded in init of `ImageGenerationAgent` class, which is imported by this `__main__.py`, I'm getting `MissingAPIKeyError`

I had to load them in the `__main__.py` to get image generation working with both cli host and the web host.


kthota-g:
This has been addressed as part of https://github.com/google/A2A/pull/79

PR #46, opened 2025-04-10
Fix Env problem and update minor docs 

Saedbhati:


kthota-g:
This issue has been addressed via https://github.com/google/A2A/pull/79

PR #52, opened 2025-04-10
feature: Golang sample implementation

AlexsJones:
Okay, I got very caffeinated and decided we need something in Golang, because this was the only way I could wrap my head around this. The PR introduces a Go implementation of the Agent-to-Agent (A2A) protocol, focusing on the core server and client components. It does not yet cover additional golang agents ( I'll do that in another PR if you accept this ).

CLA signed... 

### Server Implementation
Implemented a JSON-RPC 2.0 server/client
Added support for core A2A methods:
tasks/send: Send new tasks
tasks/get: Retrieve task status
tasks/cancel: Cancel running tasks
Implemented thread-safe task storage
Error handling with A2A error codes etc
Created extensive test coverage for server functionality
### Client Implementation
 support for all core A2A methods
Created test suite for client (Thanks Copilot 👍)
### Models Package
Defined type-safe structures for:
Task management
Error handling
Implemented JSON serialization/deserialization
Added validation for all data structures
Added streaming support

### Documentation
Created README files for:
Server package
Client package
Models package
Added usage examples and API documentation
Included testing instructions

## Pending Work
Agent implementation is not included in this PR and will be addressed in a future update
The current implementation focuses on the core protocol and infrastructure
Agent-specific functionality (like skill execution, capability negotiation, etc.) will be implemented separately.


samples/go/go.mod
```diff
@@ -0,0 +1,3 @@
+module a2a
```
AlexsJones, line 1: Could change this if you want to the FQDN rather than relative, LMK

bwplotka (approved):
Nice! Looks clean and solid to me!

Would be nice to have a Go example for A2A protocol 💪🏽 

I don't maintain this repo, but I will do a friendly ping internally for feedback 🤗

samples/go/server/server.go
```diff
@@ -0,0 +1,205 @@
...
+		basePath:    basePath,
+		taskStore:   make(map[string]*models.Task),
+		taskHistory: make(map[string][]*models.Message),
+	}
+}
+
+// Start starts the A2A server
+func (s *A2AServer) Start() error {
+	mux := http.NewServeMux()
+	mux.HandleFunc(s.basePath, s.handleRequest)
```
tucnak, line 39: This is an antipattern: basically, if want to use any other HTTP server, I would have a really hard time because you made the handler private...

samples/go/server/server.go
```diff
@@ -0,0 +1,205 @@
...
+	return http.ListenAndServe(fmt.Sprintf(":%d", s.port), mux)
+}
+
+// handleRequest handles incoming HTTP requests
+func (s *A2AServer) handleRequest(w http.ResponseWriter, r *http.Request) {
```
tucnak, line 44: This should just be renamed to [Handler](https://pkg.go.dev/net/http#Handler) and problem solved. I haven't looked through much else of the code, but I reckon similar adjustments are necessary.

tucnak: This is an antipattern: basically, if want to use any other HTTP server, I would have a really hard time because you made the handler private...

Simply rename `handleRequest` to `ServeHTTP` so it satisfies [Handler](https://pkg.go.dev/net/http#Handler) and you'll be fine.

samples/go/server/server.go
```diff
@@ -0,0 +1,205 @@
...
+		handler:     handler,
+		port:        port,
+		basePath:    basePath,
+		taskStore:   make(map[string]*models.Task),
+		taskHistory: make(map[string][]*models.Message),
+	}
+}
+
+// Start starts the A2A server
+func (s *A2AServer) Start() error {
```
tucnak, line 36: There's no need for this; we're all adults.

samples/go/server/server.go
tucnak: Oh, I mean, there's different transports so you migth not need a rename, but a whole new function after all...

AlexsJones: Yes you're right, updated

zeroasterisk:
@whitlockjc would you be so kind as to push this one through.  I would like to land this go example.

AlexsJones:
Just holla if I need to change anything

whitlockjc:
I'm on it...started a few times but deadlines have distracted me. 😆

samples/go/client/client.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// Client represents an A2A protocol client
+type Client struct {
+	baseURL    string
+	httpClient *http.Client
+}
+
+// NewClient creates a new A2A client
+func NewClient(baseURL string) *Client {
```
whitlockjc, line 21: This is more of an enhancement, and not a requirement, but it would be nice to have a `NewClientFromAgentCard` or something of the like, where you have an `AgentCard` created/pulled from a remote, and use the URL from within it.

Again, not a blocker but something to consider.

samples/go/client/client.go
```diff
@@ -0,0 +1,146 @@
...
+	return &Client{
+		baseURL: baseURL,
+		httpClient: &http.Client{
+			Timeout: 30 * time.Second,
+		},
+	}
+}
+
+// SendTask sends a task message to the agent
```
whitlockjc, line 30: I don't see streaming support, is this something you plan on contributing in the future? Here is our Python reference implementation equivalent: https://github.com/google/A2A/blob/main/samples/python/common/client/client.py#L38

samples/go/models/jsonrpc.go
```diff
@@ -0,0 +1,46 @@
...
+	JSONRPC string `json:"jsonrpc,omitempty"`
+}
+
+// JSONRPCRequest represents a JSON-RPC request object base structure
+type JSONRPCRequest struct {
+	JSONRPCMessage
+	// Method is the name of the method to be invoked
+	Method string `json:"method"`
+	// Params are the parameters for the method
+	Params interface{} `json:"params,omitempty"`
```
whitlockjc, line 24: Excellent catch here, the JSONRPC spec allows `params` to be both a `slice` and a `map`.

samples/go/models/requests.go
```diff
@@ -0,0 +1,99 @@
+package models
```
whitlockjc, line 1: Another streaming-related feedback, [SendTaskStreamingRequest](https://github.com/google/A2A/blob/main/specification/json/a2a.json#L1343) is missing.

samples/go/server/README.md
```diff
@@ -0,0 +1,83 @@
+# A2A Server (Go)
+
+This directory contains a Go server implementation for the Agent-to-Agent (A2A) communication protocol.
+
+## Features
+
+- JSON-RPC 2.0 compliant server
+- Supports core A2A methods:
+  - `tasks/send`: Send a new task
```
whitlockjc, line 9: Missing streaming.

samples/go/server/server.go
```diff
@@ -0,0 +1,205 @@
...
+	handler     TaskHandler
+	port        int
+	basePath    string
+	taskStore   map[string]*models.Task
+	taskHistory map[string][]*models.Message
+	mu          sync.RWMutex
+}
+
+// NewA2AServer creates a new A2A server instance
+func NewA2AServer(handler TaskHandler, port int, basePath string) *A2AServer {
```
whitlockjc: Simile to `samples/go/client/client.go`, it would be nice to have a helper method for instantiating an `A2AServer` with an `AgentCard`.

samples/go/server/server.go
```diff
@@ -0,0 +1,205 @@
...
+		if id, ok := request.ID.(string); ok {
+			idStr = id
+		} else {
+			s.sendError(w, "", models.ErrorCodeInvalidRequest, "Invalid request ID format")
+			return
+		}
+	}
+
+	switch request.Method {
+	case "tasks/send":
```
whitlockjc: Streaming support missing.

whitlockjc (changes requested):
Streaming support seems to be missing, and while I think we could handle this in a follow-up, I think there is enough interest in this to ensure it's as complete as possible off the rip. Thoughts?

samples/go/client/client.go
whitlockjc: By the way, I don't care as much about naming so much as I do the ability to create a new A2A client/server manually _(no `AgentCard`)_ and via the `AgentCard`.

AlexsJones:
> Thank you for your contribution! The major concern right now is that streaming support seems to be missing, and while I think we could handle this in a follow-up, I think there is enough interest in this to ensure it's as complete as possible off the rip. Thoughts?

I had thought about breaking it up, but I'll see what I can get done tonight/tomorrow and update the PR

samples/go/client/client.go
AlexsJones: Done

AlexsJones:
Okay, added streaming, had a bit of a refactor to avoid a double closure issue.

whitlockjc:
I'm on it.

AlexsJones:
What do you think @whitlockjc ? I have some time if changes are needed mid week!

rwrz:
Have you checked this implementation from @trpc-group?
https://github.com/trpc-group/trpc-a2a-go

It can be helpful (or a distraction right now).

whitlockjc (approved):
👍

zeroasterisk:
LGTM.

Can we either cleanup to pass the (new) spelling and linting CI/CD checks or recommend changes to the checks?

AlexsJones:
Hi, anything you need me to do @zeroasterisk @whitlockjc  ?

holtskinner:

PR #54, opened 2025-04-10
Update README.md

mandle420:
fixed the demo video link that wasn't showing up in my browser. (either librewolf or brave) 
While I believe the author wanted the video showing up on the page, either the code was wrong, my browser couldn't parse the link, or maybe not supported on git? 
Either way, it works as a link, and that's about all the coding I want to do. Feel free to fix if thats not what you want.


rajeshvelicheti (commented):
Looks like this has already been addressed. Thanks.

PR #58, opened 2025-04-10
fix: correct the path in theREADME

z2z23n0:
Fixed the path to the command line in README.md of demo google_adk to prevent users from copying commands directly and finding errors, so that the command form of demo google_adk is the same as that of langgraph.


kthota-g:
Thanks for the PR. We have addressed this issue as part of https://github.com/google/A2A/pull/79

PR #80, opened 2025-04-10
fix(format): format demo and sample code with ruff

imfing:
The current codebase includes inconsistent formatting and many unused imports.

This PR applies [ruff](https://docs.astral.sh/ruff/), which is listed in the [dev dependencies](https://github.com/google/A2A/blob/251cd37d5e6ccbc985bb8d0d78c03d665fa86699/demo/ui/pyproject.toml#L34) to:
- Format the Python code
- Sort and fix import statements
- Remove unused imports

Note: the code style expectations are not clearly defined in the contributing guide. It’s unclear whether the project intends to follow [Google Python Style](https://google.github.io/styleguide/pyguide.html), use [yapf](https://github.com/google/yapf), or rely on ruff.
Since ruff is already present in the pyproject.toml, I’ve assumed it’s intended to be used for formatting.

If that’s not the case, feel free to close this PR or suggest the preferred formatting tool.


ghost:
assets/css/compiled/main.css

PR #90, opened 2025-04-11
fix: web UI build fails on 'uv run main.py' 

elevenarms:
The "Demo Web App" client for the Agent sample fails to build and run using 'uv run main.py' - the hatch build system expects the README.md to be in the same directory as the project. For the sake of cleanliness I've also flattened the directory structure to remove the UI folder. Moving the README.md fixes the build and the app runs great. I also made a quick GIF of the UX and added to the README.md

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -12,7 +12,7 @@ This demo application showcases agents talking to other agents over A2A.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


holtskinner:

PR #93, opened 2025-04-11
feat(langgraph): Add all major LLM providers supported by langchain

HomeroRR:
Support added in the langgraph python example for the following:

langchain-google-genai>=2.0.10
langchain-openai>=0.3.12
langchain-anthropic>=0.3.10
langchain-groq>=0.3.2
langchain-cohere>=0.4.3
langchain-nvidia-ai-endpoints>=0.3.9
langchain-fireworks>=0.2.9
langchain-mistralai>=0.2.10
langchain-together>=0.3.0
langchain-ibm>=0.3.10
databricks-langchain>=0.4.2
langchain-xai>=0.2.2
langchain-perplexity>=0.1.0


samples/python/agents/langgraph/__main__.py
```diff
@@ -19,8 +19,23 @@
 def main(host, port):
     """Starts the Currency Agent server."""
     try:
-        if not os.getenv("GOOGLE_API_KEY"):
-            raise MissingAPIKeyError("GOOGLE_API_KEY environment variable not set.")
+        if (
+            not os.getenv("GOOGLE_API_KEY")
```
princejha95: Can't we do it like below ?

`def check_environment_variables_any(variable_names):
  if not any(os.getenv(var) is not None for var in variable_names):
    raise APIKeyError(f"None of the required environment variables are set: {', '.join(variable_names)}")` 

variable_names is a set of environment variables supported by the protocol

samples/python/agents/langgraph/agent.py
```diff
@@ -60,9 +73,48 @@ class CurrencyAgent:
         "Set response status to error if there is an error while processing the request."
         "Set response status to completed if the request is complete."
     )
-     
+
     def __init__(self):
-        self.model = ChatGoogleGenerativeAI(model="gemini-2.0-flash")
+        if os.getenv("GOOGLE_API_KEY"):
```
princejha95: Let's use switch case instead of multiple if-else statements ?

HomeroRR: Done. 
Python>=3.13 is a prerequisite so we can assume switch statements are available.

samples/python/agents/langgraph/__main__.py
HomeroRR: I like your idea.
I have changed the code to use the SUPPORTED_API_KEYS set.

darkhaniop:
I have a question about this patch. In this implementation, `get_api_key()` loops through known API keys in a predetermined order. Which means, if we have more than one key in our env, we cannot make the LangGraph agent pick the one that comes later in the `SUPPORTED_API_KEYS` list.

Then, let's say someone tries to launch multiple agent servers as described in [README](https://github.com/google/A2A/blob/8c8e0eabfe0558159044455e917b38c7472a9fe1/samples/python/README.md), `ReimburementAgent` (setting `GOOGLE_API_KEY`) and the LangGraph-based one (e.g., with `OPENAI_API_KEY`):

1. `cd samples/python` (where they have their `.env`).
2. `uv run agents/google_adk` *and* `uv run agents/langgraph`

Both of the agents would still use `GOOGLE_API_KEY`.

Maybe if the script has access to more than one key, it can be updated to require specifying the provider to use? Or provide other ways of choosing the provider?

princejha95:
> I have a question about this patch. In this implementation, `get_api_key()` loops through known API keys in a predetermined order. Which means, if we have more than one key in our env, we cannot make the LangGraph agent pick the one that comes later in the `SUPPORTED_API_KEYS` list.
> 
> Then, let's say someone tries to launch multiple agent servers as described in [README](https://github.com/google/A2A/blob/8c8e0eabfe0558159044455e917b38c7472a9fe1/samples/python/README.md), `ReimburementAgent` (setting `GOOGLE_API_KEY`) and the LangGraph-based one (e.g., with `OPENAI_API_KEY`):
> 
> 1. `cd samples/python` (where they have their `.env`).
> 2. `uv run agents/google_adk` _and_ `uv run agents/langgraph`
> 
> Both of the agents would still use `GOOGLE_API_KEY`.
> 
> Maybe if the script has access to more than one key, it can be updated to require specifying the provider to use? Or provide other ways of choosing the provider?

+1 to what you said. Every agent should read all the env variables because env variables are set at the system level and all running process should have access to it.

What we can do is have each agent define what keys they are going to use and perform validation at the agent level if any of the required env vars which the agent needs to use is not set.

HomeroRR:
> > I have a question about this patch. In this implementation, `get_api_key()` loops through known API keys in a predetermined order. Which means, if we have more than one key in our env, we cannot make the LangGraph agent pick the one that comes later in the `SUPPORTED_API_KEYS` list.
> > Then, let's say someone tries to launch multiple agent servers as described in [README](https://github.com/google/A2A/blob/8c8e0eabfe0558159044455e917b38c7472a9fe1/samples/python/README.md), `ReimburementAgent` (setting `GOOGLE_API_KEY`) and the LangGraph-based one (e.g., with `OPENAI_API_KEY`):
> > 
> > 1. `cd samples/python` (where they have their `.env`).
> > 2. `uv run agents/google_adk` _and_ `uv run agents/langgraph`
> > 
> > Both of the agents would still use `GOOGLE_API_KEY`.
> > Maybe if the script has access to more than one key, it can be updated to require specifying the provider to use? Or provide other ways of choosing the provider?
> 
> +1 to what you said. Every agent should read all the env variables because env variables are set at the system level and all running process should have access to it.
> 
> What we can do is have each agent define what keys they are going to use and perform validation at the agent level if any of the required env vars which the agent needs to use is not set.

I agree. Users should have a way to specify the api key they want if they have multiple in their environment.
I have updated the PR to now take in an optional `SELECTED_API_KEY_NAME` which will let users select which api key they want to use.

princejha95:
> > > I have a question about this patch. In this implementation, `get_api_key()` loops through known API keys in a predetermined order. Which means, if we have more than one key in our env, we cannot make the LangGraph agent pick the one that comes later in the `SUPPORTED_API_KEYS` list.
> > > Then, let's say someone tries to launch multiple agent servers as described in [README](https://github.com/google/A2A/blob/8c8e0eabfe0558159044455e917b38c7472a9fe1/samples/python/README.md), `ReimburementAgent` (setting `GOOGLE_API_KEY`) and the LangGraph-based one (e.g., with `OPENAI_API_KEY`):
> > > 
> > > 1. `cd samples/python` (where they have their `.env`).
> > > 2. `uv run agents/google_adk` _and_ `uv run agents/langgraph`
> > > 
> > > Both of the agents would still use `GOOGLE_API_KEY`.
> > > Maybe if the script has access to more than one key, it can be updated to require specifying the provider to use? Or provide other ways of choosing the provider?
> > 
> > 
> > +1 to what you said. Every agent should read all the env variables because env variables are set at the system level and all running process should have access to it.
> > What we can do is have each agent define what keys they are going to use and perform validation at the agent level if any of the required env vars which the agent needs to use is not set.
> 
> I agree. Users should have a way to specify the api key they want if they have multiple in their environment. I have updated the PR to now take in an optional `SELECTED_API_KEY_NAME` which will let users select which api key they want to use.

The problem with this approach is we have to copy this logic for all different types of agents. Rather than doing that, can we put this logic of checking env variable, checking availability of api keys etc in some common package so that it can be reused across agents ?

holtskinner:

Issue #84, opened 2025-04-11
Mismatch between AI Response and Structured Output in LangGraph Example


claudiourbina:
In the LangGraph example, you are using create_react_agent with a structured output (which is required for A2A). However, under the hood, what create_react_agent does is run the model twice: once with bind_tools and once with with_structured_output. With bind_tools, the model knows about the tool, and its response is correct, but it's not formatted, so the status can't be extracted. On the other hand, with with_structured_output, the model doesn't know about the tools (because none are bound), so the response doesn't include any information about them, and it's completely different from the bind_tools response. In fact, sometimes this response doesn't even make sense. If you look at the state, you'll see that the first option is correct, while the second one is not, as the first will appear in the message list, and the second will be in the structured output field.

```json
{
  "messages": [
    {
      "role": "HumanMessage",
      "content": "Hello, I want to build a minisite"
    },
    {
      "role": "AIMessage",
      "content": "Hello! I can help you with that. To build the minisite, I'll need you to provide information about the tool you'd like to add. I'll need the name, description, key benefits, key features, FAQs, real use cases, and how to get started with the tool.\n\nIs it okay if we start with that?"
    }
  ],
  "structured_response": {
    "status": "input_required",
    "message": "Hello! To help you build your minisite, I need more information. What topic would you like the minisite to be about? Do you have any content ready, such as text or images? Do you have any design or layout preferences?"
  }
}
```

kthota-g:
The langgraph agent is a simple example to demonstrate streaming, non-streaming, multi-turn examples over A2A. While there can be loss of precision, for the original message, due to the second LLM call to generate structured output, the currency conversion example is simple and it still helps with the purpose of demonstrating various interactions in the protocol.

For complex agents, generating structured output might require different strategies. For example, you may pass additional instructions with tool descriptions to response_format parameter to set context during structured output generation.

PR #113, opened 2025-04-12
fix: Prioritize message history over structured output in LangGraph agent

crazywolf132:
This PR fixes issue #84 where there's a mismatch between the AI response in the message history and the structured output in the LangGraph example.

## Problem
When using `create_react_agent` with a structured output format, LangGraph runs the model twice:
1. First with `bind_tools` to get a response that knows about the tools (stored in message history)
2. Then with `with_structured_output` to get a structured response (stored in structured_response)

These two responses can be inconsistent, as mentioned in issue #84. The first response (with tools) appears in the message history, while the second response (structured) appears in the `structured_response` field.

## Solution
The fix modifies the `get_agent_response` method to prioritize the message from the message history over the structured output when they conflict:

```python
# Get the last message from the message history
# This is the response from the first LLM call with bind_tools
messages = current_state.values.get('messages', [])
last_message_content = None
if messages and len(messages) > 0:
    last_message = messages[-1]
    if isinstance(last_message, AIMessage) and last_message.content:
        last_message_content = last_message.content

# Use the message from the message history if available, otherwise use the structured response
message_content = last_message_content or structured_response.message
```

## Testing
I created a simple test script to verify that the fix works as expected. The test confirms that:

1. The original implementation uses only the structured content, which can be inconsistent with the message history
2. The fixed implementation prioritizes the message content from the message history, which is more likely to be correct

This fix ensures that the response shown to the user is consistent with the tool usage and reasoning shown in the message history, which is important for a good user experience.

### DIFF
--- a/samples/python/agents/langgraph/agent.py
+++ b/samples/python/agents/langgraph/agent.py
@@ -25,7 +25,7 @@ def get_exchange_rate(
 
     Returns:
         A dictionary containing the exchange rate data, or an error message if the request fails.
-    """    
+    """
     try:
         response = httpx.get(
             f"https://api.frankfurter.app/{currency_date}",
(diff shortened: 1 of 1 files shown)


binbandit:
bump

binbandit:
bump

binbandit:
@kthota-g Could I please get a review?

holtskinner:

Issue #104, opened 2025-04-11
REQUEST: Create a separate task state explicitly for rejected tasks


Mocuto:
I was looking through the documentation and noticed it specifically called out that a server agent may reject a task given to it by a client: 

<img width="911" alt="Image" src="https://github.com/user-attachments/assets/f1a9aa8f-5f6e-408f-886c-db16ebece279" />

However I notice that there isn't a task state for rejected tasks:

<img width="223" alt="Image" src="https://github.com/user-attachments/assets/d7d0b548-6256-430a-bd44-f4101a382587" />

I believe a task being rejected is distinct from it being failed or an error, as it communicates explicitly that either the agent does not support the task being requested or that it cannot be completed due to some other factors (i.e., the client may be on the wrong subscription tier for the agent product).

Can we amend the task states with a clear "rejected" state? We can then include in via the Messages construct the rejection reason.

holtskinner:
Done. This is in Version 0.2.1 of the protocol

PR #115, opened 2025-04-12
feat: add rejected task state to a2a protocol

crazywolf132:
This PR adds a new 'rejected' task state to the A2A protocol, addressing issue #104.

## Problem
Currently, the A2A protocol has states for tasks that are completed, canceled, or failed, but there's no way to explicitly indicate that a task was rejected. This makes it difficult to distinguish between tasks that failed due to errors and tasks that were explicitly rejected by the agent.

## Changes
1. Added the "rejected" state to the TaskState enum in the JSON schema (`specification/json/a2a.json`)
2. Added the "rejected" state to the TaskState enum in the Python implementation (`samples/python/common/types.py`)
3. Added the "rejected" state to the TaskState type in the TypeScript implementation (`samples/js/src/schema.ts`)

## Testing
The changes have been tested by:
1. Verifying that the Python implementation correctly defines and uses the "rejected" state
2. Verifying that the TypeScript implementation correctly defines the "rejected" state
3. Validating that the JSON schema remains valid after the changes

## Documentation
No documentation updates were needed as the task states are primarily defined in the schema files that have been updated.

### DIFF
--- a/samples/js/src/schema.ts
+++ b/samples/js/src/schema.ts
@@ -96,6 +96,7 @@ export type TaskState =
   | "completed"
   | "canceled"
   | "failed"
+  | "rejected"
   | "unknown";
 
 /**
--- a/samples/python/common/types.py
+++ b/samples/python/common/types.py
@@ -15,6 +15,7 @@ class TaskState(str, Enum):
     COMPLETED = "completed"
     CANCELED = "canceled"
     FAILED = "failed"
+    REJECTED = "rejected"
     UNKNOWN = "unknown"
 
 
@@ -100,7 +101,7 @@ class TaskStatusUpdateEvent(BaseModel):
 
 class TaskArtifactUpdateEvent(BaseModel):
     id: str
-    artifact: Artifact    
+    artifact: Artifact
     metadata: dict[str, Any] | None = None
 
 
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -72,7 +72,7 @@
           },
           "url": {
             "title": "Url",
-            "type": "string"            
+            "type": "string"
           },
           "provider": {
             "anyOf": [
@@ -715,7 +715,7 @@
         },
         "title": "GetTaskPushNotificationResponse",
         "type": "object"
-      },      
+      },
       "GetTaskRequest": {
         "properties": {
           "jsonrpc": {
@@ -1126,7 +1126,7 @@
             },
             "title": "Parts",
             "type": "array"
-          },          
+          },
           "metadata": {
             "anyOf": [
               {
@@ -1192,7 +1192,7 @@
           },
           "token": {
             "title": "Token",
-            "anyOf": [              
+            "anyOf": [
               {
                 "type": "string"
               },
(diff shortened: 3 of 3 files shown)


binbandit:
bump

binbandit:
bump

binbandit:
@kthota-g Could I please get a review?

kthota-g:
@crazywolf132 We will add new states in the next revision of the spec. We will park this PR for now.

Mocuto:
Do you have an ETA on when this revision to the spec will be available? My team is actively building an agent2agent demo that will require `rejected` task states. Additional info on timeline would be very helpful for project planning

samples/js/src/schema.ts
```diff
@@ -96,6 +96,7 @@ export type TaskState =
   | "completed"
   | "canceled"
   | "failed"
+  | "rejected"
```
djsamseng, line 99: Isn't this a duplicate of `canceled` or `failed`? Couldn't this be implemented already?

A. If it's a single task synchronous request, then reject the request and don't create the task
```python
return SendTaskResponse(
  id=request.id,
  error=JSONRPCError(
    code=-32000, # Server error
    message=f"Server error: {str(e)}"
  )
)
```

B. If it's a subscription asynchronous request, then cancel the task that was already started (or reject the initial request before starting the asynchronous request which is the same as A.)

```python
task_status = TaskStatus(
  state=TaskState.CANCELED,
  message=Message(
    role="agent",
    parts=[
      typing.cast(Part, {
        "type": "text",
        "text": "No ollama agent running to process the request"
      })
    ]
  )
)
task_update_event = TaskStatusUpdateEvent(
  id=task_id,
  status=task_status,
  final=True,
)
await self.enqueue_events_for_sse(
  task_id=task_id,
  task_update_event=task_update_event,
)
```

Mocuto: Please see the original issue https://github.com/google/A2A/issues/104. I will briefly resummarize the rationale:
1. It's in your own documentation that it is possible for a task to be rejected by the server
2. Rejecting a task is semantically entirely different from an server error. A server error indicates that the server encountered an unexpected issue or problem, whereas a task reject explicitly indicates that the task requested is out of scope for the agent. This is different from a typically HTTP error code, where maybe a parameter is misconfigured; this is the server agent _choosing_ not to do a task given its own internal policy for determining what work should be done.

binbandit:
@holtskinner This one is ready to merge

holtskinner:
This has been addressed in the official spec https://github.com/google/A2A/blob/ab0509dc7a04aa909c985779b21b7e8040ee5042/types/src/types.ts#L253

binbandit:
> This has been addressed in the official spec
> 
> https://github.com/google/A2A/blob/ab0509dc7a04aa909c985779b21b7e8040ee5042/types/src/types.ts#L253

Bit rough that it just gets put into a google employees commit, without credit...

This PR was raised quickly after the launch of A2A publicly... Not like i just put it up, and it was after the commit from @kthota-g 

This kind of stuff discourages people from contributing to this community.

PR #136, opened 2025-04-13
add initial config for remote agent address

zhenjunMa:
# Feature
add initial config for remote agent address, and if it exists, display in the UI

### DIFF
--- a/demo/ui/utils/config.py
+++ b/demo/ui/utils/config.py
@@ -1,0 +1,39 @@
+import tomllib
+from pathlib import Path
+
+def get_project_root() -> Path:
+    """Get the project root directory"""
+    return Path(__file__).resolve().parent.parent
+
+PROJECT_ROOT = get_project_root()
+
+class Config:
+
+    address = []
+
+    def __init__(self):
+        self._load_initial_config()
+
+    @staticmethod
+    def _get_config_path() -> Path:
+        root = PROJECT_ROOT
+        config_path = root / "config" / "config.toml"
+        if config_path.exists():
+            return config_path
+        raise FileNotFoundError("No configuration file found in config directory")
+
+    def _load_initial_config(self):
+        config_path = self._get_config_path()
+        with config_path.open("rb") as f:
+            raw_config = tomllib.load(f)
+        base_remote_agent_addresses = raw_config.get("remote_agent_addresses", {})
+        self.address = list(base_remote_agent_addresses.values())
+
+    @property
+    def remote_agent_addresses(self) -> list[str]:
+        return self.address
+
+config = Config()
+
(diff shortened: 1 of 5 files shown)


demo/ui/utils/config.py
```diff
@@ -0,0 +1,39 @@
...
+    @staticmethod
+    def _get_config_path() -> Path:
+        root = PROJECT_ROOT
+        config_path = root / "config" / "config.toml"
+        if config_path.exists():
+            return config_path
+        raise FileNotFoundError("No configuration file found in config directory")
+
+    def _load_initial_config(self):
+        config_path = self._get_config_path()
```
princejha95: _get_config_path is a static method and should not be called using self

demo/ui/utils/config.py
```diff
@@ -0,0 +1,39 @@
...
+        with config_path.open("rb") as f:
+            raw_config = tomllib.load(f)
+        base_remote_agent_addresses = raw_config.get("remote_agent_addresses", {})
```
princejha95: Let's define a constant for **remote_agent_addresses**?

demo/ui/utils/config.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_project_root() -> Path:
+    """Get the project root directory"""
+    return Path(__file__).resolve().parent.parent
+
+PROJECT_ROOT = get_project_root()
+
+class Config:
+
+    address = []
```
princejha95: Let's initialize inside the constructor only ?

demo/ui/utils/config.py
```diff
@@ -0,0 +1,42 @@
...
+        root = PROJECT_ROOT
+        config_path = root / "config" / "config.toml"
+        if config_path.exists():
+            return config_path
+        raise FileNotFoundError("No configuration file found in config directory")
+
+    def _load_initial_config(self):
+        config_path = Config._get_config_path()
+        with config_path.open("rb") as f:
+            raw_config = tomllib.load(f)
```
princejha95: Let's put the entire **with ...** inside try except block as tomllib.load() can return different types of errors and we should catch those errors return a user-friendly error message back to client to avoid leakage of any critical info

demo/ui/utils/config.py
```diff
@@ -0,0 +1,42 @@
+import tomllib
```
princejha95, line 1: I'll suggest writing UTs as well for these changes or you can take it in a different PR as well. Whatever suits you

zhenjunMa: This feature is my own idea. Let’s first see if the community recognizes this concept😄

holtskinner:

PR #149, opened 2025-04-14
Add New Semantic Kernel Agent Sample

moonbox3:
Add a new Semantic Kernel Agent Sample

- Introduces a comprehensive sample demonstrating how to implement and consume a Semantic Kernel–driven travel agent.
- Showcases how to chain multiple specialized sub-agents used as plugins -- such as a currency exchange agent and an activity planning agent -- under a primary TravelManagerAgent, all orchestrated via Semantic Kernel.
- Illustrates streaming responses, enabling partial updates to a frontend interface, as well as how to end the streaming loop when the agent requires user input or has completed.
- Currently manages state via SK's `ChatHistoryAgentThread` which uses the provided `sessionId`. The thread and underlying messages are ephemeral and not persisted right now.

Example SK Agent on the left, with the sample host/cli on the right:

image


moonbox3:
Hi @rajeshvelicheti and @kthota-g, do we have an ETA on when we could get this merged to main? Thank you for your help.

samples/python/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "a2a-samples"
 version = "0.1.0"
 description = "Agent2Agent samples"
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = ">=3.11"
```
moonbox3: I didn't find any dependency that requires Py 3.13 (docs show Py 3.13 as required version, however this config allows for 3.12). One can install uv with Py 3.11/3.12 by running `uv venv --python "3.11"` or `uv venv --python "3.12"`.

rajeshvelicheti (changes requested):
This is great, thank you so much for bringing Semantic Kernel Example in. 

Is there we could isolate this in to its own project? Basically add a new pyproject.toml under python/semantickernel

We did similar for the crewai example.

moonbox3:
> This is great, thank you so much for bringing Semantic Kernel Example in.
> 
> Is there we could isolate this in to its own project? Basically add a new pyproject.toml under python/semantickernel
> 
> We did similar for the crewai example.

Thank you for the feedback, @rajeshvelicheti. Will make the change ASAP.

moonbox3:
@rajeshvelicheti, I've made the updates to make the Semantic Kernel sample a standalone project. I've updated our README accordingly, and after testing, it is working well. Please have another look. Thank you for your help.

samples/python/agents/semantickernel/task_manager.py
```diff
@@ -0,0 +1,200 @@
...
+                    new_status.state = TaskState.INPUT_REQUIRED
+                    new_status.message = Message(
+                        role="agent",
+                        parts=[{"type": "text", "text": text_content}],
+                    )
+                    # End the stream if we need user input
+                    final = True
+                elif is_done:
+                    new_status.state = TaskState.COMPLETED
+                    new_status.message = Message(
```
kthota-g: hi, this block needs some changes to send TaskArtifactUpdateEvent during streaming. something like this:

```python
                elif is_done:
                    new_status.state = TaskState.COMPLETED
                    artifact = Artifact(parts=[{"type": "text", "text": text_content}], index=0, append=False)                    
                    # End the stream if the agent is fully done
                    final = True
                else:
                    # Still "WORKING"
                    new_status.message = Message(
                        role="agent",
                        parts=[{"type": "text", "text": text_content}],
                    )

                if artifact:
                    task_artifact_update_event = TaskArtifactUpdateEvent(
                        id=request.params.id, artifact=artifact
                    )
                    await self.enqueue_events_for_sse(
                        request.params.id, task_artifact_update_event
                    )   
                # Persist + notify
                updated_task = await self.update_store(request.params.id, new_status, [artifact] if artifact else None)
```

samples/python/agents/semantickernel/__main__.py
```diff
@@ -0,0 +1,66 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+load_dotenv()
+
+
+@click.command()
+@click.option("--host", default="localhost")
+@click.option("--port", default=10000)
```
kthota-g: Can you change the default port to 10020 so there are no conflicts with existing agents?

kthota-g (commented):
@moonbox3 couple of additional changes

samples/python/agents/semantickernel/__main__.py
moonbox3: Done, thank you

samples/python/agents/semantickernel/task_manager.py
moonbox3: Updated this, appreciate the feedback.

moonbox3:
@rajeshvelicheti @kthota-g we're trying really hard to get this in ASAP, any chance we can push it forward? Appreciate your support.

rajeshvelicheti (approved):

kthota-g (approved):

moonbox3:
Thanks again for the support.

Newread23 (commented):
@a2a-bot

PR #179, opened 2025-04-16
[New Sample] LlamaIndex Workflows Sample Integration

logan-markewich:
This PR does a few things
- adds a LlamaIndex Workflows sample integration with A2A, highlighting file upload, streaming, session based state management, and general llama-index usage
- updates readme's to include links to this sample code
- updates the CLI example to optionally allow attaching files to requests

Let me know if there is anything I can improve here! The task manager was a little complicated to figure out, I mostly adapted what the langgraph example had.


samples/python/agents/llama_index_file_chat/__main__.py
```diff
@@ -0,0 +1,77 @@
...
+from dotenv import load_dotenv
+
+load_dotenv()
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+@click.command()
+@click.option("--host", "host", default="localhost")
+@click.option("--port", "port", default=10000)
```
rajeshvelicheti: It might help if the default port does not conflict with existing samples.

rajeshvelicheti (changes requested):
Thank you so much for the contribution, excited to see the LlamaIndex sample. 

LGTM overall, but a minor change that might help with maintaining the port numbers overall.

kthota-g:
Also the CLA check is failing that prevents the PR merge.

samples/python/agents/llama_index_file_chat/__main__.py
logan-markewich: ah good catch

logan-markewich:
@kthota-g I just added myself to the CLA -- hopefully the test passes on the next run

samples/python/agents/llama_index_file_chat/__main__.py
logan-markewich: Updated! (and updated the README to include the proper args to set the cli endpoint properly)

rajeshvelicheti (approved):

rajeshvelicheti (changes requested):
@logan-markewich thanks for making the port changes. I was quickly testing this with the cli and it looks great. However, the demo ui (demo/ui) will be broken. It does not support files yet. Is it something that you will be able to address?

logan-markewich:
@rajeshvelicheti ah I completely missed that there was a UI lol

Let me see what is required to support this...

logan-markewich:
@rajeshvelicheti oh boy, this is quite complicated lol but basically, the UI runs an agent, with other agents as tools

So we need to
1. Support file input in the UI
2. Possibly support passing files from the top-level agent to the tool (I hope ADK supports this, I have not confirmed)

Going to see how hard this is to do now..

I might suggest filling in the README for the UI in at some point as well, took me a while to figure out what this UI was doing 😅

rajeshvelicheti:
> @rajeshvelicheti oh boy, this is quite complicated lol but basically, the UI runs an agent, with other agents as tools
> 
> So we need to
> 
> 1. Support file input in the UI
> 2. Possibly support passing files from the top-level agent to the tool (I hope ADK supports this, I have not confirmed)
> 
> Going to see how hard this is to do now..
> 
> I might suggest filling in the README for the UI in at some point as well, took me a while to figure out what this UI was doing 😅

Yeah you are right, the UI runs the agent, there is also a host agent.

Support file input in the UI, the artifacts service could be your friend here. But there is some bit of code that might still be required.

demo/ui/service/server/adk_host_manager.py

I agree, it is not very straightforward, please let me know if you have any questions, we can find some time that works for both of us.

Alternatively, we just update the Readme suggesting the agents works with cli and UI support might be coming later.

kthota-g (commented):
@logan-markewich couple of additional changes.

edit: please ignore, wrong PR

logan-markewich:
@rajeshvelicheti Yea I don't think I have quite enough time to learn how the UI connects together and I don't want to hold up this PR -- added a note in the readme for this example. 

We can make the UI support multimodal in another PR another day :) (this is something that would be useful to any agent/example, not just this one!)

Also resolved merge conflicts from SK changes, should be good to merge

rajeshvelicheti:
Thanks for the contribution @logan-markewich. Great to see lllamaindex sample.

rajeshvelicheti:
https://github.com/google/A2A/issues/197 to track the UI limitation.

PR #185, opened 2025-04-16
Implement /agent-card endpoint presented in client.ts

egor-baranov:
In file /samples/js/src/client/client.ts there are an agentCard method that triggers a GET request trying to retrieve this data. But from the server-side it's defined only in /.well-known/agent.json. So we need to additionally implement the same method through endpoint compatible with specification

### DIFF
--- a/samples/python/common/server/server.py
+++ b/samples/python/common/server/server.py
@@ -46,6 +46,9 @@ class A2AServer:
         self.app.add_route(
             "/.well-known/agent.json", self._get_agent_card, methods=["GET"]
         )
+        self.app.add_route(
+            "/agent-card", self._get_agent_card, methods=["GET"]
+        )
 
     def start(self):
         if self.agent_card is None:


olegchir:
Looks legit

egor-baranov:
Doc string says

```typescript
/**
   * Retrieves the AgentCard.
   * Note: The standard A2A protocol doesn't define a JSON-RPC method for this.
   * This implementation fetches it from a hypothetical '/agent-card' endpoint
   * on the same server, assuming it's provided out-of-band.
   * Caches the result after the first successful fetch.
   */
```

But there are no implementations to retrieve an agent card from /.well-known/agent.json. Should it be a distinguished method in API or there are need to be a custom implementation of AgentCard? I just don't see a point to have the same method which is provided from the agent.json out of the box. But not presented by an API client

Any thoughts on that?

mbleigh:
I don't really understand what you're trying to do here - the [spec specifies](https://google.github.io/A2A/#/topics/agent_discovery?id=open-discovery) that `/.well-known/agent.json` is the discovery URL, there is no `/agent-card` endpoint as far as I know. Can you help me understand what part of the spec you feel is lacking and needs to be implemented here?

holtskinner:

PR #182, opened 2025-04-16
Split common and agent-specific dependencies

darkhaniop:
This PR addresses a problem described in issue #181, to help:
* reduce bloating dependency trees of new agent projects (both in a2a-samples and external);
* establish better patterns for adding new agent examples to this repository;
* help reduce merge conflicts by lessening the need to touch `samples/python/pyproject.toml` due to updates in sample agents/clients.

## Summary

Moves subproject-specific, e.g. `agents/langgraph`, dependencies into the respective directories:
* Adds `agents/googl_adk/pyproject.toml`.
* Adds `agents/langgraph/pyproject.toml`.
* Adds `hosts/cli/pyproject.toml`.
* Adds `hosts/multiagent/pyproject.toml` and a short readme.
* Updates dependencies in the upper-level `samples/python/pyproject.toml`.
* Updates sample agent run instructions in `samples/python/README.md`.

Issue #181, opened 2025-04-16
Split common and agent-specific dependencies


darkhaniop:
## Problem

Currently, if we want to use `common` modules, `A2AServer` and/or `common.types`, by installing `samples/python` in a new project, all the dependencies for `agents/google_adk`, `agents/langgraph`, and hosts are also installed. **Which adds ~90-100 packages to the dependency tree of the new project, while modules likely to be used in other projects (`common`+`utils`) require only 24 relatively small packages.**

## Proposed solution

I propose splitting "common" package requirements from those of agent and client examples. This can be done with minimal changes to the current code base by moving dependencies to `pyproject.toml` definitions in the respective sub-projects, e.g., `agents/langgraph/pyproject.toml`.

Doing this would:
* help reduce bloating dependency trees of new agent projects (both in a2a-samples and external);
* and help establish better patterns for adding new agent examples to this repository.

## Relevant issues and PRs

* [PR #149](https://github.com/google/A2A/pull/149#pullrequestreview-2769116396)

darkhaniop:
Closing. Resolved in #182 and #208.

PR #208, opened 2025-04-17
Fix: Add missing "a2a-samples" dependency in hosts/*

darkhaniop:
Adds "a2a-samples" as a dependency in `hosts/cli/pyproject.toml` and `hosts/multiagent/pyproject.toml`. Also, fixes `uv.lock` being out of sync (was missing some dev-dependencies).

Discovered this issue after running `uv sync` in a freshly cloned copy.

## Details

After the "splitting subproject-specific and common dependencies" change was applied (PR #182), I wanted to verify that environments worked as expected. So, I cloned the `main` branch into a new location (with no preexisting `.venv`'s) and after running `uv sync` noticed that `uv.lock` was out of sync and the "a2a-samples" dependency was missing in `hosts/cli` and `hosts/multiagent`. While the `multiagent` still worked from `demo/ui` because `demo/ui` directly requires "a2a-samples," the CLI client did not.

With this fix, the CLI client works as expected.

### uv version:

* uv 0.6.14

---

Refs: basically, this is a follow-up to PR #182

### DIFF
--- a/samples/python/hosts/cli/pyproject.toml
+++ b/samples/python/hosts/cli/pyproject.toml
@@ -5,6 +5,7 @@ description = "A CLI application that demonstrates the capabilities of an A2ACli
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
+    "a2a-samples",
     "asyncclick>=8.1.8",
     "sse-starlette>=2.2.1",
     "starlette>=0.46.1",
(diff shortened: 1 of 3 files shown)


pstephengoogle (approved):
Thanks!

darkhaniop:
@pstephengoogle Thanks for reviewing this quickly. BTW, I believe #181, the issue related to the previous PR, can be closed.

*Edit: I just closed 181, for some reason, I thought GitHub wouldn't allow me to do that.*

PR #229, opened 2025-04-18
a2a:cli - Correctly retrieve agent URL

gate4ai-alex:
Fix: Correctly retrieve agent URL.

This PR addresses an issue where the base server URL was being used instead of the agent-specific URL from the AgentCard.


holtskinner:

PR #233, opened 2025-04-19
fix the Part schema

logan272:
Making the `Part.type` field optional. This change will resolve some TypeScript type errors in the js sample code.

### DIFF
--- a/samples/js/src/schema.ts
+++ b/samples/js/src/schema.ts
@@ -310,7 +310,7 @@ export type FileContent = FileContentBytes | FileContentUri;
  * Represents a part of a message containing text content.
  */
 export interface TextPart {
-  type: "text";
+  type?: "text" | null;
 
   /**
    * The text content.
@@ -330,7 +330,7 @@ export interface FilePart {
   /**
    * Type identifier for this part.
    */
-  type: "file";
+  type?: "file" | null;
 
   /**
    * The file content, provided either inline or via URI.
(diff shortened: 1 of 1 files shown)


logan272:
Signed the CLA.

elliot-at-liminalnook:
@logan272 , is it not the typescript that is wrong rather than the schema?

logan272:
@robert-at-pretension-io We can differentiate the Part type from its content type, which makes the type field of Part not strictly necessary. Making it optional could save us some keystrokes and reduce the number of bytes that need to be transported.

logan272:
@robert-at-pretension-io Yes, that's the schema. I only updated the TypeScript sample to fix the type errors. If you're fine with it, I can also update the PR to include changes to the schema and documentation.

samples/js/src/schema.ts
```diff
@@ -310,7 +310,7 @@ export type FileContent = FileContentBytes | FileContentUri;
  * Represents a part of a message containing text content.
  */
 export interface TextPart {
-  type: "text";
+  type?: "text" | null;
```
markgoho, line 313: just as a general principle, `undefined` is preferred to `null` - I wonder why `type?: "text"` wouldn't be acceptable and why `| null` is necessary

logan272: I agree that null is not necessary. I noticed that null is used consistently throughout the codebase. Think we can remove all the null usage in a further PR.

holtskinner:

PR #250, opened 2025-04-20
Add type safety and environment variable support

htondro:
This PR includes two main fixes:

1. Added dotenv package and environment variable support:
   - Added dotenv package to handle environment variables
   - Implemented proper environment variable loading in both coder and movie-agent modules

2. Fixed type errors in agent implementations:
   - Added missing `type: "text"` property to message parts in both agents
   - Fixed type definitions for message parts in error and status messages

These changes improve type safety and configuration management in the agent modules.

### DIFF
--- a/samples/js/package-lock.json
+++ b/samples/js/package-lock.json
@@ -14,6 +14,7 @@
         "@types/express": "^5.0.1",
         "body-parser": "^2.2.0",
         "cors": "^2.8.5",
+        "dotenv": "^16.5.0",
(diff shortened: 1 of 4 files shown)


markgoho (commented):
consider this PR does two completely separate things, a good practice would be to separate this into two PRs so each change could be independently reviewed and approved (or rejected)

personally, I use Bun and wouldn't need a library added to handle environment variables, Bun just automatically detects them

holtskinner:

PR #267, opened 2025-04-22
Avoid using null and replace null with undefined where applicable.

logan272:
This PR replaces all instances of null with undefined where applicable, following the general principle that undefined is preferred over null.

See also: https://github.com/google/A2A/pull/233


holtskinner:

PR #271, opened 2025-04-22
Update a2a.json to omit null properties

logan272:
Fix: https://github.com/google/A2A/issues/264

See also:
- https://github.com/google/A2A/pull/267
- https://github.com/google/A2A/pull/233

### DIFF
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -13,15 +13,7 @@
             "type": "array"
           },
           "credentials": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "null"
-              }
-            ],
-            "default": null,
+            "type": "string",
             "title": "Credentials"
           }
         },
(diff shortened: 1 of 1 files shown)


rajeshvelicheti:
@kthota-g could you please check this. Generation tool might be adding these and I can change it if required.

kthota-g:
yes, this is caused due to schema generation for attributes defined as `attr: Foo | None = None`. Agree that null need not be explictly set for optional fields.

whitlockjc:
> yes, this is caused due to schema generation for attributes defined as `attr: Foo | None = None`. Agree that null need not be explictly set for optional fields.

You can work around this using `.model_dump(exclude_none=True)`, so if this PR can be updated to fix the test suite so that all calls to `model_dump` pass `exclude_none=True`, we should be able to move this forward.

holtskinner:
I think this is obsolete after https://github.com/google-a2a/A2A/pull/352

PR #283, opened 2025-04-23
Fix the test module problem

vongosling:
Fix the common.types not resolved problem in virtual env

### DIFF
--- a/tests/test_a2a_spec.py
+++ b/tests/test_a2a_spec.py
@@ -1,12 +1,16 @@
 # test_schema.py
 import pytest
 import json
+import sys
 from uuid import uuid4
 from datetime import datetime
 from pathlib import Path
 
+# Add parent directory to sys.path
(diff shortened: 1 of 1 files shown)


rajeshvelicheti:
Following these steps might be more useful for now as the structure is altered.

cd samples/python
uv venv
source .venv/bin/activate
cd ../../tests
uv run pytest -v -s

Please let me know if this does not work.

vongosling:
I just used the newest code, but it doesn't work.  If use your code, more errors will occur. That's why I test one file using ’uv run pytest -v -s test_a2a_spec.py'. You could clone code from scratch, Maybe it still needs to change the instruction, such as changing the toml file.

holtskinner:
Not sure if this is still needed, but if it is, those test files have been moved to https://github.com/google-a2a/a2a-samples

PR #296, opened 2025-04-25
Proposal: Implement A2A Agent Registrar for Curated Agent Discovery(Registry-Based)

### LINKED ISSUE #297: Proposal: Implement A2A Agent Registrar for Curated Agent Discovery(Registry-Based)
I propose implementing a  Agent Registrar to facilitate curated discovery of agents across organizations. This would provide a centralized mechanism for agent registration, discovery, and management while respecting organizational boundaries and visibility controls.

PR: https://github.com/google/A2A/pull/296 

### Benefits

- Enterprise-grade agent discovery with organizational boundaries
- Standardized registry interface for consistent agent discovery
- Controlled visibility for sensitive or internal-only agents
- Authentication model preserving security of agent registrations

liuzengh:
close #297


topics/agent_discovery.md
```diff
@@ -20,7 +20,769 @@ We recommend enterprises host their agent cards at a well-known path. Specifical
...
+    "artifacts": [
+      {
+        "name": "organization-registration-result",
+        "parts": [
+          {
+            "type": "data",
+            "data": {
+              "registrationStatus": "success",
+              "organizationID": "organization-fintech-123",
+              "organizationToken": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...",
```
darkhaniop: By looking at all of the request-response examples, I noticed that `organizationToken` is obtained when registering the organization (no mention of other kinds of tokens), and the same token is used to manage the organization (update, add, and remove agents) as well as to discover non-public agents.

Unless an additional mechanism to issue tokens with limited scope/time is added, I think this "organization registration" response should include another token with **read-only permissions**. This additional token can then be distributed to the internal clients (e.g., partners, non-dev employees) to allow them to discover private agents.

What do you think?

sandyskies: The organizationToken is similar to a admin token and can be implemented based on the OAuth Bearer token. Of course, each organization can issue some tokens controlled by RBAC (Role-Based Access Control) to others for use, such as read-only tokens, etc.

darkhaniop: @sandyskies, it makes sense, thanks for the explanation.

liuzengh:
The previously submitted PR was closed because the repository administrator moved the documentation from the gh-pages branch to the main branch via force-push. It has been resubmitted to the main branch by new PR(#418).

PR #311, opened 2025-04-28
Update host_agent.py  prompt  function call name 

Frankdog:
google-a2a/a2a-samples#44

function call name  in tools is `send_task`
while in prompt is  `create_task`
change prompt  `create_task` to `send_task `


holtskinner:

PR #313, opened 2025-04-28
website: remove references to forked A2A library and use direct Google imports instead

cndoit18:
The forked version has behind the main branch.


cndoit18:
cc @holtskinner

PR #318, opened 2025-04-28
Update the Task Manager in the Google ADK sample to be reusable

whitlockjc:
This commit decouples the `TaskManager` in `samples/python/agents/google_adk` to be reusable by all Google ADK samples.

**Note:** This could potentially be refactored into a Google ADK specific `common` directory.


rajeshvelicheti:
Should we wait for the sdk and check if this could be generic across ?

whitlockjc:
> Should we wait for the sdk and check if this could be generic across ?

That was originally my thought but as we use Google ADK for sample agents, this simplifies the onus on the agent author quite a bit. Thoughts?

rajeshvelicheti (approved):

PR #322, opened 2025-04-28
Fix a2a spec inconsistencies with sample scripts, JSON-schema, and JSON-RPC specs

darkhaniop:
In client-server messaging, the A2A protocol uses the [JSON-RPC 2.0](https://www.jsonrpc.org/specification) protocol. The protocol specification is written using the [JSON-Schema specification](https://json-schema.org/). Fully complying with the standards can help validate the payloads more accurately and help better test different client and server implementations.

In addition to making the A2A spec more compliant with the mentioned standards, changes in this PR are meant to fix a few minor bugs in the A2A protocol spec (mismatches from what was implemented in the sample A2A server and clients.

### Summary of changes

* Add a missing definition for "TaskSendParams.acceptedOutputModes".

* Remove "default" values for "const" fields. Compliance with JSON schema, ver. draft6+ and reduce ambiguity. https://json-schema.org/understanding-json-schema/reference/const

* Make fields "jsonrpc" and "id" required on requests and responses. Compliance with JSON RPC v2.0.

* Make fields "error.code" and "error.message" required on errors.
* Change field "error.data" to allow "any" and "null" and make it "not required."

* Sort non-top-level dict keys to make it more aligned with the pydantic JSON schema outputs.

### Additional notes

While I sorted dictionaries within each model so that the keys of each model would be in the same order as the pydantic-based scripts' exports, at the same time, **I decided not to sort the top-level keys** (`$defs/ModelKey`). Because doing so would make the diff ugly and harder to read the applied changes. The top-level key sorting can be done afterward.


darkhaniop:
@holtskinner, I believe yesterday I happened to open this PR and fight the commit-checker bots around the same time you were working on these workflows.

Even though my commit modified only one file, which has not changed recently, my PR branch shows up as "out-of-date with the base branch." Do you recommend leaving it like this for the reviewers, or should I merge the `main` branch?

darkhaniop:
I merged recent fixes to the spec file from `main`, so the diff became cleaner. Also, I created two issues (#353 and #354) to explain separate problems that this PR addresses.

Please let me know if you have any questions.

holtskinner (approved):

specification/json/a2a.json
```diff
@@ -626,7 +629,6 @@
         "properties": {
           "code": {
             "const": -32600,
-            "default": -32600,
```
darkhaniop: To help with the review, let me explain why "default" values for error codes should not be in the spec.

The client should be able to validate received ErrorResponses using the A2A specification. However, the inclusion of the "default" codes leads to ErrorResponses with the missing `error.code` property to pass the validation (while such messages are invalid). So, this PR removes all "default" values from `error.code` definitions.

I think it is not as critical with the `eror.message` field, so I left them unchanged (to reduce the diff size).

specification/json/a2a.json
```diff
@@ -801,6 +798,10 @@
             "$ref": "#/$defs/JSONRPCError"
           }
         },
+        "required": [
+          "jsonrpc",
+          "id"
+        ],
```
darkhaniop: Every JSON RPC response object must include the `id` property, which is used to correlate the response object with the corresponding request.

specification/json/a2a.json
```diff
@@ -922,7 +921,6 @@
           },
           "method": {
             "const": "tasks/send",
-            "default": "tasks/send",
```
darkhaniop: The situation with default values of the `method` property is similar to `error.code` (see above). The A2A server that receives RPC requests, first, should be able to validate the requests; second, it should be able to differentiate the request method based on the `method` property (instead of substituting a default value if it is missing).

Therefore, the spec should not state "default" values for the `method` property.

darkhaniop:
I linked some comments to the code lines to explain the diffs.

specification/json/a2a.json
darkhaniop: These bugs seem to be addressed in #599. Later, I will edit the main message to reflect it.

types/src/types.ts
```diff
@@ -522,6 +522,7 @@ export interface JSONRPCMessage {
  * Represents a JSON-RPC 2.0 Request object.
  */
 export interface JSONRPCRequest extends JSONRPCMessage {
+  id: number | string;
```
kthota-g: JSONRPCRequest itself can omit id as per spec, but all the A2A Requests require an id specified. `id: number | string` should be specified for individual request types extending JSONRPCRequest.

types/src/types.ts
```diff
@@ -559,6 +560,7 @@ export interface JSONRPCError {
  * Represents a JSON-RPC 2.0 Result object.
  */
 interface JSONRPCResult extends JSONRPCMessage {
+  id: number | string;
```
kthota-g: result can also be null

darkhaniop: From my understanding, JSON-RPC spec discourages the usage of id=null for successful responses. Also, in `types.ts`, this particular interface represents the "successful" response, because it defines `error?: never;` and is used only as a parent for `*SuccessResponse` interfaces.

With that in mind, what do you think about renaming `JSONRPCResult` to `JSONRPCSuccessResponse` (to mirror the `JSONRPCErrorResponse` interface)?

darkhaniop: If the answer to the above is positive, in the renamed `JSONRPCSuccessResponse`, do you think we should still add "null" as an allowed value for "id"?

types/src/types.ts
darkhaniop: Moved to individual `*Request` interfaces in f53206b399373c034e533d1d02cb76560ce1f805.

types/src/types.ts
darkhaniop: In 244dff7a88e62f4cf1136fa65e7baba261b3f4e2, `JSONRPCResult` renamed to `JSONRPCSuccessResponse`. So, the "id" prop in parent response interfaces is now defined as:
```typescript
interface JSONRPCSuccessResponse extends JSONRPCMessage {
  id: number | string;
  // ...
}

export interface JSONRPCErrorResponse extends JSONRPCMessage {
  id: number | string | null;
  // ...
}
```

kthota-g: Thanks for the changes. Though unlikely for id to be null in success cases, the rpc response object regardless of error/result, can have id set to null if request.id can't be read. So, let's go with `number | string | null` for both these cases.

Also, I see a2a.json did not pick up the `null` type for JSONRPCErrorResponse.id. I will take a look into this issue.

darkhaniop: Looking at my own comment, I realized that `JSONRPCErrorResponse` is exported while `JSONRPCSuccessResponse` is not. 8dad30753f111104361db01d9f57ab5c6adcd7dd, adds the "export" statement (only for the cosmetic effect, the generated spec does not change from this).
BTW, thanks for reviewing the PR.

darkhaniop: > Thanks for the changes. Though unlikely for id to be null in success cases, the rpc response object regardless of error/result, can have id set to null if request.id can't be read. So, let's go with `number | string | null` for both these cases.

Updated in 47c4f716d5fc37a0a4bed1af1794126280bd5ae6.

kthota-g: The null type seems to require specifying an annotation so the generator can pick up the change. Adding the following where id can be null should hint the generator to add null type.

```
/**
   * An identifier established by the Client that MUST contain a String, Number.
   * Numbers SHOULD NOT contain fractional parts.
   * @nullable true
   */
```

darkhaniop: Added in 6ee3b309131fee626081d6a2bccfd7f6fcef41eb.

kthota-g: Thanks @darkhaniop. I will work on changes required in samples and other components and will merge this into main.

darkhaniop: > Thanks @darkhaniop. I will work on changes required in samples and other components and will merge this into main.

Sure, and thanks for checking the output. I'm glad we also caught other small mistakes (such as this last one with nulls).

darkhaniop:
Just noticed one more minor typo in a docstring (only affects the "description" prop of GetTaskResponse). In https://github.com/google/A2A/blob/6ee3b309131fee626081d6a2bccfd7f6fcef41eb/types/src/types.ts#L663-L667

Can be fixed as:
```diff
 /**
- * JSON-RPC success response for the 'tasks/get' method.
+ * JSON-RPC response for the 'tasks/get' method.
  */
 export type GetTaskResponse = GetTaskSuccessResponse | JSONRPCErrorResponse;
```

*(I did not apply this fix, in case there are already offline changes on top the latest commit 6ee3b309131fee626081d6a2bccfd7f6fcef41eb)*

PR #352, opened 2025-04-30
Remove 'null' support for JSON Schema types

### LINKED ISSUE #264: Suggestion: a2a.json schema need not have null types as an option for non-required properties
I noticed a repeated pattern
`
"anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
`
When the property isn't required, it seems cleaner for the user to omit it rather than giving  them the option specify a null.

whitlockjc:
This commit also makes running the A2A specification tests much simpler.

Fixes #264 🦕

### DIFF
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -13,15 +13,7 @@
             "type": "array"
           },
           "credentials": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "null"
-              }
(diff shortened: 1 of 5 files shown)


whitlockjc:
I'll work on the lint changes, but not sure they all apply. Seems like lint was configured after this file was added.

kthota-g (approved):

PR #375, opened 2025-05-02
feat: adding support for authorization at http layer (Python)

didier-durand:
# Description

I re-use the code that you supplied A2A/samples/python/common as-is for my use case.

The production env in which I work is thoroughly protected: our http gateway requires request signature.

For this purpose, I had to integrate an additional package `httpx_auth` (see https://pypi.org/project/httpx-auth/ and https://github.com/Colin-b/httpx_auth)

I also needed to change client.py  to accept this auth signer: see commit diff for details. This changed code allows to go through our http gateway (tests are fine!)

Please, accept my PR so that I can keep using the reference code with no patch on my side.

**Note**: I may come back with another for the SSE part if needed when I am there in my implementation

Some similar update may have to be done for Javascript but I don't program with it.

Didier

- [X] Follow the [`CONTRIBUTING` Guide](https://github.com/google/A2A/blob/main/CONTRIBUTING.md).
- [X ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [X ] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format).

### DIFF
--- a/samples/python/common/client/client.py
+++ b/samples/python/common/client/client.py
@@ -23,13 +23,14 @@ import json
 
 
 class A2AClient:
-    def __init__(self, agent_card: AgentCard = None, url: str = None):
+    def __init__(self, agent_card: AgentCard = None, url: str = None, auth=None):
         if agent_card:
             self.url = agent_card.url
         elif url:
             self.url = url
         else:
             raise ValueError("Must provide either agent_card or url")
+        self.auth=auth
 
     async def send_task(self, payload: dict[str, Any]) -> SendTaskResponse:
         request = SendTaskRequest(params=payload)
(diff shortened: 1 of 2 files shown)


holtskinner:
Please move this PR to the the SDK repository https://github.com/google/a2a-python/ if these changes haven't already been made in the SDK.

didier-durand:
Hi,

I checked the code of the SDK: in its current form, I should be able to implement my use case all by myself thanks to the addition of **kwargs when instantianting the httpx. The kwargs allow to add any httpx param you need: I'll be able to use `auth` that I was requesting in my previous PR.

The new code is now really flexible! Well done.

PR #397, opened 2025-05-03
feat: A2A-MCP Connector: Intuitive Bridge Between Agent-to-Agent Protocol and Model Context Protocol

sms03:
# Description

A simple agent that connects Agent-to-Agent (A2A) protocol and Model Context Protocol (MCP) with ease, providing an intuitive and easy-to-use interface. Built with Google ADK (Agent Development Kit).

Features
Easy MCP Tool Registration: Register MCP tools with a simple interface
Tool Discovery: List and discover registered MCP tools
Unified Access: Call any registered MCP tool through a consistent interface
A2A-Compatible: Fully compliant with the A2A protocol for seamless interoperability
Streaming Support: Provides streaming updates during task processing
Persistent Tool Registry: Tools can be saved to disk and loaded on restart
Real MCP Integration: Actually connects to and calls MCP tools
Browser-Friendly Interface: Helpful web UI when accessed via browser


sms03:
resolved minor conflicts.

samples/python/agents/a2a_mcp_connector/agent.py
```diff
@@ -0,0 +1,291 @@
...
+    except Exception as e:
+        logger.error(f"Error saving MCP tool registry: {e}")
+
+def register_mcp_tool(tool_id: str, tool_url: str, tool_description: str) -> Dict[str, Any]:
+    """
+    Register an MCP tool with the connector.
+    
+    Args:
+        tool_id (str): Unique identifier for the MCP tool
+        tool_url (str): URL endpoint for the MCP tool
```
bobtista: If I understand the [spec](https://modelcontextprotocol.io/specification/2025-03-26/server/tools) correctly, MCP doesn't use endpoints per tool, it uses JSON RPC requests to the MCP server (whether transport is STDIO or SSE). You can see its use in the MCP SDKs eg [here](https://github.com/modelcontextprotocol/python-sdk/blob/ed25167fa5d715733437996682e20c24470e8177/src/mcp/server/stdio.py#L12)

bobtista: I saw your new commit, I'm currently on mobile app, apologies for the simple comment - did an AI write this commit? It's unclear to me how your mcp server/client are using either SSE or STDIO transport. The tool calling still seems to have a post request to a tool url. It seems to me like an ai hallucination.

sms03: Hi bobtista, nope AI didnt write the commit. It's not an AI hallucination but rather a distinction between the transport for the A2A protocol's own messages and the transport used for invoking external MCP-compliant tools. The A2A server acts as a client to the MCP tool. 
Please do let me know what do i need to do from your perspective?

bobtista: I'm still not at my computer, maybe this is a naming thing? An agent can connect to a list of A2A agents, grabbing each one's card. A given agent's mcp client can connect to a list of mcp servers - each server can return a list of tools. Agents should have their mcp clients make requests to a connected mcp server using SSE or STDIO when they want to use one of its tools.
In your pr, I'm seeing code where it is trying to connect to a list of tools by url, and then trying to call those tools directly, also by url. Is this right? I can review better when I'm back home.

bobtista: Yeah, if I understand correctly, the code here is trying to use each tool like it's a micro service with its own URL, instead of using the MCP server. The agents using A2A can give each other tasks in plain english. Agents can use tools provided by connected MCP servers. 
Think of it as something like this:
Agent A asks Agent B "Hey what are your skills?" via A2A
"I'm good at getting the weather" 
Agent A asks Agent B "Ok what's the weather today in Paris?"
Agent B is connected to a weather MCP server.
Agent B submits a request to the weather MCP server to use its "get_weather" tool. (It got the list of available tools using the "tools/list" method)
Agent B gets the response from the weather MCP server and uses it to return the result to agent A as an artifact, completing the initial task.

bobtista: I'd probably try to use an existing MCP client implementation. Import them from the mcp repo [sse](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/sse.py) [stdio](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/stdio/__init__.py#L97)
Does this make sense?

sms03: I understand what you've said here, I've done few changes please if you would look into it I'd appreciate it. Thanks

bobtista: It's progress :) You don't need to create your own json rpc requests manually if you're using the imported MCP client - they are handling the json rpc for you. Also, you'd want to choose STDIO or SSE as the transport type for a given MCP server, and you don't need to make a new connection and list all tools on a per tool call basis. Finally, I'm still seeing code where you're trying to explicitly call a tool by its URL, and there still seems to be confusion about an MCP tool vs an MCP server. I'd recommend reading through he MCP docs, reviewing code from some sample MCP servers and clients, and using them locally eg in Claude Desktop to get a solid understanding. 

You may find that the A2A-MCP Connector concept isn't really a thing. An agent can use tools with MCP. It can also interact with other agents with A2A. They're both just kind of add ons to an agent. It seems to me that what you're making is an agent that is both A2A compliant and has an MCP client ready to use. You can make an example agent that is A2A compatible, and uses an MCP server eg the [git mcp server](https://github.com/modelcontextprotocol/servers/tree/main/src/git). This agent would be able to receive task requests from an A2A client in plain english like "Hey can you look at whatever files have changed and commit them with a brief commit message?" Or something more long lived, like if your commit automatically runs tests that take an hour, the A2A agent could let you know if the commit was rejected or not. The agent would use the tools that the git mcp server provides under the hood to accomplish these tasks. Does this all make sense?

sms03: I have made the changes see if it fits. Please do let me know, thanks for your time.

holtskinner:
Can you edit this PR to use the SDK from https://github.com/google/a2a-python/

samples/python/agents/a2a_mcp_connector/agent.py
bobtista: I'm trying to be helpful here and trust what you're saying, but this really does feel like I'm interacting with an AI in Cursor - or at least that you're lying that AI didn't write this code. It seems to me that there are some core misunderstandings - I really recommend asking clarifying questions and reading/understanding the docs before pushing changes and asking for more reviews. I would consider pausing on this PR and getting caught up to speed on how these things work, particularly with MCP - follow the [three tutorials here
](https://modelcontextprotocol.io/quickstart/user). You can make regular agents that use MCP from here - You can also use Pydantic's MCP implementation for agents, or use [Google's ADK](https://google.github.io/adk-docs/tools/mcp-tools/#step-1-attach-the-mcp-server-to-your-adk-agent-via-mcptoolset) Check out the "Key considerations" part in that link, I think it's relevant to your Connector concept. In this case, if you want to continue, consider using Google's new [A2A python SDK ](https://github.com/google/a2a-python/)

sms03: to be honest i have used and taken help of AI, but trust me code isnt AI generated plus i dont even use Cursor, I'll look into it what i can do to follow the needs.

benclarkeio:
Although this pull request has been closed. For future reference, this is the most feature-rich and up-to-date [A2A MCP Server](https://github.com/a2anet/a2a-mcp).

PR #169, opened 2025-04-15
Export google_a2a as a python module instead of common and agents

djsamseng:
Sets up the file structure correctly to allow for the modules to be imported with a proper namespace.

Previously developers would have to
```python
import common
```

Now developers can
```python
import google_a2a.common
```

This will allow us to further enhance this module to become production ready in the future. And in the meantime it allows those of us experimenting with A2A to avoid namespace clashes.

See the samples/python/README.md for an example setup.

Addresses the issues raised [in this comment](https://github.com/google/A2A/issues/148#issuecomment-2802282457)

Test Plan:
- Note I don't have a GOOGLE_API_KEY so I couldn't fully run, but it exit cleanly thus verifying the only changes (the top level imports)
```bash
uv run src/google_a2a/agents/langgraph
uv run src/google_a2a/hosts/cli
```
- Additionally tested importing google_a2a from another project via `uv add /path/to/A2A/samples/python` and was able to `import google_a2a.common.server` and create and run an `A2AServer`. This verifies the server does in fact run.

PR #408, opened 2025-05-05
fix(docs/tutorials/python): Update a2a imports.

daavoo:
The current imports don't work after following the project setup.

- [X] Follow the [`CONTRIBUTING` Guide](https://github.com/google/A2A/blob/main/CONTRIBUTING.md).
- [X] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [X] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format).
- [X] Appropriate docs were updated (if necessary)

### DIFF
--- a/docs/tutorials/python/4-agent-skills.md
+++ b/docs/tutorials/python/4-agent-skills.md
@@ -42,8 +42,7 @@ This conforms to the skills section of the [Agent Card](documentation?id=represe
 Let's create this Agent Skill in code. Open up `src/my-project/__init__.py` and replace the contents with the following code
 
 ```python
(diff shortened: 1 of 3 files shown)


djsamseng (commented):
Thanks. I think it would be preferable to merge https://github.com/google/A2A/pull/169 but if not then we should also update the instructions for running the cli to require cloning the entire A2A repository

daavoo:
> Thanks. I think it would be preferable to merge #169 but if not then we should also update the instructions for running the cli to require cloning the entire A2A repository

I missed that PR!

PR #415, opened 2025-05-05
feat: Protocol updates: Support stateless transactions

pstephengoogle:
# Protocol updates

This PR makes some clarifications around several concepts and intends to make simple 'chit-chat' like scenarios, or stateless interactions, easier. In particular:

- Introduces a messageId that is created by the message creator (both client and server)
- Clarifies the ownership of a taskId. This is always determined by the server (if/when a task is needed). Client should send follow up messages with the relevant taskId to continue interactions around an existing task
- Changes the name 'sessionId' to 'contextId' and allows it to be attached to messages and tasks. The server is responsible for creating the contextId and the client is responsible for using it to continue interactions with shared context.
- Introduces a new message/send endpoint. This is parallel to the task/send endpoint (which is deprecated as of this PR and will be removed in a future PR). This endpoint returns either a Message object (if there is not state/task to track) or a task object. If the request is blocking, it will only return when the task is in a suspended state. If the request is non-blocking, it will return the status of the initial task creation. The client can then poll for task updates.
- Introduces a new message/stream endpoint. This is parallel to the task/sendSubscribe endpoint (which is deprecated as of this PR and will be removed in a future PR). The endpoint has the following expectations:
   - If the response is a simple Message with no statefulness, returns a Message and the channel should be closed.
   - Otherwise, the first thing returned is an initial task object with the 'submitted' state. Each subsequent event is a TaskStatusUpdateEvent or TaskArtifactUpdateEvent until the final TaskStatusUpdateEvent is received.

The demo UI app and the cli app have been updated to use the new endpoints, and no longer use the old ones. The biggest delta is that the new endpoints create the task and context identifiers when the remote agent deems it necessary, and the client is not responsible (nor should) generate these itself. Also, all messages are required to have a messageId with a value created by the sender. To continue a multi-turn interaction the same contextId must be used in subsequent messages send from the client to the server. To continue an interaction within the scope of a task, the contextId and taskId must be included in the message sent from the client.


docs/tutorials/python/9-ollama-agent.md
```diff
@@ -102,21 +102,22 @@ class MyAgentTaskManager(InMemoryTaskManager):
     else:
       self.ollama_agent = None
 
-  async def on_send_task(self, request: SendTaskRequest) -> SendTaskResponse:
+  async def on_send_message(self, request: SendMessageRequest) -> SendMessageResponse:
     # ...
-    received_text = request.params.message.parts[0].text
+    received_text = self.get_user_unput(request.params)
```
ToddSegal: typo

docs/documentation.md
```diff
@@ -146,7 +146,7 @@ A task can be completed by a remote agent immediately or it can be long-running.
 
 A Task is a stateful entity that allows Clients and Remote Agents to achieve a specific outcome and generate results. Clients and Remote Agents exchange Messages within a Task. Remote Agents generate results as Artifacts.
 
-A Task is always created by a Client and the status is always determined by the Remote Agent. Multiple Tasks may be part of a common session (denoted by optional sessionId) if required by the client. To do so, the Client sets an optional sessionId when creating the Task.
+A Task is always created by the Remote Agent and the status is always determined by the Remote Agent. Multiple Tasks may be part of a common session (denoted by optional contextId) if required by the client. The Client sends a request to the Remote Agent and provides an optional contextId. The Remote Agent then correlates the request with the other contextual information for that id and decides if it is creating a Task or simply responds with a Message.
```
ToddSegal: do we need to clarify the definition of 'session'? IMO its a purely virtual thing.

docs/documentation.md
```diff
@@ -164,7 +164,7 @@ Tasks are used to transmit [Artifacts](#artifact) (results) and [Messages](#mess
 ```typescript
 interface Task {
   id: string; // unique identifier for the task
-  sessionId: string; // client-generated id for the session holding the task.
+  contextId: string; // server-generated id for the session holding the task.
```
ToddSegal: same comment about 'session' definition.

docs/documentation.md
```diff
@@ -189,15 +189,19 @@ interface TaskArtifactUpdateEvent {
...
-  message: Message;
-  historyLength?: number; //number of recent messages to be retrieved
-  // where the server should send notifications when disconnected.
+// Configuration of the send message request.
+interface MessageSendConfiguration {
+  acceptedOutputModes: string[];
+  historyLength?: number;
   pushNotification?: PushNotificationConfig;
-  metadata?: Record<string, any>; // extension metadata
+  blocking?: boolean;
```
ToddSegal: make sure to add appropriate documentation for this new flag.

docs/documentation.md
```diff
@@ -189,15 +189,19 @@ interface TaskArtifactUpdateEvent {
...
+}
+// Send by the client to the agent as a request. May create, continue or restart a task.
+interface MessageSendParams {
+  id: string;  // identifier of request. Client generated.
```
ToddSegal, line 201: why is id in the sendparams and not embedded in the message itself?

docs/documentation.md
```diff
@@ -240,6 +244,9 @@ interface Message {
   role: "user" | "agent";
   parts: Part[];
   metadata?: Record<string, any>;
+  messageId: string; // identifier created by the message creator.
+  taskId?: string; // identifier of task the message is related to, optional.
```
ToddSegal, line 248: are there times where taskId is required? e.g. when this message is the response to an `Input-Required` request from the remote agent?

docs/tutorials/python/9-ollama-agent.md
pstephengoogle: fixed

docs/documentation.md
pstephengoogle: It is only required when the client it trying to communicate against an existing task. So in this case, they should send it to respond to the task. If they don't, the server won't know that the message is in service of the task, and will assume it is potentially a new task

samples/python/common/client/client.py
```diff
@@ -62,12 +75,28 @@ async def send_task_streaming(
...
+        request = SendMessageStreamRequest(params=payload)
+        with httpx.Client(timeout=None) as client:
+            with connect_sse(
+                client, 'POST', self.url, json=request.model_dump()
+            ) as event_source:
+                try:
+                    for sse in event_source.iter_sse():
+                        yield SendMessageStreamResponse(**json.loads(sse.data))
+                except json.JSONDecodeError as e:
+                    raise A2AClientJSONError(str(e)) from e
```
swapydapy, line 90: This would be a server error, right? The decode failed at "json.loads(sse.data)"

docs/documentation.md
pstephengoogle: There is one in the message itself. I am unsure if we need this one, was really following the structure of other objects. Happy to remove this field if needed (can be a follow on PR)

docs/documentation.md
pstephengoogle: PTAL at updated PR

docs/documentation.md
pstephengoogle: PTAL at update PR

docs/documentation.md
pstephengoogle: done

samples/python/common/types.py
```diff
@@ -237,13 +284,17 @@ class TaskResubscriptionRequest(JSONRPCRequest):
...
 A2ARequest = TypeAdapter(
     Annotated[
-        SendTaskRequest
-        | GetTaskRequest
-        | CancelTaskRequest
-        | SetTaskPushNotificationRequest
-        | GetTaskPushNotificationRequest
-        | TaskResubscriptionRequest
-        | SendTaskStreamingRequest,
+        Union[
```
swapydapy: Using shorthand is recommended https://docs.python.org/3/library/stdtypes.html#union-type => 
"The union type expression enables cleaner type hinting syntax compared to [typing.Union](https://docs.python.org/3/library/typing.html#typing.Union)." Same for Google style guide: [go/pystyle#none-type](https://engdoc.corp.google.com/eng/doc/devguide/py/totw/035.md?cl=head)

Same goes for list vs List. I think "list" is preferred.

samples/python/common/server/task_manager.py
```diff
@@ -297,3 +422,98 @@ async def dequeue_events_for_sse(
             async with self.subscriber_lock:
                 if task_id in self.task_sse_subscribers:
                     self.task_sse_subscribers[task_id].remove(sse_event_queue)
+
+    def _extract_task_and_context(
```
swapydapy, line 426: Can we improve the name and maybe split these into two methods.

get_or_create_task_id()
get_or_create_context_id()

samples/python/common/types.py
```diff
@@ -133,21 +140,41 @@ class TaskQueryParams(TaskIdParams):
...
+    acceptedOutputModes: Optional[List[str]] = None
+    historyLength: Optional[int] = None
+    # If blocking is true, push notification is ignored. Both of these
+    # options are ignored in streaming interactions.
+    pushNotification: Optional[PushNotificationConfig] = None
+    blocking: Optional[bool] = None
+
+
+class MessageSendParams(BaseModel):
+    id: str
```
swapydapy, line 165: As pointed out by Todd as well, since message has its own ID, not sure why we need this "id" anymore.

samples/python/hosts/multiagent/remote_agent_connection.py
```diff
@@ -32,78 +28,35 @@ def __init__(self, agent_card: AgentCard):
...
+                # In the case a message is returned, that is the end of the interaction.
+                if isinstance(response.result, Message):
+                    return response
+
+                # Otherwise we are in the Task + TaskUpdate cycle.
+                if task_callback and response.result:
                     task = task_callback(response.result, self.card)
                 if hasattr(response.result, 'final') and response.result.final:
                     break
+                if not response.result:
```
swapydapy: nit: can we move it before "if task_callback and response.result:". That simplifies the mental flow and even the subsequent if condition.

samples/python/hosts/multiagent/host_agent.py
```diff
@@ -116,8 +122,8 @@ def before_model_callback(
     ):
         state = callback_context.state
         if 'session_active' not in state or not state['session_active']:
-            if 'session_id' not in state:
-                state['session_id'] = str(uuid.uuid4())
+            # if 'context_id' not in state:
```
swapydapy: is this commented code not needed?

samples/python/hosts/multiagent/host_agent.py
```diff
@@ -151,45 +157,41 @@ async def send_task(
...
+            ),
+            configuration=MessageSendConfiguration(
+                acceptedOutputModes=['text', 'text/plain', 'image/png'],
             ),
-            acceptedOutputModes=['text', 'text/plain', 'image/png'],
-            # pushNotification=None,
-            metadata={'conversation_id': sessionId},
         )
-        task = await client.send_task(request, self.task_callback)
+        task = await client.send_message(request, self.task_callback)
```
swapydapy: nit: lets rename the variable to response, it says "task" but then we check if it is type of message.

samples/python/agents/google_adk/agent.py
```diff
@@ -17,11 +15,12 @@
 
 
 def create_request_form(
-    date: str | None = None,
-    amount: str | None = None,
-    purpose: str | None = None,
+    date: Optional[str] = None,
```
swapydapy: Nit: Using str  | None is preferred in newer versions of python.

docs/documentation.md
```diff
@@ -146,7 +146,7 @@ A task can be completed by a remote agent immediately or it can be long-running.
 
 A Task is a stateful entity that allows Clients and Remote Agents to achieve a specific outcome and generate results. Clients and Remote Agents exchange Messages within a Task. Remote Agents generate results as Artifacts.
 
-A Task is always created by a Client and the status is always determined by the Remote Agent. Multiple Tasks may be part of a common session (denoted by optional sessionId) if required by the client. To do so, the Client sets an optional sessionId when creating the Task.
+A Task is always created by the Remote Agent and the status is always determined by the Remote Agent. Multiple Tasks may be part of a common coherent set of interactions (denoted by optional contextId) if required by the client. The Client sends a request to the Remote Agent and provides an optional contextId. The Remote Agent then correlates the request with the other contextual information for that id and decides if it is creating a Task or simply responds with a Message.
```
swapydapy, line 149: How about if we rephrase this to better clarify the contextId and Tasks.

- A Task is always created by the Remote Agent and the status is always determined by the Remote Agent.
- Client can bundle the communication to an agent within a context by specifying a contextId. All tasks created by the remote agent will be associated to that contextId.
- This contextId is optional. If not specified for a request, a fresh contextId is created by the Remote Agent.
- Remote Agent correlates the request with the other contextual information for that id and decides if it is creating a Task(s), resuming a task or simply responds with a Message.

swapydapy (approved):

samples/python/common/client/client.py
pstephengoogle: I am following the existing patterns already in place

samples/python/hosts/multiagent/host_agent.py
pstephengoogle: fixed

samples/python/hosts/multiagent/host_agent.py
pstephengoogle: updated

samples/python/agents/google_adk/agent.py
pstephengoogle: fixed

docs/documentation.md
pstephengoogle: I was trying to stay with the existing formatting. Lets get the support teams to help with the content rewrites

samples/python/common/server/task_manager.py
pstephengoogle: We can do that in the SDK. We don't have any examples of just context (and task won't exist without context)

samples/python/common/types.py
pstephengoogle: Agreed, will do in another PR

samples/python/hosts/multiagent/remote_agent_connection.py
pstephengoogle: done

samples/python/hosts/multiagent/host_agent.py
pstephengoogle: removed

samples/python/common/types.py
pstephengoogle: updated

PR #417, opened 2025-05-06
feat: This is an A2A-compatible agent that uses Claude 3 Sonnet to interact with postgres databases through the MCP (Model Context Protocol) 

barnendu:
MCP SQL Agent

This is an A2A-compatible agent that uses Claude 3 Sonnet to interact with postgres databases through the MCP (Model Context Protocol).

Features

MCP Integration
SQL database operations and queries
Streaming responses
Push notifications
A2A protocol compatibility


holtskinner:

PR #419, opened 2025-05-06
Update README.md

palladius:
typo in docs

# Description

- [ ] Follow the [`CONTRIBUTING` Guide](https://github.com/google/A2A/blob/main/CONTRIBUTING.md).
- [ ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [ ] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Watch [this demo video](https://storage.googleapis.com/gweb-developer-goog-blog-
 
 The Agent2Agent (A2A) protocol facilitates communication between independent AI agents. Here are the core concepts:
 
-- **Agent Card:** A public metadata file (usually at `/.well-known/agent.json`) describing an agent's capabilities, skills, endpoint URL, and authentication requirements. Clients use this for discovery.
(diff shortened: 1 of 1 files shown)


holtskinner (approved):

wilsonsilva:
Is this a typo? 

> They are [URL](https://en.wikipedia.org/wiki/URL) path prefixes that start with /.well-known/.
> - https://en.wikipedia.org/wiki/Well-known_URI

> This memo defines a path prefix for "well-known locations",
   "/.well-known/", in selected Uniform Resource Identifier (URI)
   schemes.
> - https://datatracker.ietf.org/doc/html/rfc8615

> A well-known URI is a URI [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)] whose path component begins with
   the characters "/.well-known/", provided that the scheme is
   explicitly defined to support well-known URIs.
>
>   For example, if an application registers the name 'example', the
   corresponding well-known URI on 'http://www.example.com/' would be
   'http://www.example.com/.well-known/example'.
> - https://datatracker.ietf.org/doc/html/rfc8615#section-3

palladius:
OMG I'm so sorry, I felt it was a typo. Let's remove this!

PR #456, opened 2025-05-08
AgentCard Auth: New AgentCard field: supportsAuthenticatedExtendedCard

dmandar:
# Description
Part 1 of enabling authenticated agent cards. Add a new optional field to the Agent card 'supportsAuthenticatedExtendedCard' that, if enabled, allows client to download an expanded version of the agent card post authentication from a new '/authenticatedExtendedCard' endpoint.
Update: spec tests now pass


mikeas1 (approved):
Also work with Holt to get documentation updated.

PR #460, opened 2025-05-08
ci: Change Linters to only check `docs/` and `specification/` folders

holtskinner:


surferkid333 (commented):
Shower sweet  least to understand

PR #462, opened 2025-05-08
feat: Support named message parts via optional 'name' field

### LINKED ISSUE #461: [Feat]:

Yes. When using the A2A protocol to send tasks with multiple `Part` elements in a message (e.g., a `TextPart` for a query and a `FilePart` for an image), there is currently no standardized way to name or label these parts. This leads to ambiguity and requires agents to rely on the order of parts or use informal conventions via metadata to interpret inputs correctly.

This becomes especially problematic when:

- Multiple parts of the same type are sent (e.g., several `TextPart`s),
- Agents are built to work with named parameters (e.g., `query`, `context`, `image`,`channel`),
- Parts are dynamically generated or added in variable order,
- Developers want to build stricter contracts or validation logic for agent inputs.

While it is possible to include a name or similar identifier in the `metadata` field of each part, this approach lacks consistency and discoverability. Having a dedicated `name` field would make input handling clearer, easier to validate.

Without a naming mechanism, the protocol lacks a robust way to map inputs to their intended semantic meaning.

I'd like to propose extending the Part model (e.g., `TextPart`, `FilePart`, etc.) to optionally include a `name: str `field.

This would allow each part to be explicitly labeled (e.g., `name="query"`, `name="invoice"`) and give agents a consistent and reliable way to reference expected inputs.

Example:

```
{
  "type": "text",
  "text": "What's the total from this invoice?",
  "name": "query"
},
{
  "type": "file",
  "file": {
       "mimeType": "application/pdf",
       "bytes": "<base64-encoded-content>",
       "uri": "gs://some_bucket/invoice.pdf"
  },
  "name": "invoice"
}
```

This change would improve readability and debuggability, reduce reliance on ordering, enable stronger validation and agent design and be backward-compatible if name is made optional.

As mentioned above, part names can be added to `metadata` field, but this is inconsistent and not standardized. A dedicated `name` field would make input handling clearer and validation more straightforward.

### Additional context

_No response_

### Code of Conduct

inesmcm26:
# Description
This pull request introduces an optional `name` field to all `Part` types (`TextPart`, `FilePart`, `DataPart`) within the A2A protocol.

### Problem Addressed:

Currently, identifying parts in a `TaskSendParams` message relies either on their position or on custom conventions using the `metadata` field. While using `metadata` is technically possible, it lacks consistency, standardization, and discoverability.

This is particularly problematic when:

- Multiple parts of the same type are included.
- Agents are designed to consume named parameters of the same type (e.g., `query: str` and `context: str`)
- Parts are dynamically generated by the request sender or appear in variable order.

### Proposed Solution:

Introducing a dedicated `name` field provides a standardized, first-class way to label parts. This is more explicit, easier to validate, and more consistent across implementations than using `metadata`. It also helps avoid mismatches between clients and agents, and makes it simpler for task managers to map parts to agent invocation inputs.

### Implementation Details:

- Updated JSON schema definitions to include an optional name field in each Part type.
- Ensured backward compatibility by making the field optional.
- Updated the protocol documentation to reflect the new `name` field and its usage.

### Benefits:

- Improved clarity in `Part`s structures
- Enables request parts validation

Fixes #461

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -250,10 +250,12 @@ A fully formed piece of content exchanged between a client and a remote agent as
 ```typescript
 interface TextPart {
   type: "text";
+  name?: string;
   text: string;
 }
 interface FilePart {
   type: "file";
+  name?: string;
   file: {
     name?: string;
     mimeType?: string;
@@ -265,6 +267,7 @@ interface FilePart {
 }
 interface DataPart {
   type: "data";
+  name?: string;
   data: Record<string, any>;
 }
 type Part = (TextPart | FilePart | DataPart) & {
--- a/docs/specification/agent-to-agent-communication.md
+++ b/docs/specification/agent-to-agent-communication.md
@@ -114,10 +114,12 @@ A fully formed piece of content exchanged between a client and a remote agent as
 ```typescript
 interface TextPart {
   type: "text";
+  name?: string;
   text: string;
 }
 interface FilePart {
   type: "file";
+  name?: string;
   file: {
     name?: string;
     mimeType?: string;
@@ -129,6 +131,7 @@ interface FilePart {
 }
 interface DataPart {
   type: "data";
+  name?: string;
   data: Record<string, any>;
 }
 type Part = (TextPart | FilePart | DataPart) & {
--- a/docs/specification/details.md
+++ b/docs/specification/details.md
@@ -183,6 +183,7 @@ Represents a distinct piece of content within a `Message` or `Artifact`. A `Part
 | Field Name | Type               | Required | Description              |
 | :--------- | :----------------- | :------- | :----------------------- |
 | `type`     | `"text"`           | Yes      | Identifies this as text. |
+| `name`     | `string`           | No       | The part name.           |
 | `text`     | `string`           | Yes      | The textual content.     |
 | `metadata` | `object` \| `null` | No       | Part-specific metadata.  |
 
@@ -191,6 +192,7 @@ Represents a distinct piece of content within a `Message` or `Artifact`. A `Part
 | Field Name | Type                                    | Required | Description                     |
 | :--------- | :-------------------------------------- | :------- | :------------------------------ |
 | `type`     | `"file"`                                | Yes      | Identifies this as a file.      |
+| `name`     | `string`                                | No       | The part name.                  |
 | `file`     | [`FileContent`](#66-filecontent-object) | Yes      | Contains the file details/data. |
 | `metadata` | `object` \| `null`                      | No       | Part-specific metadata.         |
 
@@ -199,6 +201,7 @@ Represents a distinct piece of content within a `Message` or `Artifact`. A `Part
 | Field Name | Type               | Required | Description                                                     |
 | :--------- | :----------------- | :------- | :-------------------------------------------------------------- |
 | `type`     | `"data"`           | Yes      | Identifies this as structured data.                             |
+| `name`     | `string`           | No       | The part name.                                                  |
 | `data`     | `object`           | Yes      | The structured JSON data payload (e.g., form data, parameters). |
 | `metadata` | `object` \| `null` | No       | Part-specific metadata.                                         |
 
--- a/llms.txt
+++ b/llms.txt
@@ -98,12 +98,15 @@ Key features of the A2A protocol highlighted by the specification and samples in
 *   **`Part` (Union Type):** Represents a piece of content within a Message or Artifact.
     *   **`TextPart`:**
         *   `type`: "text"
+        *   `name`: (string) The part name
         *   `text`: (string) The textual content.
     *   **`FilePart`:**
         *   `type`: "file"
+        *   `name`: (string) The part name
         *   `file`: (`FileContent`) File details (bytes or URI).
     *   **`DataPart`:**
         *   `type`: "data"
+        *   `name`: (string) The part name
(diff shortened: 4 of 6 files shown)


holtskinner:
Sorry about closing this PR, this should have been addressed. @pstephengoogle @kthota-g Can one of you review and see if this should be added to the protocol?

holtskinner:
@inesmcm26 Can you rebase from main and make your change in types/types.ts ?

inesmcm26:
Hi @holtskinner! I’ve rebased from main and applied the changes in `types/types.ts` as requested. Please let me know if you’d like me to update anything else

holtskinner (approved):

swapydapy:
Can we instead rename to partId? that way the purpose of a unique identifier is much more clearer. "Name" provides a human-readable string, and doesn't necessarily imply uniqueness.

inesmcm26:
> Can we instead rename to partId? that way the purpose of a unique identifier is much more clearer. "Name" provides a human-readable string, and doesn't necessarily imply uniqueness. 

Thanks for the suggestion @swapydapy! I get the concern about `name` not implying uniqueness, and I agree that `partId` would better communicate that.

My goal with `name` is to keep things intuitive and clear for developers working on agents and agent executors. Since agents are typically invoked using fixed parameters names like `query` or `context`, having human-readable labels for each part makes mapping them to the agent’s invocation arguments straightforward and unambiguous.
Something like a UUID-style `partId` would be harder to work with in practice.

Just to clarify: were you thinking of `partId` as something auto-generated and unique like a UUID, or more of a developer-defined label?

swapydapy:
Ah, I see. Shouldn't [DataPart](https://a2aproject.github.io/A2A/latest/specification/#653-datapart-object) fit that use case better? It was designed for the intention of named params or key-value pairs, acting as input to the agent.

swapydapy:
I was thinking about using partId from streaming perspective. Added a new discussion: #796

inesmcm26:
That's a fair question. Wrapping everything in a `DataPart` (even other nested parts) is possible, but I think it undermines the strength of the multi-part message protocol.

The real issue is clarity. A `TextPart` says “this is text,” a `FilePart` says “this is a file,” and a `DataPart` says “this is some arbitrary structured thing, go figure it out”. If you move everything into a `DataPart`, suddenly you need extra logic on the agent executor side just to figure out what’s what: is this a string? A nested part? A file reference? You lose clarity, and add parsing overhead.

With multi-part messages, you’re guaranteed a list of known typed parts. You can iterate over them, process them consistently, and the agent executor doesn’t have to guess or parse nested structures. Adding a `name` field to each part lets us preserve that clarity while also allowing agent executors to map message parts to agent inputs.

Does that make sense? What do you think about it?

swapydapy:
I think we are talking about two things:

1. Mapping message parts to agent inputs or outputs: This is something specific to an agent, and agent would advertise that it needs a specific message structure for input like that to work. At a general protocol level, I feel DataPart is more suited to send pre-defined & structured inputs to the agent. And since agent has already advertised that it needs what inputs it needs, client should sent that in message DataPart. There's no guessing required on the agent side, since agent already know which keys mean what. Same is the case for artifact, defined keys should be sent in DataPart.

2. Clear type of general message contents: This is exactly what part types are for. Allows the consumer (client or agent) of a message to know what is the type of the message part.

At a general level (without any pre-defined structure), for a multi-part message, consumer would need to infer the intent for each part.

So the question is, what is the purpose of part-name, possibly:
1. Semantic hint on the contents of the part. Helps the consumer to infer the intent of the part.
2. Human readable name.

Eg: https://github.com/a2aproject/A2A/discussions/796#discussioncomment-13606595

Artifact: Flight booking. Agent didn't pre advertise how it generates artifacts for a flight book. If they had, I would suggest, that stuff goes into DataPart.

Part A: (TextPart, name="summary") Summary of the flight, seats type etc.
Part B: (TextPart, name="directions") Directions on how to reach airport, how to hotel once your reach.
Part C: (FilePart, name="invoice") PDF invoice of the flight booking.
Part D: (FilePart, name="travel restrictions & baggage") baggage allowance, and restricted items
Part D: (DartPart) Ticket PNR, Time of flights, flight No, Seat No.

inesmcm26:
Thanks for the discussion, I do agree with you after all.

If an agent explicitly defines how it expects message parts, wrapping inputs inside a single `DataPart` makes sense. Naming message parts then requires the same effort as correctly structuring the `DataPart`. In this case, the agent should clearly announce its expected format, and the sender should follow it. I previously argued that wrapping everything in a `DataPart` reduced clarity and added parsing overhead, but I no longer hold that view: if expectations are clear, parsing one `DataPart` is no harder than handling multiple named parts.

But when no predefined structure exists, a `name` field on each part is still valuable as you pointed out. It adds semantic clarity and human-readable hints that help consumers infer intent without guessing. This is useful to to let the consumer quickly understand each part’s role, display meaningful labels for debugging and UX for example, etc.
Your example is perfect to show this. Having explicit names immediately communicate each part’s purpose.

Currently there’s no explicit way for an agent to announce exactly what input fields it expects - only general input/output modes are defined for agent skills. It might be worth extending the protocol to support this, allowing agents to declare expected input names and types in their agent card. I’ll create an issue to propose this.

darrelmiller:
Hi @inesmcm26  Based on your last comment, are you ok with closing this PR?

inesmcm26:
Yes, I'll close it! Thanks @darrelmiller

PR #467, opened 2025-05-09
Extended Agent Cards: Create a new private endpoint for extended agent cards

dmandar:
Create a new /agent/getAuthenticatedExtendedCard endpoint that returns an extended agent card. The actual additions to the agent card and the authentication are not part of the common code and will vary based on implementation.

### DIFF
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -1666,9 +1666,6 @@
         },
         {
           "$ref": "#/$defs/TaskResubscriptionRequest"
-        },
-        {
-          "$ref": "#/$defs/GetAuthenticatedExtendedCardRequest"
         }
       ],
       "title": "A2ARequest"


samples/python/common/client/client.py
```diff
@@ -42,7 +45,7 @@ def __init__(
             self.url = url
         else:
             raise ValueError('Must provide either agent_card or url')
-        self.timeout = timeout
+        self.default_headers = default_headers or {}
```
whitlockjc, line 48: What if instead we had a `self.credentials` or some client-level storage of credential _(and possibly it's related Security Scheme)_ and updated the `_send_request` to update the request prior to sending it, based off of the details in the client-level storage of credentials?

dmandar: yeah, that could formalize things for creds. I think there is value in exposing headers for other custom things like cache control, load balancing etc. I can have that as a separate enhancement PR

PR #471, opened 2025-05-09
feat: Add prototype version of Python SDK to A2A Repo

holtskinner:
# Description

Initial prototype of A2A Python SDK, for educational purposes only


koverholt (commented):
LGTM, verified that the prototype lib installs, I can run the agent, and ran test suite!

koverholt (approved):

PR #505, opened 2025-05-12
feat: adding support for Azure OpenAI in Semantic Kernel

oliverlabs:
# Description

Fixes google-a2a/a2a-samples#45  🦕


oliverlabs:
@swapydapy, could you please review?

oliverlabs:
@DJ-os perhaps you could review as well?

oliverlabs:
@didier-durand or perhaps you could review as you were the one who committed the semantic kernel sample. Thanks.

didier-durand:
Hi, sorry: I am not accredited to make reviews on this repo. Didier

samples/python/agents/semantickernel/.envexample
```diff
@@ -0,0 +1,9 @@
+## OpenAI configurations - uncomment if using OpenAI
+# OPENAI_API_KEY="<example>"
+# OPENAI_CHAT_MODEL_ID="gpt-4.1"
+#
+#
+## Azure OpenAI configuration - uncomment if using Azure OpenAI
+# AZURE_OPENAI_ENDPOINT=""
```
jland-redhat: Why make this this variable `AZURE_OPENAI_ENDPOINT` and not just `OPENAI_ENDPOINT`. Does not really matter if it is deployed through Azure, vLLM, etc.. it is all the same thing right?

didier-durand:
Hi @jland-redhat,  this article may help to understand the differences between the Azure endpoint for OpenAI and the native one.

https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints

samples/python/agents/semantickernel/.envexample
oliverlabs: The naming convention is per Microsoft documentation.

jland-redhat:
> Hi @jland-redhat, this article may help to understand the differences between the Azure endpoint for OpenAI and the native one.
> 
> https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints

Ok I missed the custom library.

So still have a few suggestions I would like to run by y'all.

There is a lot of commented code here which I am not in love with. 

What if instead we simplify the environment Variables. So instead of the current setup, what if we primarily stuck to `OPENAI_API_KEY` and `OPENAI_ENDPOINT`. (I see `AZURE_OPENAI_DEPLOYMENT_NAME` is being set to `deployment_name`, but it's unclear where this variable is actually used in the code.)

Including `OPENAI_ENDPOINT` also means that on  line 122 we could conditionally set that value if it exist. This would solve the issue for allowing us to connect to other LLMs that use the same specification as OpenAIChatCompletion such as the ones served through vLLM.


And for the Azure/Non-Azure Logic from what I can tell, the only code difference is on lines 137 and 155, where `service=AzureChatCompletion(` is called. We could introduce a single variable, like ENABLE_AZURE_OPENAI, and use it to conditionally leverage the Azure libraries in these two instances if it's set to True.

What are y'alls thoughts on that?

oliverlabs:
> > Hi @jland-redhat, this article may help to understand the differences between the Azure endpoint for OpenAI and the native one.
> > https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints
> 
> Ok I missed the custom library.
> 
> So still have a few suggestions I would like to run by y'all.
> 
> There is a lot of commented code here which I am not in love with.
> 
> What if instead we simplify the environment Variables. So instead of the current setup, what if we primarily stuck to `OPENAI_API_KEY` and `OPENAI_ENDPOINT`. (I see `AZURE_OPENAI_DEPLOYMENT_NAME` is being set to `deployment_name`, but it's unclear where this variable is actually used in the code.)
> 
> Including `OPENAI_ENDPOINT` also means that on line 122 we could conditionally set that value if it exist. This would solve the issue for allowing us to connect to other LLMs that use the same specification as OpenAIChatCompletion such as the ones served through vLLM.
> 
> And for the Azure/Non-Azure Logic from what I can tell, the only code difference is on lines 137 and 155, where `service=AzureChatCompletion(` is called. We could introduce a single variable, like ENABLE_AZURE_OPENAI, and use it to conditionally leverage the Azure libraries in these two instances if it's set to True.
> 
> What are y'alls thoughts on that?

Sure. See the flag implementation in the latest commit. I tested it with Azure OpenAI as working successfully. However, I don't have an OpenAI api key to test with. If you could test it, that'd be great. Cheers.

jland-redhat:
> > > Hi @jland-redhat, this article may help to understand the differences between the Azure endpoint for OpenAI and the native one.
> > > https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints
> > 
> > 
> > Ok I missed the custom library.
> > So still have a few suggestions I would like to run by y'all.
> > There is a lot of commented code here which I am not in love with.
> > What if instead we simplify the environment Variables. So instead of the current setup, what if we primarily stuck to `OPENAI_API_KEY` and `OPENAI_ENDPOINT`. (I see `AZURE_OPENAI_DEPLOYMENT_NAME` is being set to `deployment_name`, but it's unclear where this variable is actually used in the code.)
> > Including `OPENAI_ENDPOINT` also means that on line 122 we could conditionally set that value if it exist. This would solve the issue for allowing us to connect to other LLMs that use the same specification as OpenAIChatCompletion such as the ones served through vLLM.
> > And for the Azure/Non-Azure Logic from what I can tell, the only code difference is on lines 137 and 155, where `service=AzureChatCompletion(` is called. We could introduce a single variable, like ENABLE_AZURE_OPENAI, and use it to conditionally leverage the Azure libraries in these two instances if it's set to True.
> > What are y'alls thoughts on that?
> 
> Sure. See the flag implementation in the latest commit. I tested it with Azure OpenAI as working successfully. However, I don't have an OpenAI api key to test with. If you could test it, that'd be great. Cheers.

Hey Oliver,

Validated the change:
```
Validation using the default OpenAI API:
{"artifacts":[{"artifactId":"89f49deb-ff46-4338-8179-cf8e96bd2c24","description":"Result of request to agent.","name":"current_result","parts":[{"kind":"text","text":"The current exchange rate for USD to EUR is approximately 1 USD to 0.88331 EUR."}]}],"contextId":"c7872837-23e6-4a91-a0d1-3913902bc467","history":[{"contextId":"c7872837-23e6-4a91-a0d1-3913902bc467","kind":"message","messageId":"31a7404a-0b80-48e8-a908-80e264b8b935","parts":[{"kind":"text","text":"Dollar to Euro"}],"role":"user","taskId":"1e05a9a9-95b9-43c5-b2f0-f28524bbf27a"},{"contextId":"c7872837-23e6-4a91-a0d1-3913902bc467","kind":"message","messageId":"333e887b-83ea-41d4-b726-d1de225907af","parts":[{"kind":"text","text":"Processing function calls..."}],"role":"agent","taskId":"1e05a9a9-95b9-43c5-b2f0-f28524bbf27a"},{"contextId":"c7872837-23e6-4a91-a0d1-3913902bc467","kind":"message","messageId":"ff4fcba2-acfe-4e2d-8f76-594938a0632f","parts":[{"kind":"text","text":"Building the output..."}],"role":"agent","taskId":"1e05a9a9-95b9-43c5-b2f0-f28524bbf27a"}],"id":"1e05a9a9-95b9-43c5-b2f0-f28524bbf27a","kind":"task","status":{"state":"completed"}}
```

I was also hoping that we could have provided a custom endpoint to the `OpenAIChatCompletion` but the endpoint is hardcoded and changing it would involve creating a custom service which is annoying more complex than I was hoping. May try to pursue getting that working once this PR is up.

But this all LGTM!

oliverlabs:
@jland-redhat this is easy. I could add that in if you like (but you'd have to retest it). This would add one extra variable in the `.env` file and one extra parameter for the `def __init__(self)`. How do I get this PR merged now? :) I believe you need to re-approve yours now, too.

upd: I see, google/a2a-eng needs to approve.

jland-redhat (approved):

jland-redhat:
@oliverlabs I think you need to get one of the repo owners to approve in order to move forward. Not sure the best way to poke them to take a look

oliverlabs:
@moonbox3 @holtskinner @kthota-g @koverholt , perhaps one of you could merge?

kthota-g:
@moonbox3 would you be able to review this PR?

moonbox3:
> @moonbox3 would you be able to review this PR?

Definitely, @kthota-g. Will have a look today.

samples/python/agents/semantickernel/agent.py
```diff
@@ -93,18 +93,53 @@ class SemanticKernelTravelAgent:
...
+                raise ValueError('OPENAI_API_KEY environment variable must be set when not using Azure OpenAI.')
+            
+            model_id = os.getenv('OPENAI_CHAT_MODEL_ID', 'gpt-4.1')
 
         # Define a CurrencyExchangeAgent to handle currency-related tasks
-        currency_exchange_agent = ChatCompletionAgent(
-            service=OpenAIChatCompletion(
+        # Configure AzureChatCompletion service based on our settings
+        chat_service = None
+        if use_azure:
```
moonbox3: I'd prefer to see the use of a service selector, or something similar. We support many AI services and if we were to continue to add more here, the if, elif, else would grow to be too complicated.

samples/python/agents/semantickernel/agent.py
```diff
@@ -178,74 +206,55 @@ async def invoke(self, user_input: str, session_id: str) -> dict[str, Any]:
...
-                yield {
-                    'is_task_complete': False,
-                    'require_user_input': False,
-                    'content': 'Processing function calls...',
-                }
-                plugin_event.clear()
-
-            if any(isinstance(i, StreamingTextContent) for i in chunk.items):
-                if not text_notice_seen:
+            if any(
```
moonbox3: Please have a look at what is in main. This is no longer handled this way (as of SK >=1.30.0) - `FunctionCallContent` and `FunctionResultContent` are handled via the `on_intermediate_message` callback.

samples/python/agents/semantickernel/agent.py
```diff
@@ -93,18 +93,53 @@ class SemanticKernelTravelAgent:
     SUPPORTED_CONTENT_TYPES = ['text', 'text/plain']
 
     def __init__(self):
-        api_key = os.getenv('OPENAI_API_KEY', None)
-        if not api_key:
-            raise ValueError('OPENAI_API_KEY environment variable not set.')
-
-        model_id = os.getenv('OPENAI_CHAT_MODEL_ID', 'gpt-4.1')
+        # Check if Azure OpenAI should be used
+        use_azure = os.getenv('ENABLE_AZURE_OPENAI', 'false').lower() == 'true'
```
moonbox3: From a sample perspective, I'd prefer to keep this as simple as possible. We do plan to introduce other agent types here at some point (like the `AzureAIAgent`). This constructor is going to grow and become hard to manage.

Once we do `load_dotenv()` the env vars will be available when we create the `AzureChatCompletion()` or `OpenAIChatCompletion()` objects. It's less code to show:

`chat_service = AzureChatCompletion()` then to override everything with kwargs. Pydantic settings will handle the env vars underneath.

samples/python/agents/semantickernel/agent.py
```diff
@@ -296,9 +305,11 @@ async def _ensure_thread_exists(self, session_id: str) -> None:
         Args:
             session_id (str): Unique identifier for the session.
         """
-        if self.thread is None or self.thread.id != session_id:
+        # Replace check with self.thread.id when
+        # https://github.com/microsoft/semantic-kernel/issues/11535 is fixed
+        if self.thread is None or self.thread._thread_id != session_id:
```
moonbox3: This has already been addressed and is fixed in main. Please pull in the changes.

samples/python/agents/semantickernel/agent.py
```diff
@@ -93,18 +93,53 @@ class SemanticKernelTravelAgent:
...
+            
+            model_id = os.getenv('OPENAI_CHAT_MODEL_ID', 'gpt-4.1')
 
         # Define a CurrencyExchangeAgent to handle currency-related tasks
-        currency_exchange_agent = ChatCompletionAgent(
-            service=OpenAIChatCompletion(
+        # Configure AzureChatCompletion service based on our settings
+        chat_service = None
+        if use_azure:
+            chat_service = AzureChatCompletion(
```
moonbox3:
```suggestion
            chat_service = AzureChatCompletion()
```
Let's take advantage of Pydantic settings and not need to explicitly grab each env var.

Let's also please add something to the README about configuring the AI service for Azure OpenAI use:

```
#### Azure OpenAI

```bash
AZURE_OPENAI_CHAT_DEPLOYMENT_NAME=<deployment-name>  # e.g., "gpt-4.1"
AZURE_OPENAI_ENDPOINT=<endpoint>  # in the form of `https://<resource>.openai.azure.com/`
AZURE_OPENAI_API_KEY=<api-key>  # only required if using api key auth
AZURE_OPENAI_API_VERSION=<api-version>  # optional, defaults to the latest Azure OpenAI GA API version of `2024-10-21` if not provided
```
```

File: samples/python/agents/semantickernel/agent.py [hunk 7]
```diff
@@ -93,18 +93,53 @@ class SemanticKernelTravelAgent:
...
+                deployment_name=deployment_name,
                 api_key=api_key,
+                endpoint=endpoint,
+                api_version=api_version,
+            )
+        else:
+            # For standard OpenAI, use OpenAIChatCompletion with the correct model name
+            chat_service = OpenAIChatCompletion(
```
moonbox3: Same here...

```suggestion
            chat_service = OpenAIChatCompletion()
```

samples/python/agents/semantickernel/agent.py
```diff
@@ -93,18 +93,53 @@ class SemanticKernelTravelAgent:
...
-        api_key = os.getenv('OPENAI_API_KEY', None)
-        if not api_key:
-            raise ValueError('OPENAI_API_KEY environment variable not set.')
-
-        model_id = os.getenv('OPENAI_CHAT_MODEL_ID', 'gpt-4.1')
+        # Check if Azure OpenAI should be used
+        use_azure = os.getenv('ENABLE_AZURE_OPENAI', 'false').lower() == 'true'
+        
+        # Get API key based on selected service type
+        if use_azure:
```
moonbox3: All of this `if use_azure` and explicitly grabbing env vars isn't needed. Let's let the service grab what it needs - if something isn't configured properly, it wil fail with a message showing what's missing.

moonbox3 (changes requested):
Looks like this needs a merge from main, as well as some other updates. Appreciate you adding this functionality.

moonbox3:
Hi @oliverlabs, please let me know if you'd like some support getting the Azure OpenAI support added. I'm from the SK Python team and will be glad to help add the functionality.

oliverlabs:
@moonbox3 hi Evan, thanks for reviewing this PR. Apologies, it's a bank holiday weekend in the UK and I was OoO. I will reach out to you tomorrow with some comments.

samples/python/agents/semantickernel/agent.py
oliverlabs: Implemented the change where Azure OpenAI or standard OpenAI are selected based on the .env file variables as proposed.

samples/python/agents/semantickernel/agent.py
oliverlabs: Implemented the change where Azure OpenAI or standard OpenAI are selected based on the .env file variables as proposed.

samples/python/agents/semantickernel/agent.py
oliverlabs: Done.

samples/python/agents/semantickernel/agent.py
oliverlabs: Resolved.

samples/python/agents/semantickernel/agent.py
oliverlabs: Resolved.

samples/python/agents/semantickernel/agent.py
oliverlabs: merged code with the latest main

samples/python/agents/semantickernel/agent.py
oliverlabs: merged from the latest main code.

oliverlabs:
@moonbox3 , ~~could you please have a look at the `__init__(self)` method env variable logic? I have implemented it and it works, but I can't implement your suggestion to only use `chat_service = AzureChatCompletion()` with no variables. It fails to load when I do that.~~

UPD: fixed. I almost went mad searching for an error, but then noticed that `AZURE_OPENAI_CHAT_DEPLOYMENT_NAME` variable name was expected as opposed to `AZURE_OPENAI_DEPLOYMENT_NAME`. This should all work now. Please have a look and approve.

moonbox3:
Hi @oliverlabs, great progress. I would really prefer to have a more explicit approach to configuring the chat service, versus relying solely on if X number of env vars configured. Something like this (doesn't have to be that involved, but the enum is nice): https://github.com/microsoft/semantic-kernel/blob/main/python/samples/concepts/setup/chat_completion_services.py

The method `get_chat_completion_service_and_request_settings` gives one entry point to get the service / settings. To make it easier, we don't need to include the request settings, we can have the method be something like `get_chat_completion_service` and have it return the service only. It's pretty easy then to even include other AI Services like Google, Mistral or Anthropic, as an example - for those who want to test with the specific service. 

We can also link to SK's docs on configuring chat completion services in the README so devs can understand what needs to be included, if there are ever questions: https://learn.microsoft.com/en-us/semantic-kernel/concepts/ai-services/chat-completion/?tabs=csharp-AzureOpenAI%2Cpython-AzureOpenAI%2Cjava-AzureOpenAI&pivots=programming-language-python#creating-a-chat-completion-service along with [this doc](https://github.com/microsoft/semantic-kernel/blob/main/python/samples/concepts/setup/ALL_SETTINGS.md#semantic-kernel-settings) that spells out what env vars are required.

oliverlabs:
@moonbox3, well, it looks like they removed the samples from this repository altogether. So this was all for nothing. Thanks for your feedback though.

oliverlabs:
P.S. I rewrote the auth function to use the functions you mentioned @moonbox3. I will make a PR in the new repo, but I am not sure if you would have the maintainer rights there. I tested it and it works.

PR #527, opened 2025-05-13
Adds an example of agent-level OAuth authn/authz using Google ADK

whitlockjc:
This sample shows how to use the OpenAPI _Security Schemes_ and _Security Requirements_ to enforce agent-level authentication, using Google ADK and Google OAuth.

There are two scenarios demonstrated:

1. **Direct Authentication:** _(`U -> A [auth required]`)_
2. **Nested Authentication:** _(`U -> A -> B [auth required]`)_

For more details on how to run the sample, please review the sample's `README.md`.

**Note:** The CLI **DOES NOT** work with this yet


whitlockjc:
This is just a sample, and it **DOES NOT** take into account best practices or any sort of properly plumbing/scoping/securing of credentials. This is a PoC _(proof of concept)_ to see what it would take to support an OAuth flow to get credentials from the user to a downstream agent. In theory, the `A2AClient` and `TaskManager` would properly handle all of this but for now, plumbing was brute force and for demonstration purposes only.

dmandar:
Do we anticipate any further changes to this ADK code to get the CLI working?

whitlockjc:
> Do we anticipate any further changes to this ADK code to get the CLI working?

For the CLI to work, it would need to have a way to ask the user for credentials directly _(API Key, Bearer Token, ...)_ and/or handle the `A2AClientError` for authn/authz by handling the OAuth callback. Once it has the credentials, your work already has the `A2AClient` support for providing these things to `A2AServer` requests. But...to be done properly, I'm assuming that at some point, the SDK was a whole will need some mechanism to simplify this for A2A implementors.

whitlockjc:
To follow up to my previous comment, much like the Demo UI, supporting auth in the CLI is not a simple task. `U -> A [auth required]` could work but we'd need to come up with something fast to support `U -> A -> ... [auth required]` because the CLI itself needs to get the credentials and update the client _(and pass credentials downstream potentially)_.

whitlockjc:
I just realized we could potentially use the OAuth `state` variable to embed more useful information into the OAuth redirect URI that might help in the plumbing of information as well.

c-mysec:
Some comments (i am not an OSS contributor, so i don't know how this works):

1. On windows remove the "| jq ." and run the command using bash. Or, in powershell use:
$body = @{
    jsonrpc = "2.0"
    id = 1
    method = "message/send"
    params = @{
        id = "t1"
        sessionId = "fa09c65ea23a4ee59f9f2017ea01183a"
        configuration = @{
            acceptedOutputModes = @("text")
        }
        message = @{
            messageId = "m1"
            role = "user"
            parts = @(
                @{
                    type = "text"
                    text = "Consult the magic 8-ball and ask the following: Will the Braves win the World Series?"
                }
            )
        }
    }
} | ConvertTo-Json -Depth 10
$response = Invoke-WebRequest -Uri "http://localhost:10003/" -Method POST -Body $body -ContentType "application/json"
$response.Content

2. PersonalAssistantAgent - I think it is worth to add a comment in the readme informing that the PersonalAssistantAgent will impersonate the first user that authenticate with it (it is a "personal assistant"). Maybe also include note that an Rest API backend for multiple user clients should do something like OIDC, validate user authorization on each request and access Magic_8ball using each user's access_token.

3. The 2 applications (Magic_8ball and PersonalAssistantAgent) are using the SAME Credentials. This is a bad practice to share a secret. Google Auth is considering them the same application. I think this is bad also for a demo as some developers can follow this practice. Create 2 client Ids.

4. Is it worth for a demo to separate the 2 apps? "angent.py" and "middleware.py" have code for both apps.

holtskinner (changes requested):
Can you adjust this sample to use the A2A SDK?

whitlockjc:
@c-mysec Here is my response:

1. Done.
2. There are various places in code _(after your comments)_ that mention this is for demonstration purposes, and that this could/should be updated for user/agent/... scoping of credentials in a real-world application.
3. I refactored the whole sample to no longer work this way, the sample as-reviewed was more WIP than what is there now
4. I originally started with just one _"app"_ but when you have to juggle CLI options to tie the different pieces together, it became painful so I broke them out into their own, standalone apps.

I think a good bit of your feedback was relevant in earlier versions, but I _think_ I've addressed things.

holtskinner:

PR #530, opened 2025-05-14
fix: updating schema to match the json schema protocol

pauldotknopf:
https://json-schema.org/draft-07/schema

"$defs" should be "definitions"

# Description

### DIFF
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -2,7 +2,7 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "A2A Protocol Schema",
     "description": "JSON Schema for A2A Protocol",
-    "$defs": {
+    "definitions": {
       "AgentAuthentication": {
         "properties": {
           "schemes": {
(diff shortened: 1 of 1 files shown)


darkhaniop:
The tool used to generate `a2a.json` probably used the latest version [draft 2020-12](https://json-schema.org/draft/2020-12/schema). The key `definitions` was replaced with `$defs` in 2019-09 (predecessor of 2020-12).

From https://json-schema.org/draft/2019-09/draft-handrews-json-schema-validation-02#rfc.appendix.A
> "definitions"
>     Renamed to "$defs" to match "$ref" and be shorter to type. Schema vocabulary authors SHOULD NOT define a "definitions" keyword with different behavior in order to avoid invalidating schemas that still use the older name. While "definitions" is absent in the single-vocabulary meta-schemas referenced by this document, it remains present in the default meta-schema, and implementations SHOULD assume that "$defs" and "definitions" have the same behavior when that meta-schema is used. 

If that's the case, the only line that needs to be corrected is the `$schema` specifier from
https://github.com/google/A2A/blob/175cecd8dab1dd1d66d6c5b8b0ec7e5390a543d9/specification/json/a2a.json#L2
to:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  ...
}
```

---
*Edit: add a quote from the referenced document*

holtskinner (approved):

kthota-g:
This issue is fixed with the new types and introduction of generator.

PR #531, opened 2025-05-14
AgentCard demo ADK and CLI samples

dmandar:
Add new agent and CLI samples for showcasing authenticated extended agent cards use case.

### DIFF
--- a/samples/python/agents/extended_agent_card_adk/task_manager.py
+++ b/samples/python/agents/extended_agent_card_adk/task_manager.py
@@ -26,6 +26,7 @@ from common.types import (
     GetAuthenticatedExtendedCardResponse,
     InvalidRequestError,
     AgentSkill,
+    AgentCard,
     UnsupportedOperationError,
 )
 from google.genai import types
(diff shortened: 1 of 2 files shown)


samples/python/agents/extended_agent_card_adk/task_manager.py
```diff
@@ -0,0 +1,367 @@
...
+                else None,
+            ),
+            [Artifact(parts=parts)]
+            if task_state == TaskState.COMPLETED
+            else [],
+        )
+        return SendMessageResponse(id=request.id, result=task)
+
+    async def on_get_authenticated_extended_card(
+        self, http_request: Request, rpc_request: GetAuthenticatedExtendedCardRequest, base_agent_card: "AgentCard" # type: ignore
```
mikeas1: Remove the quotes on AgentCard and add it to the common.types import at the top. Then you can get rid of the `type: ignore` comment.

mikeas1 (approved):

samples/python/agents/extended_agent_card_adk/task_manager.py
dmandar: Oops. I had already added it to types..fixed.

PR #555, opened 2025-05-15
feat: Add a new video generation AGENT that uses the VEO model

dmandar:
Add a new video generation agent that creates videos using VEO. The agent leverages Google's Veo model to generate videos based on text prompts.

The agent takes text requests from the client, such as "Create a video of a cat playing with a fur ball". The generated video is then uploaded to Google Cloud Storage (GCS), and a signed URL for accessing the video is provided back to the client.
Currently, the Veo API does not provide real-time progress updates during video generation. To enhance user experience, this agent simulates progress updates while the video is being processed. 
Also includes changes to the demo app to display the received video. (Note: the demo app does currently show the progress of the video creation process. The CLI can be used to see the simulated progress percentage)

### DIFF
--- a/demo/ui/components/chat_bubble.py
+++ b/demo/ui/components/chat_bubble.py
@@ -1,118 +1,132 @@
 import mesop as me
-
 from state.state import AppState, StateMessage
 
 
-@me.component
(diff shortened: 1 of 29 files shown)


holtskinner:

PR #583, opened 2025-05-19
feat: Update the samples to use 0.2 SDK 

pstephengoogle:
Aligning the samples with the SDK for protocol version 0.2

Moves from the TaskManager abstraction to the AgentExecutor abstraction, simplifying the implementation costs for wrapping agents as A2A compliant


salqadri:
This is just the python samples; The javascript / typescript samples also need updating.

pstephengoogle:
Yes, the typescript part doesn't use the SDK. Will update the typescript to be compliant with 0.2 in a follow up PR

rajeshvelicheti (approved):

salqadri:
> Will update the typescript to be compliant with 0.2 in a follow up PR

Thanks! What's the ETA?

PR #589, opened 2025-05-19
fix: Replace metadata type with Record<string, any> equivalent

kthota-g:
# Description

metadata was previously represented as metadata?: Record<string, any>, to imply this is a key-value pair. However defining in this format creates a new type in json-schema. To address the json schema generation, we used metadata?: object which fixes the generation issue. However, documenting metadata?: object in specification implies that metadata can accept more than just key-value pairs.

To address this, spec will still have metadata?: Record<string, any>, but we will use equivalent `{ [key: string]: any;}` to represent this attribute in types.

### DIFF
--- a/types/src/types.ts
+++ b/types/src/types.ts
@@ -128,7 +128,9 @@ export interface Task {
   /** collection of artifacts created by the agent. */
   artifacts?: Artifact[];
   /** extension metadata */
-  metadata?: object;
+  metadata?: {
+    [key: string]: any;
+  };
   /** event type */
   kind: "task";
 }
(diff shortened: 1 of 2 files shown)


types/src/types.ts
```diff
@@ -128,7 +128,9 @@ export interface Task {
   /** collection of artifacts created by the agent. */
   artifacts?: Artifact[];
   /** extension metadata */
-  metadata?: object;
+  metadata?: {
```
swapydapy, line 131: Did you generate the python types from this new definition. Do they remain the same?

Also could you add explanation behind this change in the PR description. That would help others also get the context.

swapydapy (changes requested):

types/src/types.ts
kthota-g: Types are unaffected with this change.

swapydapy (approved):

PR #595, opened 2025-05-19
chore: Update Auth0 example with SecuritySchemes changes

mikeas1:
# Description

Update the headless_auth_agent to use the new SecuritySchemes. The code was basically already there, just actually using the types now present.

Some changes picked up from ruff formatting.


holtskinner:

PR #601, opened 2025-05-20
fix(demo): Fix bugs in demo app

pstephengoogle:
- Address changes for ADK upgrade
- Fix bug with incorrect type -> kind replacement

### DIFF
--- a/demo/ui/service/server/adk_host_manager.py
+++ b/demo/ui/service/server/adk_host_manager.py
@@ -377,7 +377,7 @@ class ADKHostManager(ApplicationManager):
                     current_task.artifacts = []
                 current_task.artifacts.append(artifact)
             else:
-                # this is a chunk of an artifact, stash it in temp store for assemling
+                # this is a chunk of an artifact, stash it in temp store for assembling
                 if artifact.artifactId not in self._artifact_chunks:
                     self._artifact_chunks[artifact.artifactId] = []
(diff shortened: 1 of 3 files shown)


demo/ui/state/host_agent_service.py
```diff
@@ -272,7 +272,7 @@ def extract_content(
         elif p.kind == 'data':
             try:
                 jsonData = json.dumps(p.data)
-                if 'kind' in p.data and p.data['kind'] == 'form':
+                if 'type' in p.data and p.data['type'] == 'form':
```
holtskinner, line 275: Shouldn't this be `kind` now? Or is this something else?

demo/ui/service/server/adk_host_manager.py
holtskinner, line 1: Fix the spelling on line 380 "assembling"

holtskinner (changes requested):

demo/ui/state/host_agent_service.py
mikeas1: Talked with Phil offline. This is part of the JSON payload in the Data part, so `type` is correct.

mikeas1 (approved):

demo/ui/service/server/adk_host_manager.py
mikeas1: Done

PR #612, opened 2025-05-21
chore: Use push notifier from sdk

swapydapy:


samples/python/agents/semantickernel/__main__.py
holtskinner, line 1: Fix spelling on line 42 to "Semantic"

holtskinner (approved):

samples/python/agents/semantickernel/__main__.py
swapydapy: Done

PR #620, opened 2025-05-21
ci: Add Auto-generation of `a2a.json` from `types.ts`

holtskinner:
Use @a2a-bot to auto-generate `specification/json/a2a.json` whenever `types/src/types.ts` is updated in a PR.

### DIFF
--- a/.github/workflows/generate-a2a-json.yml
+++ b/.github/workflows/generate-a2a-json.yml
@@ -1,0 +1,58 @@
+name: Auto-generate A2A JSON Schema
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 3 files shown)


ognis1205:
Hi @holtskinner — I have a question about this PR. Is it correct to understand that, going forward, `types/src/type.ts` will be treated as the single source of truth?

holtskinner:
@ognis1205 `types.ts` is the file that the specification `a2a.json` is generated from.

ognis1205:
Thanks for the clarification @holtskinner! So going forward, updates and spec changes will happen in `types.ts` first — I'll keep an eye on that file to stay aligned with the latest changes.

PR #635, opened 2025-05-22
Feat: A2A-MCP Interplay

rajeshvelicheti:
# Description

- [ ] Follow the [`CONTRIBUTING` Guide](https://github.com/google/A2A/blob/main/CONTRIBUTING.md).
- [ ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [ ] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)

Fixes #<issue_number_goes_here> 🦕


samples/python/agents/a2a_mcp/src/a2a_mcp/common/agent_runner.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        user_id: str = 'user_1',
+        app_name: str = 'A2A-MCP',
+    ):
+        self.session_service = InMemorySessionService()
+        self.session = None
+        self.app_name = app_name
+        self.user_id = user_id
+
+    async def run_agent(
```
pstephengoogle: If we aren't using this entrypoint now, lets remove it

samples/python/agents/a2a_mcp/src/a2a_mcp/common/agent_runner.py
```diff
@@ -0,0 +1,124 @@
...
+            new_message=content,
+        ):
+            if event.is_final_response():
+                response = ''
+                if (
+                    event.content
+                    and event.content.parts
+                    and event.content.parts[0].text
+                ):
+                    response = '\n'.join(
```
pstephengoogle, line 68: for readability why not yield in the if statements rather than collection the response and yielding it after? E.g. it seems you will yield an empty response if the if or elif aren't matched in this final response block. Should you throw an error or return an error state instead?

samples/python/agents/a2a_mcp/src/a2a_mcp/common/types.py
```diff
@@ -0,0 +1,137 @@
...
+class ServerConfig(BaseModel):
+    """Server Confgiguration"""
+
+    host: str
+    port: int
+    transport: str
+    url: str
+
+
+class Task(BaseModel):
```
pstephengoogle: nit: only if you think it is not a hard thing to do: Maybe name this PlannerTask or something to as not to confuse any reader in regards to an A2A Task?

samples/python/agents/a2a_mcp/src/a2a_mcp/common/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+            a2a_client = A2AClient(httpx_client, agent_card)
+            logger.info(f'Agent Card {a2a_client.url}')
+            return a2a_client
+
+    async def execute(
+        self, httpx_client: httpx.AsyncClient
+    ) -> AsyncIterable[dict[str, any]]:
+        logger.info(f'Executing node {self.id}')
+        if self.node_key == 'planner':
+            self.client = await self.get_planner_resource(httpx_client)
```
pstephengoogle: is there a reason that this client should be a class member and not a local variable?

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

pstephengoogle (approved):

Issue #109, opened 2025-04-12
Suggestion: Dynamic Capabilities in A2A Protocol


abhinav24jha:
Hi A2A team,

Firstly, thank you for open-sourcing the A2A protocol, it's a promising foundation for structured agent interactions, and I’m excited about its potential.

I noticed that the current capabilities system is static, which limits flexibility in real-world adaptive systems.

I've tried to outline this limitation and a possible direction to address it in the following write-up:

https://rigorous-macrame-0de.notion.site/Problem-with-A2A-Protocol-Proposed-Solution-1d2b6c0f2015809eb100cfc1c025895e

I’d absolutely love any thoughts, whether this direction makes sense, aligns with the protocol's design philosophy, or if there's an alternative approach you're exploring.

If anything is unclear or worth adjusting, I’m happy to revise. Thanks again for all your work on A2A!

Sincerely,
Abhinav

heiko-hotz:
Hi @abhinav24jha , first of all thanks so much for this detailed write up, this is fantastic. You clearly have thought quite a bit about A2A and its capabilities and limitations.

I believe the core problem you have identified is valid and significant. The tension between static capability declaration and the emergent, flexible nature of tool-using LLM agents is a real challenge for agent orchestration and discovery. Relying solely on predefined lists _can_ be limiting. And I do like that your proposal adheres to A2A's fundamental principle of agent encapsulation by avoiding the sharing of internal tools, memory, or thought processes.

That being said, in my personal opinion (which might be different from the opinion of the A2A team) the proposed solution is very complex. It introduces significant complexity in several areas:

- Dynamic Rubric Generation: Reliably generating a meaningful and objective rubric programmatically for any arbitrary sub-task seems like a substantial challenge in itself, potentially requiring its own sophisticated reasoning capabilities.
- Agent Self-Assessment Burden: Requiring every participating agent to implement potentially complex internal reasoning (like Chain-of-Thought) specifically to evaluate itself against these dynamic rubrics adds considerable overhead and raises questions about the consistency and reliability of self-assessment across different agent implementations.
- Latency and Overhead: The multi-step process (identify gap -> generate rubric -> query agents -> self-assess -> score -> assign) could introduce significant latency before the actual sub-task execution begins.

[...]

PR #642, opened 2025-05-23
feat(schema): support open agent discovery under shared base URL via API Catalog

### LINKED ISSUE #641: [Feat]: Support multi-agent hosting under a single base URL via RFC 9264-compatible API Catalog

Currently, the specification assumes that each agent is served from a unique base URL, effectively mapping one agent per host. However, in practice, it is common for a single host to act as a routing layer for multiple agents, each accessible via different paths under the same base URL. This setup creates the need for a standardized (open) discovery mechanism that allows clients to enumerate and resolve multiple agents served from a common origin.

I propose supporting the discovery of multiple agents under a shared base URL by leveraging the [API Catalog](https://datatracker.ietf.org/doc/draft-ietf-httpapi-api-catalog/08/) format defined in the corresponding Internet Draft. Specifically, a host could expose an API Catalog that lists multiple agent entries using linkset objects. Each entry would contain:

 - An `anchor` representing the base path for an individual agent (e.g., https://example.com/agents/agent-a)
 - A `describedby` relation pointing to an Agent Card that describes the agent's capabilities, metadata, and entry points.

This approach would allow clients to dynamically discover and integrate with multiple agents hosted under a single origin, in a standardized and interoperable way.

An alternative approach discussed in the GitHub Discussion involves using a custom schema to define and expose agent API endpoint information.

### Additional context

 - [GitHub Discussion](https://github.com/google/A2A/discussions/166)
 - [Internet-Draft api-catalog: a well-known URI and link relation to help discovery of APIs](https://datatracker.ietf.org/doc/draft-ietf-httpapi-api-catalog/08/)
 - [RFC 9264 Linkset: Media Types and a Link Relation Type for Link Sets](https://www.rfc-editor.org/rfc/rfc9264)

### Code of Conduct

ognis1205:
# Description

Fixes #641 🦕

## Points for Review

1. Does it make sense to include **API Catalog-based Open Discovery** in the specification?  
2. Should `application/json+agent-card` be added as an **IANA-registered media type**?  
3. The PR currently proposes the API Catalog-based notation but does **not yet finalize a single canonical format**. Should we make a decision to finalize it in this PR, or leave it flexible for now?

~~Note: This PR focuses on the schema/spec-level support. A separate PR demonstrating the PoC SDK implementation and demo using this proposal is available at [a2a-python #109](https://github.com/google/a2a-python/pull/109), and is intended to complement this spec-level change.~~


cdavernas:
Awesome!!!

ognis1205:
Hey @cdavernas , Thanks for the comment!

The schema defined in this PR is implemented to conform to the following RFC and Internet Draft.  
I’d appreciate it if you could also review this along with the upcoming PRs to `a2a-python`:

- [Internet-Draft api-catalog: a well-known URI and link relation to help discovery of APIs](https://datatracker.ietf.org/doc/draft-ietf-httpapi-api-catalog/08/)
- [RFC 9264 Linkset: Media Types and a Link Relation Type for Link Sets](https://www.rfc-editor.org/rfc/rfc9264)

darkhaniop:
The generated `AgentLinkTarget` schema definition does not include the `additionalProperties` property, without which link metadata, such as link `title`, cannot be supplied. Did it fail to pick up this declaration from `LinkTarget`?
https://github.com/google/A2A/blob/60ba67bc9a2ae82ee320424e6b2683334ebf4c3c/types/src/types.ts#L551-L555

By the way, I like the simplicity of this approach.

ognis1205:
Hey @darkhaniop, thanks again for the comment and review!

I double-checked the JSON Schema spec, and it turns out that by default, all additional properties are allowed—meaning the schema will accept extra fields even without explicitly specifying `additionalProperties`:

https://json-schema.org/understanding-json-schema/reference/object#additionalproperties

It seems that the presence of `additionalProperties: {}` (which is equivalent to `additionalProperties: true`) in the `LinkTarget` schema—but not in `AgentLinkTarget`—is due to how `typescript-json-schema` generates the output:

https://github.com/ognis1205/A2A/blob/60ba67bc9a2ae82ee320424e6b2683334ebf4c3c/specification/json/a2a.json#L1139

That can definitely make the generated schema a bit confusing to read. That said, a couple of points to keep in mind:

 - When generating downstream code like `google/a2a-python`, we might need to ensure the appropriate options are set—even though there’s no issue with how things are currently handled within `google/A2A` itself.
 - From a schema validation perspective, the current setup is fine. However, when thinking in terms of SDK models, it would mean we don’t expose APIs to set fields like `title`, for instance (though this shouldn’t be a problem when parsing raw JSON).

What do you think? Let me know if I’m missing anything important.

P.S.
Many thanks to @Kevsy and @darrelmiller for the elegant simplicity of the schema.

darkhaniop:
@ognis1205, you are right about the current schema output being okay for validation, I did not know that. I just checked the json-schema.org docs you linked, and realized that I was incorrect in thinking that by default any unmatched property is disallowed. From the linked JSON schema docs page:

> By default any additional properties are allowed.

I think your bullet points raise a valid question about SDKs implementing the interfaces as defined in `a2a.json`. If our goal is to explicitly show that additional metadata in `AgentLinkTarget` objects should be allowed (I do not know if you propose to show that or not), maybe including `additionalProperties` in its schema definition would be beneficial. Because doing so would result in more consistently generated downstream code.

<details>
<summary>Pydantic model generation test</summary>

### Code generation test

To check how explicitly stating `additionalProperties` vs. omitting it in the schema may affect downstream SDKs and libraries that generate interfaces from `a2a.json`, I tested type generation for Pydantic.

I used the `datamodel-code-generator` package to generate types from the following schema:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "WithoutAddProps": {
      "properties": {
        "href": {"type": "string"}
      }
    },
    "WithAddProps": {
      "additionalProperties": true,
      "properties": {
        "href": {"type": "string"}
      }
    }
  }
}
```
Which results in (omitting the root model):
```python
class WithoutAddProps(BaseModel):
    href: Optional[str] = None


class WithAddProps(BaseModel):
    class Config:
        extra = Extra.allow

    href: Optional[str] = None
```
While a link object with extra fields validates successfully against both types, the first variant would lose extra fields on deserialization:
```python
link_obj = {"href": "https://example.com", "title": "Example"}
obj_wo_add_props = WithoutAddProps.model_validate(link_obj)
obj_w_add_props = WithAddProps.model_validate(link_obj)

print("Loaded using WithoutAddProps:\n", obj_wo_add_props.model_dump())
print("Loaded using WithAddProps:\n", obj_w_add_props.model_dump())
```
Output:
```
Loaded using WithoutAddProps:
 {'href': 'https://example.com'}
Loaded using WithAddProps:
 {'href': 'https://example.com', 'title': 'Example'}
```

P.S. Although it looks like `datamodel-code-generator` outputs some deprecated syntax, I used this method because [Pydantic docs](https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/) features it, so other A2A library developers may rely on it.
</details>

ognis1205:
Hey @darkhaniop, thank you for sharing your investigation! This really helps!

For now, it’s still uncertain whether this feature request itself will be accepted by the community, so I think it might be better to prioritize getting the protocol evaluated by a broader audience—through PR demos including `google/a2a-python`—rather than pushing for this specific change. (The properties other than `href` aren’t critical at this point.)

That said, depending on how things develop, the points you raised could become important, so it’s definitely worth keeping them in mind for the future.

ognis1205:
I’ve removed the WIP status from this PR due to changes in the repository structure. The related SDK implementation will continue in `a2a-python`, but for the demo, I plan to publish it using a fork of the relevant SDK—either in the official `sample` repository or in my personal repository. Looking forward to your review.

ognis1205 (commented):
~~TODO: To complete this change, the specification document `docs/specification.md` should also be updated accordingly.~~

Done.

ognis1205:
## PoC: SDK Example for Agent Mounting and API Catalog Discovery

I’ve implemented a proof-of-concept SDK that demonstrates how to define and mount multiple A2A agents using `A2AStarletteRouteBuilder` and `A2AStarletteBuilder`.  Each agent is configured with its own `AgentCard`, skill set, and handler, and mounted to a shared Starlette application. The resulting server responds with a valid A2A catalog at `/.well-known/api-catalog.json`, like the following:

```json
{
  "linkset": [
    {
      "anchor": "http://localhost:9999/a2a/hello/",
      "describedby": [
        {
          "href": "http://localhost:9999/a2a/hello/agent.json",
          "type": "application/json"
        }
      ]
    },
    {
      "anchor": "http://localhost:9999/a2a/echo/",
      "describedby": [
        {
          "href": "http://localhost:9999/a2a/echo/agent.json",
          "type": "application/json"
        }
      ]
    }
  ]
}
```

### Minimal Example Code

```python
@click.command()
@click.option('--host', default='localhost')
@click.option('--port', default=9999)
def main(host: str, port: int):
    hello_skill = AgentSkill(
        id='hello_world',
        name='Returns hello world',
        description='just returns hello world',
        tags=['hello world'],
        examples=['hi', 'hello world'],
    )
    hello_card = AgentCard(
        name='Hello World Agent',
        description='Just a hello world agent',
        url=f'http://{host}:{port}/a2a/hello',
        version='1.0.0',
        defaultInputModes=['text'],
        defaultOutputModes=['text'],
        capabilities=AgentCapabilities(streaming=True),
        skills=[hello_skill],
        supportsAuthenticatedExtendedCard=False,
    )
    hello_handler = DefaultRequestHandler(
        agent_executor=HelloWorldAgentExecutor(),
        task_store=InMemoryTaskStore(),
    )
    hello_agent = A2AStarletteRouteBuilder(
        agent_card=hello_card,
        http_handler=hello_handler,
    )

    echo_skill = AgentSkill(
        id="echo",
        name="Echo input",
        description="Returns the input text as is",
        tags=["echo"],
        examples=["Hello!", "Repeat after me"],
    )
    echo_card = AgentCard(
        name="Echo Agent",
        description="An agent that echoes back your input.",
        url=f"http://{host}:{port}/a2a/echo",
        version="1.0.0",
        defaultInputModes=["text"],
        defaultOutputModes=["text"],
        capabilities=AgentCapabilities(streaming=True),
        skills=[echo_skill],
        supportsAuthenticatedExtendedCard=False,
    )
    echo_handler = DefaultRequestHandler(
        agent_executor=EchoAgentExecutor(),
        task_store=InMemoryTaskStore(),
    )
    echo_agent = A2AStarletteRouteBuilder(
        agent_card=echo_card,
        http_handler=echo_handler,
    )

    server = (
        A2AStarletteBuilder()
            .mount(hello_agent)
            .mount(echo_agent)
            .build()
    )
    uvicorn.run(server, host=host, port=port)


if __name__ == "__main__":
    try:
        main()
    except Exception:
        print(traceback.format_exc(), file=sys.stderr)
```

This can serve as a practical SDK example when reviewing this PR. The following is the related PoC implementation of the SDK:

- [google-a2a/a2a-python#109](https://github.com/google-a2a/a2a-python/pull/109) – SDK implementation

types/src/types.ts
```diff
@@ -577,6 +577,104 @@ export interface PasswordOAuthFlow {
 // --8<-- [end:PasswordOAuthFlow]
 
 // --8<-- [start:JSONRPCMessage]
```
holtskinner: Can you put your change above this line? 

```
// --8<-- [start:JSONRPCMessage]
```
It's used for Code Snippets in the documentation.

holtskinner (changes requested):

types/src/types.ts
ognis1205: Thanks for pointing that out! I've moved the schema above the JSONRPCMessage marker so it can be picked up correctly by the documentation generator.

darrelmiller:
The API Catalog draft is now a finalized RFC https://datatracker.ietf.org/doc/rfc9727/

ognis1205:
> The API Catalog draft is now a finalized RFC https://datatracker.ietf.org/doc/rfc9727/

Thanks @Kevsy @darrelmiller for all the hard work in getting this over the line — huge respect for pushing this through the standardization process. I’ll go over the finalized RFC again carefully. Looking forward to seeing it adopted more widely!

ognis1205:
Also, if you (and/or the community) think this issue/PR would be better opened or led by you, @darrelmiller, please don’t hesitate to let me know — happy to step back if needed.

ognis1205 (commented):
### TODO:

- [x] Update links to point to the finalized RFC (RFC 9727)
- [x] Remove mentions of “draft” or “proposed” status
- [x] Review and verify compliance with the finalized RFC
  - [x] https://www.rfc-editor.org/rfc/rfc9727.html#section-4.2-1

kthota-g:
Few thoughts on this:

I see Agent Catalog as a component of Registry and not part of a A2A Protocol or A2AServer itself. There will be scenarios where Agent cards are exposed via different runtimes (e.g. agents running in different ecosystems, different servers etc.) that span more than just one A2AServer that need to be added to a catalog to facilitate their discovery.

[This page](https://a2aproject.github.io/A2A/latest/topics/agent-discovery/) discusses various discovery mechanisms. If an A2AServer is exposing an agent, it can use well-known url to publish its agent card. Alternatively, an A2AServer can expose agent cards at custom urls and leverage direct configuration/private discovery for those agents.

For scenarios where multiple agent cards need to be exposed for discovery, a curated registry service would act as the catalog server and the registry server would address how the resources in a catalog would be exposed to the clients. The specification to adopt to support this functionality is to be determined, api-catalog or any other spec that suits well considering additional requirements such as policy enforcement (e.g. entitlements to determine what agents are available for a client).

ognis1205:
Hey @kthota-g , thanks for sharing your thoughts — I think you raise several important points, especially around heterogeneous agent deployments and the potential role of a (centralized) registry service.

That said, I’d like to offer a small clarification regarding the intent behind using the API Catalog (RFC 9727) **format** for the Agent Catalog.

The schema is specifically designed to support **Same-Origin, Multi-Agent** configurations through **open discovery**, where a single server (or gateway) hosts multiple agents under subpaths (e.g., `/a2a/agent1`, `/a2a/agent2`) and exposes a catalog at a well-known URL `/.well-known/api-catalog`.

In this context, the term **"Agent Catalog"** follows the naming convention established by **API Catalog** in RFC 9727 — it refers to **a standardized format and discovery surface**, **not a service-level registry or general-purpose cataloging service**. This distinction is important: while centralized registries or entitlement-aware catalogs have their own use cases, the Agent Catalog is scoped specifically for open discovery on shared origins.

That said, I do share some of your intuition — especially when considering whether the Agent Catalog should be viewed as part of the A2A protocol itself, or as a more loosely coupled discovery mechanism layered on top.

Happy to continue the discussion around how we might layer those additional capabilities while keeping compatibility with the standard where possible.

darrelmiller:
@ognis1205  I think you have done an excellent job of driving this feature and justifying its requirements.  I'm right behind you :-)

@kthota-g While I agree that discovery and protocol should ideally be distinct concerns, the protocol specification already opened the door to addressing discovery by defining a .well-known endpoint.   Unfortunately, it does so in a way that constrains a host to a single agent which I personally think is fairly impractical requirement.  This has led people to get creative. So far I have seen people abusing .well-known by adding path segments before it, I've seen implementations that support an "agents.json" and we have this suggestion to use api-catalog, all to work around this limitation.

I think the spec should either remove the proposal to use .well-known and commit to the stance that discovery and protocol are distinct concerns, or it should acknowledge that are are valid reasons for having a host with multiple agents and provide a recommended approach for that.  I would be fine with either approach, but the current status quo is confusing.

ognis1205:
Hi @darrelmiller, thank you — I really appreciate your kind words and support!

And yes, I completely agree with your assessment of the current ambiguity. In fact, I had also considered proposing the removal of the `.well-known` suggestion from the protocol layer entirely, leaving discovery as an optional, separately layered concern.

The `api-catalog` format was introduced primarily as a workaround to the single-agent limitation of the current `.well-known` usage, not as a hard commitment to tie discovery into the protocol itself.

Happy to keep exploring which path would give us the right balance between clarity, flexibility, and practical interoperability.

docs/specification.md
```diff
@@ -281,6 +281,100 @@ Provides a declaration of a combination of the target url and the supported tran
...
+
+```json
+{
+  "linkset": [
+    {
+      "anchor": "https://georoute-agent.example.com/a2a/v1/route_planner",
+      "describedby": [
+        {
+          "href": "https://georoute-agent.example.com/a2a/v1/route_planner/agent.json",
+          "type": "application/json",
```
Fannon:
```suggestion
          "type": "application/json",
```

This is maybe a general shortcoming of the API Catalog RFC, but the media type "application/json" in itself is not very descriptive. From this linkset result, you would not know that this describes an A2A Agent at all and also not that it's described via the Agent Card metadata format. To learn this, you have first to download the file and the file itself needs to be self-descriptive enough (have a discriminator what spec it is and what it describes).

But it's a very simple approach. 
For a more complex approach, based on experiences building cross-content, cross-technology catalogs, there is also the open-source [Open Resource Discovery](https://open-resource-discovery.github.io/specification/) specification, which does the same, but is more extensive and not just limited to APIs.

In case you have questions on that, feel free to get in touch.

ognis1205: Thank you for the thoughtful comment — I completely agree that simply using `application/json` as a media type is not sufficiently descriptive.

As a pragmatic interim step, using a more specific media type such as `application/vnd.a2a.agent-card+json` can serve as a practical workaround.

Incidentally, there are already active discussions on related topics — such as the IANA registration of `.well-known` URIs — for example, in [protocol-registries/well-known-uris#66](https://github.com/protocol-registries/well-known-uris/issues/66). I expect similar conversations around media types and schema identification will emerge as the specifications mature.

Regarding ORD — I’ve reviewed the documentation and find it to be a very well-designed and extensible protocol. I believe it has strong potential to inform and support open discovery initiatives.

That said, I’d like to raise two considerations:

 - **SAP ecosystem focus:** While ORD is open, it appears to have been primarily designed with usage in the SAP ecosystem in mind. In contrast, A2A’s design goal is to remain vendor-neutral. Standards implicitly tied to a specific platform may hinder broader adoption.
 - **Complexity vs. minimalism:** ORD offers rich and expressive metadata, which is impressive. However, our current goal here with A2A Open Discovery is to propose a minimal, widely interoperable mechanism. From this perspective, ORD might currently be more comprehensive than necessary.

That said, we believe ORD’s architectural insights could be highly valuable — not only for Open Discovery but also for ongoing efforts such as the Agent Catalog and Registry:

https://github.com/a2aproject/A2A/discussions/741

If you’re open to it, we’d love to have you join the conversation!

Fannon: @ognis1205 Thanks for replying so quickly and your feedback!
Yes, defining a more specific media type for agent-cards would address the concerns, then it's at least clear what links are pointing to agent cards. Maybe `application/agent-card+json` could also be sufficient? Then it's closer to your proposed `.well-known/agent-card.json`. 

I fully agree with your complexity vs. minimalism argument here, that's why I think you're making the right choice when going with this simpler approach. Something like ORD becomes only attractive when you're building a catalog / metadata architecture that combines many types of protocols and metadata and you want to connect them and govern them together. 
Yes, ORD is currently focused on the SAP ecosystem, but we recently donated it to Linux Foundation and I hope it will grow beyond it. Being vendor-neutral is a design-goal of ORD as well, but clearly it grew out of SAPs specific requirements. 

I'd be happy to join the conversation on it, exchange experiences and ideas. Thanks!

Fannon: @ognis1205  - since you plan to change the .well-known URI of agent cards anyway, have you thought about allowing multiple agent cards being returned in one JSON file there? 

It may also not possible for everyone to implement a `.well-known/...` URI on their root, because they may not have access to that. In that case, there has to be some kind of fallback - like registering your full config URL at the registry to collect it, or some means to push the agent cards to a registry.

ognis1205: Thanks for your thoughts, @Fannon !

As a matter of fact, the proposal and discussion in [#166](https://github.com/a2aproject/A2A/discussions/166) actually originated from a very similar idea — exposing a list of agent cards under the `.well-known` path.

While returning the agent cards directly from `.well-known` is a simple and fast approach, I’ve come to believe there are some advantages in using a more scalable design. That’s why the current proposal suggests returning only the list of endpoints (where each individual Agent Card can be retrieved) — specifically in the form of an [API Catalog](https://datatracker.ietf.org/doc/rfc9727/)-like format.

This has two main benefits:

 - It allows for better scalability and maintenance, since each Agent Card can be managed independently.
 - It enables domain non-owners (e.g., project contributors) to keep their cards up-to-date, as long as the domain owner lists their endpoints once.

As for uploading or notifying the registry of an Agent Card — if done via Open Discovery — using this API Catalog format should suffice. For other methods (such as when `.well-known` is not available), that probably falls into the scope of Agent Registry or Agent Catalog API/protocols, which we may want to define separately.

Fannon: I see. With this approach, you also return the agent cards under a `.well-known` path, just with the difference that you first return a list of links, which point to the agent cards (one indirection). 

But I agree that this approach scales better for bigger scenarios, because you first have a smaller list of everything that is available and then you can retrieve the metadata for just those resources you are really interested in. That's basically how it works in ORD, too. 

Question: This format is meant only for publishing agent card to the registry - or would the registry itself also have the same implementation to discover what's inside of it? Or will this be a separate, dedicated API?

ognis1205: Thanks for the thoughtful question.

A few clarifications:
 1. With this format, Agent Cards are not required to live under the `.well-known` path. The `href` fields in the response can point to any arbitrary path, which gives publishers flexibility in structuring their resources.
 2. The API Catalog format is based on IETF specifications for Open API discovery. It’s not designed solely for publishing Agent Cards to a registry. Any third party aware of this format can consume and use the information for their own purposes. Whether a registry adopts the same discovery mechanism internally is entirely up to the registry implementer.
  3. I understand the name "API Catalog" might be a bit misleading — it’s not an actual API or a dedicated endpoint for serving data like a typical REST-style service. Instead, it refers to a declarative document that catalogs available agent interfaces in a structured format. So it’s conceptually very different from something like a REST catalog or a discoverable API registry.

If the goal is to discuss this in the context of the ORD project and registry discovery, I’d strongly recommend continuing the conversation in the Agent Registry discussion I shared earlier. This PR, from a discovery standpoint, is orthogonal to the registry: it focuses on **open** discovery specifically.

The current A2A Protocol spec assumes a "one agent per host" model for open discovery. This proposal tries to relax that constraint by introducing a more flexible alternative. So rather than being about publishing to the registry or registry introspection, it’s about enabling more expressive `.well-known` discovery for individual agents or agent collections hosted under the same domain.

That said, I feel there might be a slight misalignment between what we’re proposing in this PR and the assumptions behind your question. We’re happy to clarify further if needed!

ognis1205: @Fannon 

https://open-resource-discovery.github.io/specification/spec-v1/interfaces/document-api

Is this where I can find the official specification for the Open Discovery document API in ORD?

ognis1205: I’m still learning about ORD and may be missing some important context, but I find it to be a well-thought-out specification.
My goal is to suggest something simple and fair for the broader community, and I’d love to continue learning and discussing this with you—would that be okay?

ognis1205: Regarding the questions you raised:

> This format is meant only for publishing agent card to the registry?

At the time of drafting this PR, discussions around an Agent Registry or Catalog had not yet begun. The primary motivation was to address a limitation in the A2A Protocol — specifically, its implicit assumption of a one-agent-per-host model for open discovery.

Therefore, this format is not designed solely for publishing Agent Cards to a registry. Rather, it is intended as a general-purpose mechanism to support open discovery, including but not limited to registry scenarios.

> Or would the registry itself also have the same implementation to discover what’s inside of it?

This PR does not assume that registries will adopt the exact same implementation or format for introspection. In fact, my original thinking was that a registry would likely expose a different protocol, with more emphasis on authentication, validation, and governance, which are outside the scope of open discovery.

That said, registries could choose to consume these open discovery documents — but doing so is entirely up to their implementation.

> Or will this be a separate, dedicated API?

Yes — that was the original intention when drafting this proposal. I envisioned registry interactions as being handled via a separate and dedicated API, distinct from the open discovery format described here.

Fannon: > @Fannon
> 
> https://open-resource-discovery.github.io/specification/spec-v1/interfaces/document-api
> 
> Is this where I can find the official specification for the Open Discovery document API in ORD?

Yes, but the bulk is actually in the ORD Document interface: https://open-resource-discovery.github.io/specification/spec-v1/interfaces/document 

Most of it is not relevant here, but it has a collection of "API Resources", where you can describe the existence of APIs of various protocol types incl. MCP or A2A (soon). But the actual metadata definitions are provided as a link, e.g. to an Agent Card file, or an OpenAPI document - whatever is relevant to the protocols. 
Technically it's 3 steps: First the `.well-known/open-resource-discovery`, which lists the ORD documents (and how they can be accessed), second the ORD documents and third the various metadata definition files (usually existing industry standards)

Fannon: > My goal is to suggest something simple and fair for the broader community, and I’d love to continue learning and discussing this with you—would that be okay?

Sure that totally makes sense, I was actually just interested in your approach and your thinking behind it. If you like we can schedule a call and exchange directly?

darrelmiller: As Linux Foundation is on the list of standards organizations approved by the IESG, it would be possible to register an Agent Card media type in the IANA Media Types standards tree.  e.g. application/agent-card+json as long as their is a stable URL to the specification.
As one of the reviewers for the IANA media type registry I can help with this process.

Fannon: > Yes — that was the original intention when drafting this proposal. I envisioned registry interactions as being handled via a separate and dedicated API, distinct from the open discovery format described here.

@ognis1205 - I think this is a good idea, too. Then the discovery / publishing interface can be kept simple (as in this PR) and the registry API can be more extensive and feature rich, for a good consumption experience.

ognis1205: @Fannon , sorry for the late reply.

Yes, personally I take a relatively open stance regarding the format used under `.well-known` for Open Discovery. That said, the reason I advocate for the API Catalog format is because:

 - It is extremely simple and only provides the minimal necessary functionality.
 - The schema definition is based on a web standard (linkset), allowing us to define the specification without reinventing or extending existing ones.
 - The RFC explicitly discusses how to operate when coexisting with other formats, particularly through content negotiation.

Regarding the last point, although I don’t fully understand all the details of the ORD specification yet, I believe that—even if the format doesn’t use the recommended path `.well-known/open-resource-discovery`—as long as it uses a proper Content-Type (e.g., `application/vnd.sap.ord+json`) and that type is accepted by the community, it should be possible to support ORD-compliant Open Agent Discovery via API Catalog specification.

That said, I’m open to continuing this discussion further—and perhaps this is not just about Open Discovery, but also a good opportunity to discuss the broader protocol around Agent Catalogs and Registries as well.

SEE ALSO:
https://datatracker.ietf.org/doc/rfc9727/

Fannon: @ognis1205 - right now we don't have a defined media type for ORD documents, but maybe it's a good idea to define this for the reason you outlined!

Allthough we then have the situation that there are many `.well-known` entry points. If you support a specific discovery mechanism, you would anyway go for the corresponding well-known URI, e.g. if you want to discover A2A agent cards, you might want to try with `.well-known/agent-card.json` first anyway - but then there's the limitation that you can get only one. So you try with `/.well-known/api-catalog.json`, where you can get more than one A2A agent card, but also potentially links to other API definitions like OpenAPI. 

For ORD, you start with `/.well-known/open-resource-discovery`, which would work the same as the api-catalog, but have more information about the APIs on ORD level already and support describing other aspects about the application. So I'm not sure if we would need to interlink the discovery mechanisms, as each has a well-known entry-point anyway. But I would expect that you would share the same link to the same definition files, e.g. the agent-card JSON files.

Or do you see value of linking those discovery mechanisms more closely together? E.g. from ORD config side, we could point to a `/.well-known/api-catalog.json` - but not sure of the value in it, as it would not contain more information than we would have in ORD already. And if you don't understand ORD, you wouldn't have called its entry point in the first place. (Just thinking by writing)

ognis1205: @Fannon , thanks for the thoughtful response — I realize that my PR, or the documentation so far, might not have clearly communicated some of the assumptions or motivations behind the proposal. If you don’t mind, I’d like to clarify three specific points that may have caused misunderstanding:

1. .well-known/api-catalog (not .json)

It might seem like a minor detail, but the proposed API Catalog endpoint is actually `.well-known/api-catalog`, not `.well-known/api-catalog.json`.

This is important in the context of content negotiation, as described in [RFC 9727 Section 4.2](https://www.rfc-editor.org/rfc/rfc9727#section-4.2):

> The Publisher MAY make additional formats available via content negotiation to their /.well-known/api-catalog location.

So while `application/linkset+json` is the default required format, servers can serve alternative representations via standard content negotiation. This is a central design point of this proposal.

2. Multiple Open Discovery Endpoints: Rationale

Part of the motivation behind this proposal stems from concerns around the current `.well-known/agent.json` endpoint, which implicitly assumes a one-agent-per-host model. I believe this is problematic, and would recommend either:

 - removing the .well-known-based discovery from the A2A spec entirely, or
 - replacing it with a more flexible mechanism (such as this PR proposes).

Note that I intentionally left `.well-known/agent.json` out of the scope of this PR to keep the discussion focused.
That said, it’s my opinion that the need for `.well-known/agent.json` could become obsolete if the community decides to adopt `.well-known/api-catalog` — though this is certainly open to discussion.

So, while you mentioned the possibility of clients trying both `.well-known/agent-card.json` and `.well-known/api-catalog` — if we go with the catalog-based discovery, I would argue that `.well-known/agent.json` should not be recommended or implemented at all.

The choice of discovery mechanism should be left up to the Agent Provider, and the spec should RECOMMEND—but not REQUIRE—any specific one.

3. Not Interlinking Discovery Mechanisms — but Supporting Content Negotiation

Finally, to clarify one more possible misunderstanding:

I’m not proposing interlinking different discovery mechanisms (e.g., linking from ORD to API Catalog or vice versa). Instead, I’m suggesting content negotiation as the unifying mechanism.

To illustrate, a client could perform:

```
GET https://agent.org/.well-known/api-catalog  
Accept: application/vnd.sap.ord+json
```

In this way, a single `.well-known/api-catalog` endpoint can support ORD or any other discovery format — without needing separate entry points.

This does not contradict the ORD spec, and in fact, enables ORD-based discovery to be bootstrapped from a single catalog endpoint — while remaining compatible with RFC 9727.

cc @darrelmiller — If you happen to have time, I’d be grateful for your perspective, especially if I’ve misunderstood any part of API Catalog or its use of content negotiation.


> Although we then have the situation that there are many .well-known entry points…

You’re right — proliferation of .well-known endpoints can be a concern.
However, just to clarify: the proposed path is `.well-known/api-catalog` (not .json), and supports [content negotiation per RFC 9727 §4.2](https://www.rfc-editor.org/rfc/rfc9727#section-4.2). This allows a single endpoint to serve multiple representations — so clients don’t need to try multiple `.well-known/*` paths.

> For ORD, you start with /.well-known/open-resource-discovery, which would work the same as the api-catalog…

Actually, you don’t need to use `.well-known/open-resource-discovery` if `.well-known/api-catalog` supports content negotiation. For example, a client could send:

```
GET /.well-known/api-catalog  
Accept: application/vnd.sap.ord+json
```

This enables API Catalog to serve as the single discovery entry point — even for ORD — without violating the ORD spec.

> So I’m not sure if we would need to interlink the discovery mechanisms…

Agreed — and to clarify, this proposal does not suggest interlinking between mechanisms.
Instead, it promotes content negotiation from a single `.well-known/api-catalog` as a clean, extensible solution. Each protocol can co-exist via negotiated media types, without needing explicit links between them.

ognis1205: As an additional note, while the specific schema for this has not yet been defined, I personally believe that even when relying on an external Agent Registry, it should be possible to expose the Registry’s endpoint as part of Open Discovery by leveraging the linkset format used in the API Catalog.

In other words, linkset allows us to declaratively indicate where clients can interact with a registry—making it a natural part of the discovery mechanism itself. This could help unify discovery and registration workflows, even across multiple registry implementations.

If this approach proves viable, `.well-known/api-catalog` could serve not only as the central entry point for Open Discovery, but also as a bootstrap layer for federated Agent Registry infrastructure in the future.

ognis1205:
> As Linux Foundation is on the list of standards organizations approved by the IESG, it would be possible to register an Agent Card media type in the IANA Media Types standards tree. e.g. application/agent-card+json as long as their is a stable URL to the specification.
> As one of the reviewers for the IANA media type registry I can help with this process.

@darrelmiller cc: @kthota-g @holtskinner 

Regardless of which Open Discovery specification we ultimately adopt and/or recommend, I believe it makes sense to proceed with registering `application/agent-card+json` in the IANA Media Types registry.

While I’m not sure who the best point of contact would be for this work, I’d be happy to assist with the process if needed. What do you think?

docs/specification.md
ognis1205: @Fannon 

Additional note:

> So I’m not sure if we would need to interlink the discovery mechanisms…

I also don’t believe that simply interlinking various Open Discovery Mechanisms inherently provides much value. At this stage, I think it’s perfectly reasonable for each provider—or the A2A community at large—to adopt whichever Open Discovery Mechanism best suits their network scale or preferences.

That said, once the surrounding technologies in the A2A ecosystem mature further, I do think there remains room for discussion—and perhaps open questions—on how the A2A protocol, or a future A2A Discovery protocol, might formally incorporate or standardize the use of Open Discovery mechanisms.

Fannon: @ognis1205 sorry for the late reply. 

I'm not sure it makes sense to include ORD into the .well-known API catalog discovery, because ORD has a wider scope, it's not just an API catalog, it describes an application / service to external consumers in general. And since ORD already has a .well-known URI, there wouldn't be a real advantage of putting it in there, imho. 

But that doesn't really affect your PR / decisions here, since A2A agent cards do fall well within the api-catalog scope.

> That said, it’s my opinion that the need for .well-known/agent.json could become obsolete if the community decides to adopt .well-known/api-catalog — though this is certainly open to discussion.

Yes, I think this decision need to be made. The `.well-known/api-catalog` is more versatile and can also mesh with other API discovery. I would agree that it makes the `.well-known/agent.json` / `.well-known/agent-card.json` obsolete. To avoid confusions, it may be a good idea to deprecate it in favor of the `.well-known/api-catalog`.

ognis1205: Thanks for the thoughtful input, @Fannon ! Don’t worry about the late reply.

Just to clarify, I’m not at all opposed to ORD. That said, my priorities around changes in this area are roughly the following:

 1. Update the language in the [current spec](https://a2a-protocol.org/latest/topics/agent-discovery/#discovery-strategies)—specifically, this part:
 
> This is a recommended approach for public agents or agents intended for broad discoverability within a specific domain.

I’d suggest adjusting the wording to reflect that there are now multiple valid `.well-known`-based approaches, including ORD and API Catalog. Among those, the A2A community has chosen to register `.well-known/agent-card.json` with IANA, and links to other recommended options could also be provided for broader context.

 2. Encourage use of a `.well-known`-based discovery mechanism that supports multi-agent scenarios under a single host, such as ORD or the API Catalog. The spec could recommend one of these options and provide a formal schema, as this PR begins to do.

Of these, I personally think (1) would be the most pragmatic and forward-compatible approach for the A2A community.

I’ll try to open a discussion thread later on to explore this direction more collaboratively.

ognis1205 (commented):
~~TODO:~~

~~Migrate this definition to the protobuf schema according to~~
~~https://github.com/a2aproject/A2A/blob/33f065d3daf6c4bc6f0ada184bab149eadc1985a/specification/json/README.md.~~

Done.

ognis1205:
cc: @darrelmiller 

I've reviewed the recording and comments from the last TSC meeting. My understanding is that, regarding the A2A (Open) Discovery specification, there seems to be a direction toward not recommending the use of .well-known URLs in the future.

Given that this PR keeps the references publicly open, I wanted to check with you whether it would be better to close this PR for now.

Thanks in advance for your guidance!

darrelmiller:
These requirements are now being met by efforts like http://ai-catalog.io/ and https://agenticresourcediscovery.org/

PR #647, opened 2025-05-24
Fix for Incorrect Type Annotation for uri Field in FileWithUri Interface #646

### LINKED ISSUE #646: [Bug]: Incorrect Type Annotation for uri Field in FileWithUri Interface
### What happened?

````typescript
interface FileWithUri {
  // The original filename, if known (e.g., "document.pdf", "avatar.png").
  name?: string;
  // The [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
  // of the file (e.g., "application/pdf", "image/png"). Strongly recommended for proper handling.
  mimeType?: string;
  // URI for the file.
  uri: string; // Base64 string
}
````

The uri field in the FileWithUri interface is currently documented with the following incorrect comment:
````typescript
// URI for the file.
uri: string; // Base64 string
````

This is misleading and incorrect. 

The comment states that the value should be a Base64 string, whereas the uri field is meant to be a web-accessible URI (e.g., https://example.com/file.png, data:image/png;base64,..., or a local object URI in an unlikely scenario where all the agents are running locally on the same machine). This error can cause confusion for implementers or lead to incorrect assumptions that the field must contain a Base64-encoded URI or file content instead of a link.

Suggested Fix:

````typescript
uri: string; // e.g., "https://files.example.com/document.pdf" or "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA..."

````

I am going to be submitted a PR shortly to address this.

### Relevant log output

```shell

```

### Code of Conduct

izzymsft:
- Changed field comments for the uri field.
- Added some example values for the field

Fixes #646

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -617,8 +617,9 @@ interface FileWithUri {
   // The [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
   // of the file (e.g., "application/pdf", "image/png"). Strongly recommended for proper handling.
   mimeType?: string;
-  // URI for the file.
-  uri: string; // Base64 string
+  // A web-accessible URI for the file.
+  // It can be a HTTPS URL or Data URI. URL strongly recommended for larger files that should not be embedded directly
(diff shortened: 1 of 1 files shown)


holtskinner:
Fixed in #663

PR #695, opened 2025-05-30
feat: Add gRPC and REST definitions to A2A protocol specifications

pstephengoogle:
- Requires adding a new `PushNotificationConfig.id` field to support multiple push notification callbacks per task


specification/grpc/a2a.proto
```diff
@@ -0,0 +1,607 @@
+edition = "2023";
+
+package a2a;
```
AlbumenJ: Would it be better to update it to a package that carries the version meaning, such as `a2a.v1`? If a new version is released in the future, iteration will be smoother.

specification/grpc/a2a.proto
```diff
@@ -0,0 +1,607 @@
...
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+import "google/api/annotations.proto";
+import "google/api/field_behavior.proto";
+import "google/api/client.proto";
+
+option java_multiple_files = true;
```
AlbumenJ, line 11:
```suggestion
option csharp_namespace = "A2a.V1";
option go_package = "google.golang.org/a2a/v1";
option java_multiple_files = true;
option java_outer_classname = "A2A";
option java_package = "com.google.a2a.v1";
```

Leaving the package name here will help the SDK implementation.
The above content is referenced from [health.proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto). Some packages may need optimization.

AlbumenJ (commented):
It’s exciting to see the A2A protocol building a high-performance communication protocol base on gRPC. 🥳🥳🥳

third_party/google/api/annotations.proto
holtskinner, line 1: Question here - Should these all be combined in the a2a.proto or in the grpc directory?

specification/grpc/a2a.proto
```diff
@@ -0,0 +1,607 @@
...
+// The set of states a Task can be in.
+enum TaskState {
+  TASK_STATE_UNSPECIFIED = 0;
+  // Represents the status that acknowledges a task is created
+  TASK_STATE_SUBMITTED = 1;
+  // Represents the status that a task is actively being processed
+  TASK_STATE_WORKING = 2;
+  // Represents the status a task is finished. This is a terminal state
+  TASK_STATE_COMPLETED = 3;
+  // Represents the statas a task is done but failed. This is a terminal state
```
holtskinner:
```suggestion
  // Represents the status a task is done but failed. This is a terminal state
```

specification/grpc/a2a.proto
```diff
@@ -0,0 +1,607 @@
...
+  string name = 1;
+  // A description of the agents domain of action/solution space.
+  string description = 2;
+  // The base url to interact with the agent at.
+  string url = 3;
+  // Information about the service provider of the agent.
+  AgentProvider provider = 4;
+  // The version of the agent.
+  string version = 5;
+  // A url to provide additional documentation about thea agent.
```
holtskinner:
```suggestion
  // A url to provide additional documentation about the agent.
```

holtskinner (changes requested):

types/src/types.ts
```diff
@@ -389,6 +389,8 @@ export interface PushNotificationAuthenticationInfo {
 // --8<-- [start:PushNotificationConfig]
 /**Configuration for setting up push notifications for task updates. */
 export interface PushNotificationConfig {
+  /** Push Notification Id - created by server to support multiple callbacks **/
```
```suggestion
  /** Push Notification ID - created by server to support multiple callbacks **/
```

third_party/google/api/client.proto
```diff
@@ -0,0 +1,486 @@
...
+  map<string, string> renamed_services = 2;
+}
+
+// Describes the generator configuration for a method.
+message MethodSettings {
+  // Describes settings to use when generating API methods that use the
+  // long-running operation pattern.
+  // All default values below are from those used in the client library
+  // generators (e.g.
+  // [Java](https://github.com/googleapis/gapic-generator-java/blob/04c2faa191a9b5a10b92392fe8482279c4404803/src/main/java/com/google/api/generator/gapic/composer/common/RetrySettingsComposer.java)).
```
holtskinner: This looks like an old version of the gapic-generator, should this be changed to use a newer version?

third_party/google/api/annotations.proto
pstephengoogle: We have a follow on PR to pull these from the source automatically, using ruf. We will commit this (I'll remove these third_party files) and then the other PR. This should allow the protoc scripts to be run in the relevant sdk package to generate the language specific parts

third_party/google/api/client.proto
pstephengoogle: it will, we will automatically pull from the source via ruf

specification/grpc/a2a.proto
pstephengoogle: done

specification/grpc/a2a.proto
pstephengoogle: added

third_party/google/api/client.proto
pstephengoogle: removed the third_party references in lieu of PR#702

specification/grpc/a2a.proto
```diff
@@ -0,0 +1,612 @@
...
+// Of particular note of deviation from JSONRPC approach the request metadata
+// fields are not present as they don't comply with AIP rules, and the
+// optional history_length on the get task method is not present as it also
+// violates AIP-127 and AIP-131.
+service A2AService {
+  // Send a message to the agent. This is a blocking call that will return the
+  // task once it is completed, or a LRO if requested.
+  rpc SendMessage(SendMessageRequest) returns (SendMessageResponse) {
+    option (google.api.http) = {
+      post: "/v1//message:send"
```
connyay, line 37: was this double slash intentional?

pstephengoogle: No, thanks for catching this. Will send a PR

pstephengoogle: https://github.com/google-a2a/A2A/pull/747 - PTAL

PR #704, opened 2025-06-02
docs: update specification to use "kind" (not "type") for Part objects

vicradon:
# Description

1. Several example flows in the documentation still specify "type" for Part objects
2. Since this was changed from "type" to "kind" in version 0.2.1, this change must be reflected


holtskinner (approved):
Thanks for the fix!

PR #725, opened 2025-06-04
chore(grpc): Change `a2a.proto` from proto3 to edition 2023

holtskinner:


specification/grpc/a2a.proto
```diff
@@ -1,5 +1,4 @@
-// edition = "2023";
-syntax = "proto3";
+edition = "2023";
```
rajeshvelicheti, line 1: editions do not have proper generation support yet and I had to move to proto3, any reason why we want to use editions?

rajeshvelicheti (changes requested):
editions do not have proper generation support yet and I had to move to proto3, any reason why we want to use editions?

specification/grpc/a2a.proto
holtskinner: I saw that you commented this out. I figured it made sense to use the current standard, but if there's support issues for generation with editions, we can keep as is.

PR #734, opened 2025-06-06
feat: Support for multiple `pushNotification` configurations for task [DO NOT MERGE]

kthota-g:
- Support for multiple pushNotification configurations for task

Release-As: 0.2.3

### DIFF
--- a/types/src/types.ts
+++ b/types/src/types.ts
@@ -92,6 +92,11 @@ export interface AgentSkill {
  */
 export interface AgentCard {
   /**
+   * Universally Unique identifier for the agent. 
+   * @pattern ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$
+   */
+  id: string;
+  /**
    * Human readable name of the agent.
    * @example "Recipe Agent"
    */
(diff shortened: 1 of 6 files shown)


types/src/types.ts
```diff
@@ -880,8 +885,8 @@ export interface GetTaskPushNotificationConfigRequest extends JSONRPCRequest {
  */
 export interface GetTaskPushNotificationConfigSuccessResponse
   extends JSONRPCSuccessResponse {
-  /** The result object on success. */
-  result: TaskPushNotificationConfig;
+  /** The result object on success.*/
+  result: TaskPushNotificationConfig[];
```
swapydapy, line 884: Given we can now set multiple pushNotif configs, hoes does someone remove an existing one?

swapydapy (changes requested):

types/src/types.ts
kthota-g: updating an existing config can be done with `tasks/pushNotificationConfig/set` method, by passing the payload with ID.

Remove may require a new method or make [url attribute](https://github.com/google-a2a/A2A/blob/968c650cfe10d90e5639b7469be5e6f732392f6a/types/src/types.ts#L417) optional to signal removing a config.

swapydapy: Yep. As discussed, I vote for having another method: removePushNotificationConfig.
SetPushNotificationConfig: If not id is specified, then handler creates a new config and attaches to the task.
SetPushNotificationConfig: If id is specified, then this is treated as full update. Both url and auth will be validated and updated. Partial updates can be supported by fieldMask.

GetPushNotificationConfig: Gets all configs for a task.

kthota-g:
Closing this PR. I will create a new PR to support list, get, remove pushNotificationConfigurations

PR #749, opened 2025-06-12
feat: Add support for multiple transport announcement in AgentCard

pstephengoogle:
This is designed to be backward compatible. Existing cards will work and be correct with this update. New fields are added to the agent card and default interpretation for these are provided to be backward compatible


types/src/types.ts
```diff
@@ -102,8 +118,18 @@ export interface AgentCard {
...
    */
   description: string;
-  /** A URL to the address the agent is hosted at. */
+  /**
+   * A URL to the address the agent is hosted at. This represents the
+   * preferred endpoint as declared by the agent.
+   */
   url: string;
+  /* The transport of the preferred endpoint. If empty, defaults to JSONRPC.*/
+  preferred_transport?: string;
```
kthota-g, line 127: adding `@default` annotation would add the JSONRPC as the defautl value to the schema

swapydapy (approved):

specification/grpc/a2a.proto
```diff
@@ -326,9 +336,16 @@ message AgentCard {
   string name = 1;
   // A description of the agents domain of action/solution space.
   string description = 2;
-  // The base url to interact with the agent at.
+  // A URL to the address the agent is hosted at. This represents the
+  // preferred endpoint as declared by the agent.
   string url = 3;
+  // The transport of the preferred endpoint. If empty, defaults to JSONRPC.
   // Information about the service provider of the agent.
```
dmandar: delete this AgentProvider comment. You have added a new comment below.

holtskinner (dismissed):
When you commit new updates, it should re-run the checks. (They might have failed due to the outage yesterday)

types/src/types.ts
pstephengoogle: Discussed offline, leaving as is to support new client to old card semantics

PR #758, opened 2025-06-15
docs: Add `@TJS-examples` for `AgentProvider` and `AgentExtension` interfaces

ZapzCsgo:
Add missing JSDoc `@example` annotations to `AgentProvider` and `AgentExtension` interfaces to improve TypeScript developer experience.

Examples use Google ecosystem references (ai.google.dev, OAuth 2.0) and follow existing JSDoc patterns in the codebase.

First contribution to open source ❤️

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -56,6 +56,7 @@ RPCs
 RUF
 SLAs
 SLF
+TJS
 TMDB
 URLTo
 Upserting
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -222,6 +222,13 @@
         },
         "AgentExtension": {
             "description": "A declaration of an extension supported by an Agent.",
+            "examples": [
+                {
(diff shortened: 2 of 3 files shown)


holtskinner:
@pstephengoogle I'm in favor of adding this, however these examples don't get ported into the a2a.json file or any other types files.

ZapzCsgo:
> @pstephengoogle I'm in favor of adding this, however these examples don't get ported into the a2a.json file or any other types files.

Thanks for the feedback, I will check this evening the integration of the examples in a2a.json and in the other type files so that it is clean

ZapzCsgo:
After digging, I found that this is not feasible given the current project architecture. I misunderstood your statement about A2A generation in your answer.

holtskinner:
@ZapzCsgo Looks like `@TJS-examples` is the recommended way to do this in TypeScript, and it does get populated in the a2a.json file

holtskinner (approved):

PR #770, opened 2025-06-19
feat: update spec & doc topic with non-restartable tasks

swapydapy:


docs/topics/life-of-a-task.md
holtskinner, line 1: I wonder if this would fit better as a section in `key-concepts.md`?

If you'd rather keep it separate, then be sure to add it to the navigation in `mkdocs.yml`

swapydapy: Good catch. Would like to keep it as separate, since this goes into more detail. Added links into key-concepts.

docs/topics/key-concepts.md
```diff
@@ -70,6 +71,9 @@ The Agent2Agent (A2A) protocol is built around a set of core concepts that defin
     - This requires the server to advertise the `pushNotifications` capability in its Agent Card.
     - Learn more about [Streaming & Asynchronous Operations](./streaming-and-async.md).
 
+## Agent Response: Task or Message
```
holtskinner, line 74:
```suggestion
## Agent Response: Task or Message

```

holtskinner (approved):

PR #786, opened 2025-06-24
chore: Add exceptions for buf lint and upgrade buf.yaml to v2

bufdev:
This PR does a few cleanups to `buf.yaml`:

- Upgrades `buf.yaml` to [`v2`](https://buf.build/blog/buf-cli-next-generation), which is what is generally used going forward, and matches the `buf.gen.yaml` version. In this case, virtually no changes.
- Adds lint exceptions for those lint rules that do not pass `buf lint`. This now ensures that `buf lint` passes.
- Updates a few comments in `buf.yaml` for clarification.
- Updates the `README.md` to use `buf lint` instead of `buf build`. Note that `buf build` is implicitly run via `buf lint`.
- Removes the instruction about checking for breaking changes, as there's no mechanism for this set up yet. Happy to contribute a PR to make this simple if this would be desirable.
- Removes the instruction to modify `buf.yaml`. `buf.yaml` should generally never need modification.


holtskinner:
@pstephengoogle Can you take a look at this?

holtskinner (approved):

holtskinner:
Verified that the generation of the grpc files in https://github.com/a2aproject/a2a-python work as expected.

PR #802, opened 2025-06-27
feat(spec)!: Add a required protocol version to the agent card.

dmandar:
Add a new protocol version field to the agent card.

Release-As: 0.2.5

### DIFF
--- a/specification/json/a2a.json
+++ b/specification/json/a2a.json
@@ -183,6 +183,9 @@
                 },
                 "protocolVersion": {
                     "description": "The version of the A2A protocol this agent supports.",
+                    "examples": [
+                        "0.2.5"
+                    ],
                     "type": "string"
                 },
                 "provider": {


specification/grpc/a2a.proto
```diff
@@ -341,6 +341,8 @@ message AgentInterface {
 // - Default modalities/content types supported by the agent.
 // - Authentication requirements
 message AgentCard {
+  // The version of the A2A protocol this agent supports.
+  string protocol_version = 16;
```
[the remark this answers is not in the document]
holtskinner: @dmandar I agree with this. Can you put it at the end?

types/src/types.ts
```diff
@@ -109,6 +109,11 @@ export interface AgentInterface {
  * - Authentication requirements
  */
 export interface AgentCard {
+  /**
+   * The version of the A2A protocol this agent supports.
+   * @example "0.2.9"
```
holtskinner: `@example` doesn't add examples to the resulting JSON, `@TJS-examples` does. In another PR, we should convert these examples to use the TJS format.

```suggestion
   * @TJS-examples "0.2.9"
```

holtskinner: Can you also put this new field as the last property of the `AgentCard` format to match the protobuf? Or we should change the protobuf field numbering to have `protocol_version` as the first field (I think it makes sense to have it first)

holtskinner (changes requested):

holtskinner (approved):

PR #811, opened 2025-06-30
chore(spec): Add JSDoc `@format` `@minValues` and `readonly` for properties

holtskinner:


types/src/types.ts
```diff
@@ -139,16 +150,23 @@ export interface AgentCard {
...
+   * @format uri
+   */
   iconUrl?: string;
   /** The service provider of the agent */
   provider?: AgentProvider;
   /**
-   * The version of the agent - format is up to the provider.
+   * The version of the agent. Semantic Versioning MUST be used.
    * @TJS-examples ["1.0.0"]
+   * @pattern ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-\d][0-9a-zA-Z-\d]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-\d][0-9a-zA-Z-\d]*))*))?(?:\+([0-9A-Za-z-\d]+(?:\.[0-9A-Za-z-\d]+)*))?$
```
pstephengoogle: Do we need to be this precise? Can't we just define the characters allowed and not the sequencing requirements of these characters?

types/src/types.ts
```diff
@@ -158,12 +176,19 @@ export interface AgentCard {
...
-  /** Supported media types for output. */
+  /**
+   * Supported media types for output. Must contain at least one mode.
+   * @minItems 1
+   */
   defaultOutputModes: string[];
-  /** Skills are a unit of capability that an agent can perform. */
+  /**
+   * Skills are a unit of capability that an agent can perform. An agent must have at least one skill.
+   * @minItems 1
```
pstephengoogle: I don't know we have a min number skills requirement

types/src/types.ts
```diff
@@ -290,7 +316,10 @@ export interface DeleteTaskPushNotificationConfigParams extends TaskIdParams {
 // --8<-- [start:MessageSendConfiguration]
 /**Configuration for the send message request. */
 export interface MessageSendConfiguration {
-  /** Accepted output modalities by the client. */
+  /**
+   * Accepted output modalities by the client.
+   * @minItems 1
+   */
   acceptedOutputModes?: string[];
```
pstephengoogle: Do we require this to be non-zero? I think if the caller doesn't specify the server can return whatever makes sense. Up to the client to handle from there

types/src/types.ts
holtskinner: This is a regex to detect Semantic Versioning, e.g. `1.0.0` It's probably a good idea to enforce consistency?

types/src/types.ts
holtskinner: If that's the case, then should skills be optional?

types/src/types.ts
holtskinner: This change was originally made before this field became optional. However, it could make some sense to require at least one value if it's provided. Is there a benefit to providing an empty array?

types/src/types.ts
```diff
@@ -144,17 +157,23 @@ export interface AgentCard {
...
+   */
   iconUrl?: string;
   /** Information about the agent's service provider. */
   provider?: AgentProvider;
   /**
-   * The agent's own version number. The format is defined by the provider.
-   *
+   * The agent's own version number. Semantic Versioning MUST be used.
    * @TJS-examples ["1.0.0"]
+   * @pattern ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-\d][0-9a-zA-Z-\d]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-\d][0-9a-zA-Z-\d]*))*))?(?:\+([0-9A-Za-z-\d]+(?:\.[0-9A-Za-z-\d]+)*))?$
```
```suggestion
   * @pattern ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```

holtskinner:
Cancelling PR, caused too many downstream breaking changes.

PR #814, opened 2025-06-30
feat(spec): add optional `inputFields` and `outputFields` and  to `AgentSkill`

### LINKED ISSUE #813: [Feat]: Optional Structured Field Definitions for Agent Inputs and Outputs

Yes. The protocol currently supports `inputModes` and `outputModes` (e.g., supported mime types) for agent skills, but lacks a structured way for agents to optionally declare named input and output fields they expect or produce, along with their types or schemas.

This makes it harder for clients to know the exact structure and meaning of inputs an agent expects or outputs it produces, especially when multiple fields of the same type are involved.

### Motivation

- Enable agents to declare **named input/output fields** with type information.
- Support **schemas** for better validation, logging, debugging, etc
- Let clients render structured forms to collect user input appropriately.
- Reduce ambiguity in how message/artifact parts are composed and interpreted.
- Provide a more **semantic layer** on top of the existing input/output modes.

Introduce optional structured field definitions for agent input and output, using a new `FieldDefinition` model. Agents can declare named fields, their content kind (text, file, data), optional mime types or schemas, and a description. These would complement the `inputModes`/`outputModes` on the `AgentSkill`

```python
from pydantic import BaseModel
from typing import Literal, Any, List, Optional

class FieldDefinition(BaseModel):
    """
    Describes an expected input or output field.
    """
    name: str | None = None
    kind: Literal['text', 'file', 'data'] | None = None
    mimeTypes: list[str] | None= None
    schema: dict[str, Any] | None = None  # For data parts
    description: str | None = None
    optional: bool | None = False

class AgentSkill(BaseModel):
    """
    Represents a unit of capability that an agent can perform.
    """

    description: str
    examples: list[str] | None = None
    id: str
    name: str
    tags: list[str]

    # Existing simple mime-type lists remain supported:
    inputModes: list[str] | None = None
    outputModes: list[str] | None = None

    # New optional structured fields for richer metadata:
    inputFields: list[FieldDefinition] | None = None
    outputFields: list[FieldDefinition] | None = None
```

Message senders can then match named parts accordingly. The protocol may also define conventions for how these fields map to a `Part.name` or `DataPart.data.keys`.

#### Example: Book Review Agent

An agent expects two text inputs:

- bookTitle: The book’s title
- authorName: The author’s name

And returns two text outputs:

- review-chunk: A chunk of the book's review
- publicationDate: When the book was published

```json
{
  "inputModes": ["text/plain"],
  "outputModes": ["text/plain", "text/html"],
  "inputFields": [
    { "name": "bookTitle", "kind": "text", "description": "The title of the book" },
    { "name": "authorName", "kind": "text", "description": "The author's full name" }
  ],
  "outputFields": [
    { "name": "review-chunk",
      "kind": "text",
      "mimeTypes": ["text/plain"],
      "description": "A chunk of the book's review. Sent when streaming",
      "optional": true,
    },
    { "name": "review-html",
      "kind": "text",
      "mimeTypes": ["text/html"],
      "description": "HTML with the whole book's review. Sent when not streaming.",
      "optional": true,
    },
    { "name": "publicationDate", "kind": "text", "description": "Publication date"}
  ],
}
```

Client sending a message to the agent:

```py
message = Message(
    parts=[
        Part(
           root=DataPart(
              data={
                   "bookTitle": TextPart(text="Charlotte's Web"),
                   "authorName": TextPart(text="E. B. White")
              }
        )
    ],
    # other Message fields as needed
)
```

Agent response with Task and Artifacts:

```py
result = Task(
    taskId: "task1",
    contextId="context1",
    final=True,
    status=TaskStatus(state="completed"),
    artifacts=[
        Artifact(
           artifactId="artifact1",
               name="review-html",
               parts=[Part(root=TextPart(text="<div><h3>Review</h3><p>E.B. White’s <em>Charlotte’s Web</em> is a timeless classic that weaves together friendship, loss, and the quiet beauty of ordinary life.</p></div>"))]
        ),
        Artifact(
           artifactId="artifact2",
               name="publicationDate",
                parts=[Part(root=TextPart(text="October 15, 1952"))]
        ),
)
```

Agent **streaming** response with Task and Artifacts:

```py
# First event: Task
result = Task(
    taskId: "task1",
    contextId="context1",
    final=False,
    status=TaskStatus(state="submitted")
)

# Second event: Task artifact update event (with first part of review)
result = TaskArtifactUpdateEvent(
    taskId: "task1",
    contextId="context1",
    artifact=Artifact(
        artifactId="artifact1",
            name="review-chunk",
            parts=[
                Part(root=TextPart(text="E.B. White’s Charlotte’s Web is a timeless classic that.")),
            ]
     )
)


# Third event: Task artifact update event (with publication date)
result = TaskArtifactUpdateEvent(
    taskId: "task1",
    contextId="context1",
    artifact=Artifact(
        artifactId="artifact2",
            name="publicationDate",
            parts=[Part(root=TextPart(text="October 15, 1952"))]
     )
)

# Fourth event: Task artifact update event (with second part of review)
result = TaskArtifactUpdateEvent(
    taskId: "task1",
    contextId="context1",
    artifact=Artifact(
        artifactId="artifact3",
            name="review-chunk",
            parts=[Part(root=TextPart(text="weaves together friendship, loss, and the quiet beauty of ordinary life"))]
     )
)

...
```

- Relying only on `inputModes`/`outputModes`: insufficient for cases where structure matters.
- Encoding structure in skill descriptions or agent documentation: not machine-readable
- Inferring field structure from examples: unreliable and inconsistent.

### Additional context

_No response_

### Code of Conduct

inesmcm26:
# Description

This PR introduces two new optional fields to the `AgentSkill` model - `inputFields` and `outputFields` - allowing agents to define structured, named input and output fields with detailed type and schema information. These fields complement the existing `inputModes` and `outputModes`, which specify supported mime types.

## Problem

Currently, agent skills declare only `inputModes` and `outputModes` as lists of mime types. While this indicates supported data formats (e.g., `"text/plain"`, `"image/png"`), it does **not** convey the **structure or semantics** of the input/output data, such as named fields or their types. In some use cases, this makes it difficult for clients to:

- Understand what specific inputs an agent expects or outputs it produces  
- Properly validate, log, or debug message contents  
- Create user interfaces with meaningful form fields to invoke an agent
- Handle multiple fields of the same type in a structured way

## Solution

This PR adds a new optional layer by defining a `FieldDefinition` model, which lets agents declare:

- Named input and output fields that are expected to be received and returned by a skill
- Field kinds (`text`, `file`, `data`)  
- Supported mime types for each field  
- Optional JSON schemas for structured data fields  
- Field descriptions and optionality flags  

The `AgentSkill` model is extended with two new optional lists: `inputFields` and `outputFields`. These provide a semantic, machine-readable description of the skill interface, while retaining backward compatibility by keeping `inputModes` and `outputModes`.

## Benefits

- Enables clients to build more precise and user-friendly input forms  
- Improves clarity and reduces ambiguity in message composition  
- Facilitates validation, logging, and debugging with schema support  
- Supports complex skills with multiple inputs/outputs of the same type  
- Provides a semantic layer that builds upon existing mime type declarations

Closes #813 

Idea emerged from discussion in: #462

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -240,8 +240,8 @@ Describes a specific capability, function, or area of expertise the agent can pe
 | `description` | `string`            | Yes      | Detailed skill description. [CommonMark](https://commonmark.org/) MAY be used. |
 | `tags`        | `string[]`          | Yes      | Keywords/categories for discoverability.                                       |
 | `examples`    | `string[]`          | No       | Example prompts or use cases demonstrating skill usage.                        |
-| `inputModes`  | `string[]`          | No       | [Deprecated] Accepted Media Types for input.                                   |
-| `outputModes` | `string[]`          | No       | [Deprecated] Produced Media Types for output.                                  |
+| `inputModes`  | `string[]`          | No       | Accepted Media Types for input.                                   |
+| `outputModes` | `string[]`          | No       | Produced Media Types for output.                                  |
 | `inputFields` | `FieldDefinition[]` | No       | Structured input fields expected by this skill.                                |
 | `outputFields`| `FieldDefinition[]` | No       | Structured output fields produced by this skill.                               |
 
@@ -303,9 +303,9 @@ Describes an expected input or output field for an agent skill. This allows agen
       "inputModes": ["text/plain"],
       "outputModes": ["application/json", "application/vnd.geo+json", "text/html"],
       "inputFields": [
-        {"name": "origin", "kind": "text", "optional": "false"},
-        {"name": "destination", "kind": "text", "optional": "false"},
-        {"name": "preference", "kind": "text", "optional": "false"}
+        {"name": "origin", "kind": "text", "optional": false},
+        {"name": "destination", "kind": "text", "optional": false},
+        {"name": "preference", "kind": "text", "optional": false}
       ],
       "outputFields": [
         {
@@ -313,7 +313,7 @@ Describes an expected input or output field for an agent skill. This allows agen
           "kind": "data",
           "mimeTypes": ["application/json", "application/vnd.geo+json", "text/html"],
           "description": "A JSON object containing the computed route details, including geometry, directions, and any relevant metadata.",
-          "optional": "false"
+          "optional": false
         }
       ],
     },
@@ -321,6 +321,11 @@ Describes an expected input or output field for an agent skill. This allows agen
     "id": "custom-map-generator",
       "name": "Personalized Map Generator",
       "description": "Creates custom map images or interactive map views based on user-defined points of interest, routes, and style preferences. Can overlay data layers.",
+      "tags": ["maps", "customization", "visualization", "cartography"],
+      "examples": [
+        "Generate a map of my upcoming road trip with all planned stops highlighted.",
+        "Show me a map visualizing all coffee shops within a 1-mile radius of my current location."
+      ],
       "inputModes": ["application/json"],
       "outputModes": ["image/png", "image/jpeg", "application/json", "text/html"],
       "inputFields": [
@@ -328,7 +333,7 @@ Describes an expected input or output field for an agent skill. This allows agen
           "kind": "data",
           "mimeTypes": ["application/json"],
           "description": "List of locations to plot on the map",
-          "optional": "false"
+          "optional": false
         }
       ],
       "outputFields": [
@@ -337,14 +342,14 @@ Describes an expected input or output field for an agent skill. This allows agen
           "kind": "file",
           "mimeTypes": ["image/png", "image/jpeg"],
           "description": "Rendered map image",
-          "optional": "true",
+          "optional": true,
         },
         {
           "name": "mapData",
           "kind": "data",
           "mimeTypes": ["application/json"],
           "description": "Data representing the map",
-          "optional": "true",
+          "optional": true,
         }
       ]
     }
--- a/docs/tutorials/python/3-agent-skills-and-card.md
+++ b/docs/tutorials/python/3-agent-skills-and-card.md
@@ -18,7 +18,6 @@ Key attributes of an `AgentSkill` (defined in `a2a.types`):
 - `inputModes` / `outputModes`: Supported Media Types for input and output (e.g., "text/plain", "application/json").
 - `inputFields` / `outputFields`: Structured input/output fields (see `FieldDefinition`).
 
-
 In `__main__.py`, you can see how a skill for the Helloworld agent is defined:
 
 ```python { .no-copy }
@@ -39,7 +38,6 @@ Key attributes of a `FieldDefinition` (from `a2a.types`):
 - `schema`: (optional) A JSON schema describing the expected structure for `"data"` fields.
 - `description`: (optional) Human-readable description of the field.
 - `optional`: (optional) Whether this field is optional (`True` or `False`).
-
 
 ## Agent Card
 
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -441,6 +441,23 @@ message AgentSkill {
   repeated FieldDefinition output_fields = 9;
 }
 
+// FieldDefinition describes an expected input/output field for a skill.
+message FieldDefinition {
+  // Optional name of the field.
+  string name = 1;
+  // The kind of content for this field: "text", "file", or "data".
+  string kind = 2;
+  // Optional list of supported mime types for this field.
+  repeated string mime_types = 3;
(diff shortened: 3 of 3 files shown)


holtskinner:
Makes sense to me, @kthota-g can you review?

kthota-g:
Currently examples is the mechanism to hint to the client about possible input types expected. An example is here - https://a2aproject.github.io/A2A/latest/specification/#56-sample-agent-card. If there are any missing attributes for the agent to complete the task, the agent can respond back with input-required to negotiate the required inputs. 

Also, different clients can request the output in varying formats and the clients can specify the schema for the output - https://a2aproject.github.io/A2A/latest/specification/#97-structured-data-exchange-requesting-and-providing-json

jankrynauw:
Could the `google.protobuf.Any` ([definition](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto)) type not play an elegant role here? 

```proto
message Any {
  // A URL/resource name that uniquely identifies the type of the serialized
  // protocol buffer message. This string must contain at least
  // one "/" character. The last segment of the URL's path must represent
  // the fully qualified name of the type (as in
  // `path/google.protobuf.Duration`). The name should be in a canonical form
  // (e.g., leading "." is not accepted).
  //
  // In practice, teams usually precompile into the binary all types that they
  // expect it to use in the context of Any. However, for URLs which use the
  // scheme `http`, `https`, or no scheme, one can optionally set up a type
  // server that maps type URLs to message definitions as follows:
  //
  // * If no scheme is provided, `https` is assumed.
  // * An HTTP GET on the URL must yield a [google.protobuf.Type][]
  //   value in binary format, or produce an error.
  // * Applications are allowed to cache lookup results based on the
  //   URL, or have them precompiled into a binary to avoid any
  //   lookup. Therefore, binary compatibility needs to be preserved
  //   on changes to types. (Use versioned type names to manage
  //   breaking changes.)
  //
  // Note: this functionality is not currently available in the official
  // protobuf release, and it is not used for type URLs beginning with
  // type.googleapis.com. As of May 2023, there are no widely used type server
  // implementations and no plans to implement one.
  //
  // Schemes other than `http`, `https` (or the empty scheme) might be
  // used with implementation specific semantics.
  //
  string type_url = 1;

  // Must be a valid serialized protocol buffer of the above specified type.
  bytes value = 2;
}
```

Inline with how Google uses common ([well-known](https://protobuf.dev/reference/protobuf/google.protobuf/) and [common](https://github.com/googleapis/api-common-protos)) protos to communicate well understood data structures (
`google.type.Date`, `google.type.PostalAddress`, `google.protobuf.Timestamp`, etc.),  is there not a world where agent specific definitions (via the AgentSkill) are made available at a well-known endpoint?

And then extending the Part definition to simplify the handling of these:

```proto
// Part represents a container for a section of communication content.
// Parts can be purely textual, some sort of file (image, video, etc) or
// a structured data blob (i.e. JSON).
message Part {
  oneof part {
    string text = 1;
    FilePart file = 2;
    DataPart data = 3;
  }
}

// FilePart represents the different ways files can be provided. If files are
// small, directly feeding the bytes is supported via file_with_bytes. If the
// file is large, the agent should read the content as appropriate directly
// from the file_with_uri source.
message FilePart {
  oneof file {
    string file_with_uri = 1;
    bytes file_with_bytes = 2;
  }
  string mime_type = 3;
}

// DataPart represents a structured blob. This is most commonly a JSON payload.
message DataPart {
  google.protobuf.Struct data = 1;
  google.protobuf.Any input = 2; 
  google.protobuf.Any output = 3;
}
```
Or even this:
```proto
// DataPart represents a structured object. This is most commonly a JSON payload.
message DataPart {
  google.protobuf.Any data = 1;
}
```

darrelmiller:
While I agree that there is some opportunity for improvement in how inputs and outputs of skills can be described, I think the current approach of providing media type identifiers and examples are a good compromise.  Trying to define a new schema language for input parts is a non-trivial task and one that no LLM is going to understand implicitly.  As crude as it appears, the use of examples is quite effective with LLMs.  

I sympathize with the desire here, but I don't think the proposed solution will make the situation better.   It does however, raise the question of whether there should be a metadata property in skills so that extensions could be defined to enable more strictly controlled interactions.

darrelmiller:
Currently consensus of participants of the TSC call is that this work is not a priority for the next release of A2A.  Adding this will blur the line between agent interactions and tool calling.  If people want to pass structured data they can do it via defined media types or create extensions to describe dynamically negotiated structures.

Closing for now.

PR #831, opened 2025-07-03
feat(spec): Add `tasks/list` method with filtering and pagination to the specification

### LINKED ISSUE #761: [Feat]: The ability to list tasks (tasks/list) in addition to the existing get (tasks/get)

No

An agent might have multiple tasks and as the client I would like to know what tasks are in progress, completed or failed.

Not an alternative, I think it will be beneficial just adding it to the official schema.

### Additional context

I'm working on a command line debugger for A2A agents where I would like to know what tasks failed or succeed, in order to do it, I need to be able to fetch all tasks and check their current status and history.

I think it's an essential functionality - for example, what if an A2A client wants to know if there is already tasks in progress on the A2A-server so it would not submit a new one.

### Code of Conduct

edenreich:
# Description

- [x] Fix existing linting errors (I guess they were already on the main upstream)

Fixes #761  🦕

BEGIN_COMMIT_OVERRIDE
feat(spec): Add `tasks/list` method with filtering and pagination to the specification

Release-As: 1.0.0
END_COMMIT_OVERRIDE

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,7 +2,7 @@
 site_name: A2A Protocol
 site_url: https://a2a-protocol.org/
 site_description: >-
-  The Agent2Agent protocol is an open standard that allows different AI agents to securely communicate, collaborate, and solve complex problems together.
+  The official documentation for the Agent2Agent (A2A) protocol. The A2A protocol is an open standard that allows different AI agents to securely communicate, collaborate, and solve complex problems together.
 site_author: The Linux Foundation
 site_dir: site
 


docs/specification.md
```diff
@@ -677,15 +677,55 @@ Retrieves the current state (including status, artifacts, and optionally history
...
+```ts { .no-copy }
+--8<-- "types/src/types.ts:ListTasksParams"
+```
+
+| Field Name      | Type                  | Required | Default | Description                                                                              |
+| :-------------- | :-------------------- | :------- | :------ | :--------------------------------------------------------------------------------------- |
+| `contextId`     | `string`              | No       |         | Filter tasks by context ID to get tasks from a specific conversation or session.         |
+| `status`        | [`TaskState`](#63-taskstate-enum) | No |         | Filter tasks by their current status state.                                              |
+| `limit`         | `integer`             | No       | `50`    | Maximum number of tasks to return. Must be between 1 and 1000.                          |
+| `offset`        | `integer`             | No       | `0`     | Number of tasks to skip for pagination. Must be 0 or greater.                           |
```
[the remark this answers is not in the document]
holtskinner: Is there a way to reconcile this? @edenreich

docs/specification.md
```diff
@@ -677,15 +677,55 @@ Retrieves the current state (including status, artifacts, and optionally history
...
+```
+
+| Field Name | Type                              | Required | Description                                           |
+| :--------- | :-------------------------------- | :------- | :---------------------------------------------------- |
+| `tasks`    | [`Task[]`](#61-task-object)       | Yes      | Array of tasks matching the specified criteria.       |
+| `total`    | `integer`                         | Yes      | Total number of tasks available (before pagination).  |
+| `limit`    | `integer`                         | Yes      | Maximum number of tasks returned in this response.    |
+| `offset`   | `integer`                         | Yes      | Number of tasks skipped for pagination.               |
+
+### 7.5. `tasks/cancel`
```
holtskinner, line 874: Be sure to adjust any other links using these numbers.

.gitattributes
holtskinner, line 1: You can remove this file. I added it in another PR

docs/specification.md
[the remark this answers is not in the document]
edenreich: Sure, will adjust it quickly.

edenreich:
Not sure about this linting error of 404.html file, I didn't even touched this file. I guess it will be fixed on the main upstream.

docs/specification.md
edenreich: These are now reconciled.

docs/specification.md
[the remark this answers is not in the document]
edenreich: These are now reconciled, using page_size and page_token. thanks!

holtskinner:
I did a rebase to pull in the updates from main so the lint errors don't mess with your PR.

@kthota-g and @pstephengoogle Can you please review/approve this change to the spec?

holtskinner (approved):

ReubenBond:
If I send a `tasks/list` command to a public-facing compliant A2A server without specifying a `contextId`, will I receive everyone's tasks? I believe the answer should be no, but the spec appears to not constrain the scope of the query when a `contextId` is not specified. We could add some text to say the implementation can decide the implicit scope when no `contextId` is specified. The implementation can implicitly scope a query to a user/tenant/etc. Alternatively, the scope could be an explicit parameter.

docs/specification.md
```diff
@@ -661,7 +661,33 @@ Retrieves the current state (including status, artifacts, and optionally history
 
 </div>
 
-### 7.4. `tasks/cancel`
+### 7.4. `tasks/list`
```
mikekistler: You should probably also remove lines 637-660.

edenreich: I notice there are some duplicates, cleaning it up!

edenreich: Thanks!

edenreich:
> If I send a `tasks/list` command to a public-facing compliant A2A server without specifying a `contextId`, will I receive everyone's tasks? I believe the answer should be no, but the spec appears to not constrain the scope of the query when a `contextId` is not specified. We could add some text to say the implementation can decide the implicit scope when no `contextId` is specified. The implementation can implicitly scope a query to a user/tenant/etc. Alternatively, the scope could be an explicit parameter.

@ReubenBond  Good point, I've added additional info to clarify it in the spec. Let me know if you'd like to modify it slightly - thanks!

edenreich:
I had to cleanup some changes that are not related to this Pull Request scope - it seems like some of the changes were removed from main upstream on the latest version - not sure exactly what went wrong with the Rebase but now I think the changes making sense.

darrelmiller:
@a2aproject/a2a-tsc I'm assuming that including this feature would require a minor version bump.  Features like this are a good reason for why we need to implement a proposal system because if we don't merge this promptly the author is going to run into a a lot of merge conflicts when we are ready to merge changes for the next version.

darrelmiller:
@edenreich 

I would expect the two most common filter criteria would be:
- all in-progress tasks
- recently completed tasks

The current set of filter criteria make those scenarios a bit tricky.  Having an array of status values would make it easier to identity all tasks in a terminal state, or in a non-terminal state.  Also, adding a lastUpdateDateTime as a filter parameter would make it easier to get just the tasks that users likely care about.

I also wonder whether it would be useful to provide an option to include a list of 'taskReference` objects as the result rather than a list of complete tasks.  Tasks that have large artifacts as outputs might make returning the list of tasks very expensive.

Also, by using `lastUpdatedDateTime` tasks it might be possible to significantly reduce the number of polling requests. That would be less viable if all the artifacts were returned on every poll.  

Are you proposing `tasks/list` be a required or optional method?

edenreich:
> @edenreich
> 
> I would expect the two most common filter criteria would be:
> 
> * all in-progress tasks
> * recently completed tasks
> 
> The current set of filter criteria make those scenarios a bit tricky. Having an array of status values would make it easier to identity all tasks in a terminal state, or in a non-terminal state. Also, adding a lastUpdateDateTime as a filter parameter would make it easier to get just the tasks that users likely care about.

Hi @darrelmiller, thanks for the question! According to the recent changes to the spec, there's pagination with a cursor, so the impact on the polling side is very minimal. Also, as far as I know, polling would only happen per client -> individual tasks, rather than across all tasks. For example, I wouldn't expect an A2A agent client to poll the tasks/list endpoint. Instead, for async operations, an A2A client would typically submit a task and then only poll the state of that task (or the tasks it submitted).

The tasks/list endpoint is more for observability systems that might need this type of information. From my perspective, it's useful for systems that want to answer the following questions:

“What tasks are currently in progress?”

“What tasks are currently in a working state?”

“How many are completed?”

“How many are currently queued?”

So yea I can see how filtering by status could be beneficial from useability perspective, good that you brought this up! and I could easily add that to the spec 👍

> I also wonder whether it would be useful to provide an option to include a list of 'taskReference` objects as the result rather than a list of complete tasks. Tasks that have large artifacts as outputs might make returning the list of tasks very expensive.

What if we just let the client filter by context id? would that work? so the way I understand the A2A and correct me if I'm wrong, a context might have multiple tasks and we already have the context id defined in the spec, maybe we should just allow filtering by this criteria and the status of course.

> Also, by using `lastUpdatedDateTime` tasks it might be possible to significantly reduce the number of polling requests. That would be less viable if all the artifacts were returned on every poll.

Fair point 👍

> Are you proposing `tasks/list` be a required or optional method?

I think it's more optional than a hard requirement for A2A. Just for monitoring and observability purposes.

So basically to sum it all up we can add the following filters:

- context id
- last updated date time
- status

What do you think?

swapydapy:
+1 to filters for lastUpdatedtime, list of task statuses.

Also should we consider a param to include artifacts or not, to make the Task result payload smaller.

edenreich:
> +1 to filters for lastUpdatedtime, list of task statuses.
> 
> Also should we consider a param to include artifacts or not, to make the Task result payload smaller.

@swapydapy Yea makes sense, this one we need too. I'll probably apply those adjustments by tomorrow to make it more flexible.

edenreich:
@darrelmiller @swapydapy I've added both filters to make it more flexible can you have a look?

docs/specification.md
```diff
@@ -807,7 +807,35 @@ Retrieves the current state (including status, artifacts, and optionally history
...
+
+Retrieves a list of tasks with optional filtering and pagination capabilities. This method allows clients to discover and manage multiple tasks across different contexts or with specific status criteria.
+
+**Pagination Strategy:** This method uses cursor-based pagination (via `pageToken`/`nextPageToken`) rather than offset-based pagination for better performance and consistency, especially with large datasets. Cursor-based pagination avoids the "deep pagination problem" where skipping large numbers of records becomes inefficient for databases. This approach is consistent with the gRPC specification, which also uses cursor-based pagination (`page_token`/`next_page_token`).
+
+**Security Note:** When `contextId` is not specified, implementations **MUST** ensure appropriate scope limitation based on the authenticated user's permissions. Servers **SHOULD NOT** return tasks from other users or unauthorized contexts. The implementation **MAY** choose to limit results to tasks created by the current authenticated user, tasks within a default user context, or return an authorization error if the scope cannot be safely determined.
```
yarolegovich: > When `contextId` is not specified, implementations **MUST** ensure appropriate scope limitation based on the authenticated user's permissions

The limitation should be applied regardless of whether `contextId` is specified or not 🙂

edenreich: @yarolegovich Fair point, so what do you suggest should we omit the contextId Part?

```suggestion
**Security Note:** Implementations **MUST** ensure appropriate scope limitation based on the authenticated user's permissions. Servers **SHOULD NOT** return tasks from other users or unauthorized contexts. The implementation **MAY** choose to limit results to tasks created by the current authenticated user, tasks within a default user context, or return an authorization error if the scope cannot be safely determined.
```

Like so?

edenreich: Anyways I think that's what you meant - I'll just commit it - makes sense.

docs/llms.txt
```diff
@@ -170,6 +170,10 @@ Key features of the A2A protocol highlighted by the specification and samples in
     - Retrieves the current state of a task.
     - `params`: `TaskQueryParams` (includes `id`, optionally `historyLength`).
     - `result`: `Task`.
+- **`tasks/list`:** (Request/Response)
+    - Retrieves a list of tasks with optional filtering and pagination.
+    - `params`: `ListTasksParams` (optional: `contextId`, `status`, `limit`, `offset`, `historyLength`).
+    - `result`: `ListTasksResult` (includes `tasks[]`, `total`, `limit`, `offset`).
```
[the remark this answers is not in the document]
swapydapy: This needs to be updated

docs/specification.md
```diff
@@ -1582,7 +1614,157 @@ _If the task were longer-running, the server might initially respond with `statu
...
+     "result": {
+       "tasks": [
+         {
+           "id": "3f36680c-7f37-4a5f-945e-d78981fafd36",
+           "contextId": "c295ea44-7543-4f78-b524-7a38915ad6e4",
+           "status": {
+             "state": "completed",
+             "timestamp": "2024-03-15T10:15:00Z"
+           },
+           "artifacts": [
```
swapydapy: Update this example, since artifacts are not included by default.

edenreich: Thanks, feel free to apply those changes - otherwise I'll probably look on this again tomorrow.

darrelmiller:
@edenreich The suggestion to exclude artifacts is an alternative way to address the size concerns I had.  

I am good with these changes.  My concern now is if we merge this, does that mean we bump the version to v0.4.0?

edenreich:
> @edenreich The suggestion to exclude artifacts is an alternative way to address the size concerns I had.

Great!

> I am good with these changes. My concern now is if we merge this, does that mean we bump the version to v0.4.0?

Yea, I think so, because it's a feature.

docs/specification.md
```diff
@@ -1582,7 +1614,157 @@ _If the task were longer-running, the server might initially respond with `statu
...
+                   "text": "I'd like to book a flight."
+                 }
+               ],
+               "messageId": "c53ba666-3f97-433c-a87b-6084276babe2",
+               "contextId": "c295ea44-7543-4f78-b524-7a38915ad6e4"
+             }
+           ],
+           "kind": "task"
+         }
+       ],
```
edenreich:
```suggestion
```

@swapydapy Is it fine like this? I'll apply this...

docs/specification.md
edenreich: changes applied! good catch! thanks!

edenreich:
> `lastUpdatedTime` filter:
> 
> 1. Let's rename  to `lastUpdatedAfter` to directly convey the intention of this filter.

@swapydapy  Hehe also Claude suggested this name, but I thought I would adjust it to lastUpdatedTime, sure I don't mind changing it to lastUpdatedAfter. I adjusted it back to `lastUpdatedAfter` like you suggested.

> 2. Also change the type to `number`. It should be milliseconds since the Unix Epoch. This keeps it fairly standard across platforms.

Good point. Adjusted it.

> `historyLength` filter: We should specify what is the default value? I suppose it should be 0.

Not sure how should we specify it - in text? description? I guess we can leave it to the users to decide?

holtskinner:
> > @edenreich The suggestion to exclude artifacts is an alternative way to address the size concerns I had.
> 
> Great!
> 
> > I am good with these changes. My concern now is if we merge this, does that mean we bump the version to v0.4.0?
> 
> Yea, I think so, because it's a feature.

Correct, this would need to be a bump to 0.4.0

docs/llms.txt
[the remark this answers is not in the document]
edenreich: Done ✅

pstephengoogle:
> > > @edenreich The suggestion to exclude artifacts is an alternative way to address the size concerns I had.
> > 
> > 
> > Great!
> > > I am good with these changes. My concern now is if we merge this, does that mean we bump the version to v0.4.0?
> > 
> > 
> > Yea, I think so, because it's a feature.
> 
> Correct, this would need to be a bump to 0.4.0

We have several open questions to address before we release 0.4.0
- I think we want all the official SDKs to fully support 0.3.x [how far away are we?]. I think this should happen before releasing 0.4.0 (probably the TSC call though)
- This change is easy to support old client <-> new server as the new endpoints just aren't known to the old client. The new client <-> old server, we need to ensure that the version in the agent card is checked before trying to make calls to the new endpoints. Should be simply, but calling it out

amye:
This is our first initial vote from the TSC! The TSC will vote on this change using the gitvote comment above.
However, any community member may also vote using the reactions on that comment as well, but only the members in the TSC team will be counted by the Gitvote bot.

darrelmiller:
@edenreich @pstephengoogle The current suggestion is to create a `draft-vnext` branch and rebase this PR to that branch. That will allow us to accumulate changes to 0.4.0 while the SDKs are catching up on supporting 0.3.x.

lerhaupt:
I upvoted via https://github.com/a2aproject/A2A/pull/831#issuecomment-3286560307 (presume its okay to approve in advance of the version being bumped up). Though I'm not sure how long it takes for the upvote to trigger in the tally, it hasn't yet included it. Nice to get bulk querying capabilities

muscariello (commented):
MUST-FIX (tasks/list): 
1) Add inline unit note for lastUpdatedAfter (ms epoch) in param bullets. 
2) Define includeArtifacts=false semantics: artifacts field MUST be omitted (not empty array). 
3) Add explicit InvalidParamsError cases table: pageSize out of bounds (1–100), malformed/expired pageToken, historyLength < 0, invalid status value, invalid lastUpdatedAfter (and optionally future timestamp). 
4) Make ordering normative: MUST return tasks sorted by descending last update time. 
5) Propagate pageSize (1–100, default 50) and historyLength (>=0, default 0) bounds/defaults across spec text, proto comments, JSON Schema, TS JSDoc. 
6) Decide & document final-page nextPageToken behavior (omit vs empty) and keep consistent. 
7) Clarify totalSize semantics (exact vs approximate vs optional) and align across layers.

muscariello (commented):
IMHO
SHOULD-FIX (tasks/list): 
1) Security note: listing without contextId MUST still scope to caller authorization/tenancy. 
2) Clarify status filter applies only to current state (not historical transitions). 
3) Define historyLength truncation ordering (newest-first?) and whether system events counted. 
4) Cross-transport mapping: explicitly list lastUpdatedAfter ↔ last_updated_time with units. 
5) Add REQUIRED/OPTIONAL method note in mapping table (or global statement). 
6) nextPageToken description: state omit vs empty consistently. 
7) totalSize: if exact say MUST; if approximate/optional document. 
8) Ensure every task example includes kind: "task".

muscariello (commented):
IMHO 
NICE-TO-HAVE: 
1) Add second example: includeArtifacts false + lastUpdatedAfter + follow-up pageToken request. 
2) Extensibility subsection (statuses[], orderBy, projection). 
3) Consider projection vs boolean includeArtifacts for future selective fields. 
4) Add CHANGELOG entry for spec v0.4.0 (beyond the scope of thus PR)
5) Add version annotation comments in TS.

edenreich:
> IMHO MUST-FIX (tasks/list):
> 
> 1. Add inline unit note for lastUpdatedAfter (ms epoch) in param bullets.
> 2. Define includeArtifacts=false semantics: artifacts field MUST be omitted (not empty array).
> 3. Add explicit InvalidParamsError cases table: pageSize out of bounds (1–100), malformed/expired pageToken, historyLength < 0, invalid status value, invalid lastUpdatedAfter (and optionally future timestamp).
> 4. Make ordering normative: MUST return tasks sorted by descending last update time.
> 5. Propagate pageSize (1–100, default 50) and historyLength (>=0, default 0) bounds/defaults across spec text, proto comments, JSON Schema, TS JSDoc.
> 6. Decide & document final-page nextPageToken behavior (omit vs empty) and keep consistent.
> 7. Clarify totalSize semantics (exact vs approximate vs optional) and align across layers.

@muscariello Thanks for reviewing, I have addressed those too. Good catch with the inconsistency between proto file and spec - those should be now in sync. I'll add a changelog - would highly recommend that this project uses semantic-release, it would make things much easier!

ToddSegal:
I'm in favor of the change but need to figure out how we avoid merging this as part of 0.3. This needs to be part of 0.4 and we need the links for 'latest' to point to 0.3 without this change.

amye:
> I'm in favor of the change but need to figure out how we avoid merging this as part of 0.3. This needs to be part of 0.4 and we need the links for 'latest' to point to 0.3 without this change.

So what'll happen is: the vote will close as 'pass', we mark it as 'approved for .4' and we merge it at that time. We can document this process as part of the release.

PR #841, opened 2025-07-07
refactor!: Change Well-Known URI for Agent Card hosting from `agent.json` to `agent-card.json`

holtskinner:
Based on feedback from IANA in https://github.com/protocol-registries/well-known-uris/issues/66

Release-As: 0.3.0


lukehinds:
What's the plan on rolling this out to the clients. I don't think it would be too disruptive as most are setting their own routes up?

holtskinner:
> What's the plan on rolling this out to the clients. I don't think it would be too disruptive as most are setting their own routes up?

This is going to be changed in the official SDKs in versions which support protocol version `0.3.0`

lindsaydsgdnolan-eng:
Please help me:
INFO:     127.0.0.1:40606 - "GET /.well-known/agent-card.json HTTP/1.1" 404 Not Found

lindsaydsgdnolan-eng:
> 请帮帮我：INFO： 127.0.0.1：40606 - “GET /.well-known/agent-card.json HTTP/1.1” 404 未找到

already fixed

PR #871, opened 2025-07-15
feat(proto)!: improve grpc ergonomics

### LINKED ISSUE #870: A2A rpc protos questions and suggestions
I've been reading through the [a2a.proto](https://github.com/a2aproject/A2A/blob/74ec26167cfb8e87d71dec9a2defabbbbfab55ac/specification/grpc/a2a.proto) and have some questions and suggestions.

Questions:

* What is the current support level for the protos? 
* Will we see a reference implementation? I've been unable to find any implementations, only code generated from the protos. 

### Suggestion: Avoid using multi-segment variables in `google.api.http` annotations.

While I appreciate the intention of being able to define both a gRPC _and_ REST APIs in one shot, I find the usage of multi-segment variable captures in the `google.api.http` (REST) annotations (ex: `/v1/{name=tasks/*/pushNotificationConfigs/*}`) to have produced some rather unattractive compound field values (ex: `name` with a value `tasks/my-task-id/pushNotificationConfigs/12345`). These will not only require that gRPC service implementations perform additional parsing that could have been handled by generated code, but it also requires that gRPC clients be aware of the REST api path structure when forming their requests. I am of the opinion that these annotations should be replaced with annotations that better leverage the capabilities of those implementing the `google.api.http` rule specifications.

#### Example of problematic proto

For example, the rpc for `GetTaskPushNotificationConfig`: https://github.com/a2aproject/A2A/blob/74ec26167cfb8e87d71dec9a2defabbbbfab55ac/specification/grpc/a2a.proto#L85-L91

led to a `GetTaskPushNotificationConfigRequest` message with a field `name` requiring a value of something like `tasks/my-task-id/pushNotificationConfigs/12345`. 
https://github.com/a2aproject/A2A/blob/74ec26167cfb8e87d71dec9a2defabbbbfab55ac/specification/grpc/a2a.proto#L594-L597

Code using a gRPC client would need to form the message:

```go
client.GetTaskPushNotificationConfig(ctx, &GetTaskPushNotificationConfigRequest{
  Name: fmt.Sprintf("tasks/%s/pushNotificationConfigs/%d", myTaskID, pushID),
})
```

Further, our server would be required to parse `tasks/my-task-id/pushNotificationConfigs/12345`. Who knows what kind of implications there might be for handling malformed requests.

#### Suggested

Do the following, instead:

```proto
  rpc GetTaskPushNotificationConfig(GetTaskPushNotificationConfigRequest)
    option (google.api.http) = {
      get: "/v1/tasks/{id}/pushNotificationConfigs/{push_id}"
    };
    option (google.api.method_signature) = "id,push_id";
  }

message GetTaskPushNotificationConfigRequest {
  string id = 1;
  int64 push_id = 2;
}
```

Client RPC code would look like:

```go
client.GetTaskPushNotificationConfig(ctx, &GetTaskPushNotificationConfigRequest{
  ID: myTaskID,
  PushID: pushID
})
```

And the service implementation wouldn't be need to do any special processing.

justfalter:
Improve ergonomics of gRPC client/server implementations by eliminating multi-segment variables from google.api.http path patterns, preferring instead to use simple variable captures. Modify associated rpc messages appropriately.

Relates to #870

BREAKING CHANGE: rpc field names have changed, new required fields have been added.

# Description

- [ ] Follow the [`CONTRIBUTING` Guide](https://github.com/a2aproject/A2A/blob/main/CONTRIBUTING.md).

Fixes #870  🦕

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -452,7 +452,7 @@ message AgentSkill {
 
 message TaskPushNotificationConfig {
   string task_id = 1;
-  string config_id = 2;
+  string config_id = 3;
   PushNotificationConfig push_notification_config = 2;
 }
 


specification/grpc/a2a.proto
```diff
@@ -581,43 +581,39 @@ message SendMessageRequest {
...
+  string config_id = 2 [(google.api.field_behavior) = REQUIRED];
 }
 
 message CreateTaskPushNotificationConfigRequest {
   // The task resource for this config.
-  // Format: tasks/{id}
-  string parent = 1 [(google.api.field_behavior) = REQUIRED];
+  string id = 1 [ (google.api.field_behavior) = REQUIRED];
+  // BUG?: config_id is not referenced in the google.api.http spec for CreateTaskPushNotificationConfig.
   string config_id = 2 [(google.api.field_behavior) = REQUIRED];
```
justfalter: Someone else should weigh in, but I'm pretty sure that this message should not have a `config_id`.

specification/grpc/a2a.proto
```diff
@@ -68,34 +68,34 @@ service A2AService {
...
+      get: "/v1/tasks/{id}:subscribe"
     };
   }
 
   // Set a push notification config for a task.
   rpc CreateTaskPushNotificationConfig(CreateTaskPushNotificationConfigRequest)
       returns (TaskPushNotificationConfig) {
     option (google.api.http) = {
-      post: "/v1/{parent=task/*/pushNotificationConfigs}"
+      post: "/v1/task/{id}/pushNotificationConfigs"
```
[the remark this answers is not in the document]
justfalter: I noticed this, too, but opted to carry it forward as I'm not familiar enough with its history.

specification/grpc/a2a.proto
```diff
@@ -581,43 +581,40 @@ message SendMessageRequest {
...
+  string config_id = 2 [(google.api.field_behavior) = REQUIRED];
 }
 
 message CreateTaskPushNotificationConfigRequest {
   // The task resource for this config.
-  // Format: tasks/{id}
-  string parent = 1 [(google.api.field_behavior) = REQUIRED];
+  string id = 1 [ (google.api.field_behavior) = REQUIRED];
+  // BUG?: config_id is not referenced in the google.api.http
+  // spec for CreateTaskPushNotificationConfig.
```
justfalter, line 606: Someone else should weigh in, but I'm pretty sure that this message should not have a config_id.

pstephengoogle:
The existing approach is designed around the google AIP standards. The resource name is what is captured, not the individual identifiers. This follows the standard resource based API. Individual identifiers should be extracted from the name/parent. This change shouldn't go through as it breaks this behavior

justfalter:
> The existing approach is designed around the google AIP standards. The resource name is what is captured, not the individual identifiers. This follows the standard resource based API. Individual identifiers should be extracted from the name/parent. This change shouldn't go through as it breaks this behavior

Could you please share the specific Google AIP that says you should do it in this way?

pstephengoogle:
https://google.aip.dev/122 under the sections "Fields representing resource names" and "Fields representing a resource's parent"

PR #901, opened 2025-07-22
feat!: Add mTLS to SecuritySchemes, add oauth2 metadata url field, allow Skills to specify Security

mikeas1:
# Description

This PR makes changes to our security representations in AgentCards:

1. Adds a new MutualTLSSecurityScheme type for indicating mTLS connection requirements. This matches the same field in the OpenAPI specification v3.1. This type is pretty lacking: there is no way to indicate any requirements on the certificate that is presented by the client. I expect we'll want to extend this in the future (for example, we may want to allow an agent to indicate that a SPIFFE certificate must be presented). For now, users are advised to put requirements in the description field.
2. Adds a field for specifying a URL for OAuth2 provider metadata. This is planned for OpenAPI Specification v3.2, which has not yet been released. This allows clients to fetch additional information about the OAuth provider, such as to discover if they support Dynamic Client Registration.
3. Allows AgentSkills to specify a `security` field. This allows agents to indicate security requirements for leveraging a particular skill.

I also clarified the usage and expectations for the `security` field in the base AgentCard.

This change is marked backwards incompatible due to the addition of the MutualTLSSecurityScheme, which adds a new type to an exhaustive enum (the security scheme `anyOf`).

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -371,6 +371,18 @@ message AgentCard {
   map<string, SecurityScheme> security_schemes = 8;
   // protolint:disable REPEATED_FIELD_NAMES_PLURALIZED
   // Security requirements for contacting the agent.
+  // This list can be seen as an OR of ANDs. Each object in the list describes
+  // one possible set of security requirements that must be present on a
+  // request. This allows specifying, for example, "callers must either use
+  // OAuth OR an API Key AND mTLS."
+  // Example:
+  // security {
+  //   schemes { key: "oauth" value { list: ["read"] } }
+  // }
+  // security {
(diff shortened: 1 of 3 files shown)


mikeas1:
> Can you update the docs with these changes? Or make another PR with the docs updates?

It looks like the specification.md doesn't detail the SecuritySchemes. Can we add that in a follow-up if we believe it's needed? We could just update the code boundaries to include all the specific schemes, too.

holtskinner:
> It looks like the specification.md doesn't detail the SecuritySchemes. Can we add that in a follow-up if we believe it's needed? We could just update the code boundaries to include all the specific schemes, too

I think either of those are fine

PR #917, opened 2025-07-25
feat: Add signatures to the AgentCard

### LINKED ISSUE #916: [Feat]: Allow AgentCards to be signed

There is no means to verify the integrity of an AgentCard. It is not possible for a party to attest to the content of an AgentCard.

Consumers of AgentCards may require the ability to know that an AgentCard is unchanged from when it was retrieved from a particular source, such as from a registry.

Consumers of AgentCards may require that AgentCards are vetted by a trusted source.

Allow AgentCards to contain signatures. This allows consumers to verify that the content of an AgentCard is unchanged since the signature was added. Signatures should be presented in a well-known format: I propose JSON Web Signatures (JWS). This allows verifiers to leverage pre-existing libraries for performing verification.

Signatures should cover the content of the AgentCard and also be contained by the AgentCard. That is, an AgentCard should get ` signatures` field that contains a list of signatures. This allows the AgentCard to be passed around as a single document that includes information sufficient to verify the integrity.

For a more complete description of the solution, see: https://github.com/a2aproject/A2A/discussions/199#discussioncomment-13770576

- Signatures via HTTP headers: there exists a standard for HTTP message signing. This could be used to return signatures for an AgentCard when served by an AgentCard endpoint. However, the signatures are not then contained within the AgentCard data and are tied to the transport mechanism (HTTP). We'd prefer a self-contained solution.

- Out-of-band signing: we could define a separate process for retrieving signatures of an AgentCard. This could be protocol-defined, where verifiers understand the specific means of retrieving known signatures for an AgentCard, such as from the rekor ledger offered by Sigstore. This allows great flexibility, but increases the burden on verifiers to understand external systems.

- AgentCard JWTs: rather than attaching signatures to an AgentCard, we could instead treat an AgentCard as a JWT. This format is similar to the proposal, however it inverts the relationship between AgentCard and signature: rather than an AgentCard containing a signature, the JWT contains the AgentCard. This eliminates the need to define a specialized process for producing signing input, however it bifurcates the serving format for signed and unsigned AgentCards. This means every consumer must concern themselves with whether an AgentCard is served as a JSON payload or as a JWT -- we either introduce a separate means of serving AgentCard JWTs, which increases the surface area of the protocol, or we require every implementer to handle both formats.

- Sign at registries: we could omit signatures from AgentCards, but expect that in production scenarios AgentCards are served via a registry that packages AgentCards with signatures. This effectively removes signing from the core A2A protocol and puts it off onto other registry/discovery protocols. We believe signing is a valuable feature, and so standardizing it via the core A2A protocol is appropriate.

### Additional context

See https://github.com/a2aproject/A2A/discussions/199.

### Code of Conduct

mikeas1:
# Description

This introduces a `signatures` field on the AgentCard that contains JWS formatted signatures. See issue #916 and discussion #199 for further details, including how signatures are constructed and verified.

Fixes #916 🦕


specification/grpc/a2a.proto
```diff
@@ -386,6 +386,8 @@
   // the user is authenticated, i.e. is the card from .well-known
   // different than the card from GetAgentCard.
   bool supports_authenticated_extended_card = 13;
+  // Digital signatures computed for this AgentCard.
+  repeated AgentCardSignature signatures = 14;
```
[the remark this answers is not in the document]
mikeas1: Good catch, Gemini!

.github/actions/spelling/expect.txt
holtskinner, line 1: Please add these to allow.txt instead of expect

mikeas1: Can do. FYI the job outputs a command to put these in expect.txt. I have no idea why this tool would have two files that do the same thing.

types/src/types.ts
```diff
@@ -223,6 +223,8 @@ export interface AgentCard {
    * to authenticated users. Defaults to false.
    */
   supportsAuthenticatedExtendedCard?: boolean;
+  /** JSON Web Signatures computed for this AgentCard.  */
```
```suggestion
  /** JSON Web Signatures computed for this AgentCard. */
```

types/src/types.ts
```diff
@@ -617,6 +619,24 @@ export interface TaskPushNotificationConfig {
...
+   * JSON object, as per RFC 7515.
+   */
+  protected: string;
+  /** The computed signature, Base64url-encoded. */
+  signature: string;
+  /** The unprotected JWS header values. */
+  header?: { [key: string]: any };
+}
+// --8<-- [end:AgentCardSignature]
+
```
holtskinner: This definition should probably be above the definition of `AgentCard` where it's used.

mikeas1: Done. Also moved SecuritySchemes since that was what led me astray. We don't strictly adhere to the define-before-use style of declarations (IMO, that's a good thing for readability: put the most important types first, then put the supporting types after -- we're not in C anymore, we can reference things defined later in the file), but it seems to at least hold for AgentCards.

holtskinner (approved):

PR #929, opened 2025-07-28
feat!: Add method for fetching extended card

kthota-g:
# Description
This change adds support for a new JSON-RPC method `agent/getAuthenticatedExtendedCard` for fetching authenticated extended card.

This is only applicable if `AgentCard.supportsAuthenticatedExtendedCard` bit is set to `true` that allows fetching authenticated context specific agent card.

In the previous versions, the extended card was served via `{AgentCard.url}/../agent/authenticatedExtendedCard` endpoint. This endpoint is being removed in favor of the JSON-RPC method. This endpoint will be deprecated in the SDKs and will be removed in a future release.

Release-As: 0.3.0


docs/specification.md
```diff
@@ -1014,16 +1003,19 @@ This section provides illustrative JSON examples of common A2A interactions. Tim
 
 3. **Client obtains necessary credentials out-of-band (e.g., performs OAuth 2.0 flow with Google, resulting in an access token).**
 
-4. **Client fetches the authenticated extended Agent Card:**
+4. **Client fetches the authenticated extended Agent Card using `agent/authenticatedExtendedCard` request:**
```
dmandar: nit: should this be getAuthenticatedExtendedCard like the other methods?

kthota-g: updated to agent/getAuthenticatedExtendedCard

types/src/types.ts
```diff
@@ -567,7 +567,7 @@ export interface GetTaskPushNotificationConfigParams extends TaskIdParams {
 /**
  * Defines parameters for listing all push notification configurations associated with a task.
  */
-export interface ListTaskPushNotificationConfigParams extends TaskIdParams {}
+export interface ListTaskPushNotificationConfigParams extends TaskIdParams { }
```
```suggestion
export interface ListTaskPushNotificationConfigParams extends TaskIdParams {}
```

holtskinner (approved):

PR #966, opened 2025-08-08
fix(spec): Suggest Unique Identifier fields to be UUID

### LINKED ISSUE #869: [Feat]:  Explicitly define allowed ID formats (e.g. taskId, configId, etc.)

Inconsistencies in ID parsing and validation (e.g. `taskId`, `configId`) lead to unexpected rejections or silent failures when using characters such as `-`, `.`, or `_`.

For example, `task-1` fails to match `\w+`, but this is not documented anywhere. This causes confusion for clients or API consumers trying to follow best practices for ID naming.

https://github.com/a2aproject/a2a-python/blob/dc95e2a03fe87815702d0a134195480608346ff3/src/a2a/utils/proto_utils.py#L17

I’d like to see the supported character set and format explicitly defined for identifiers like:

- `taskId`
- `configId`
- `artifactId`

This could be expressed as a regular expression, such as:

```
[a-zA-Z0-9_.-]+
```

N/A

### Additional context

N/A

### Code of Conduct

holtskinner:


specification/grpc/a2a.proto
```diff
@@ -73,7 +74,7 @@ service A2AService {
   rpc CreateTaskPushNotificationConfig(CreateTaskPushNotificationConfigRequest)
       returns (TaskPushNotificationConfig) {
     option (google.api.http) = {
-      post: "/v1/{parent=task/*/pushNotificationConfigs}"
+      post: "/v1/{parent=tasks/*}/pushNotificationConfigs"
```
izzyacademy: good catch!

kthota-g:
hi, this would be a breaking change and won't be backward compatible. The SDKs we build are using UUIDs, but custom implementations that are currently targeting to support 0.3.0 may not be using UUIDs.

types/src/types.ts
```diff
@@ -853,7 +879,9 @@ export interface PushNotificationConfig {
  * A container associating a push notification configuration with a specific task.
  */
 export interface TaskPushNotificationConfig {
-  /** The ID of the task. */
+  /**
+   * The unique identifier (UUID) of the task.
```
kthota-g: Add UUID as an example

kthota-g (approved):

types/src/types.ts
```suggestion
   * The unique identifier (e.g. UUID) of the task.
```

PR #977, opened 2025-08-12
docs - update specification to clarify web hook certificate verification

alextibbles:
# Webhook Certificate Checks

## User Benefit
While it is specified that Clients SHOULD verify Server identity through the certificate, it is not currently explicit that Servers also have this responsibility when delivering an update to a web hook URL. It would be beneficial to users if uniform requirements on identity verification were explicitly specified. Having a variation in requirements that depends on whether responses are delivered synchronously or asynchronously is surprising and seems unintentional.
The risk involved in not confirming the web hook certificate includes a potential breach of confidentiality of response payload.

## This Change
This PR proposes an additional statement making explicit the A2A Server responsibility when delivering asynchronous payload updates to authenticate the destination certificate.

## Compatibility
In theory, a previously fully conforming implementation that did not perform a check on a web hook server certificate would not meet the new SHOULD statement. I couldn't locate information about what this implies eg. with respect to versioning.
In practice, it seems at least somewhat unlikely that implementations does this because verification of server certificates is the default behavior across many different software toolsets for some time.


darrelmiller:
Considering the overall requirement to use HTTPS and the presence of section 4.2 saying that A2A clients should verify the A2A servers identity, then this addition does make sense.  However, I'm not convinced section 4.1 is appropriate as it feels dangerously close to telling people how to use HTTPS and [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-4.2.2) already does this authoritatively.

It might be worth defining a distinct "security considerations" section that gives leeway call out specific risks like unvalidated certs without trying to include it in normative text.

Anyway, this change makes the specification more consistent and so I think it should be approved.

darrelmiller (approved):

PR #1000, opened 2025-08-22
fix: Revert "chore(gRPC): Update a2a.proto to include metadata on GetTaskRequest"

### LINKED ISSUE #1002: [Bug]: GetTaskRequest.metadata gRPC change breaks http serialization
### What happened?

#996 introduced GetTaskRequest.metadata to gRPC which makes it in sync with the JSON RPC schema TaskQueryParams. However, GetTaskRequest is a GET in REST protocol, and given that metadata is a struct, it cannot be easily serialized or mapped to HTTP params.

This is causing multiple build failures in Google.

### Relevant log output

```shell
http: map field 'google.protobuf.Struct.fields' referred to by message 'a2a.v1.GetTaskRequest' cannot be mapped as an HTTP parameter.
```

### Code of Conduct

pwwpche:
Reverts a2aproject/A2A#996

Fixes #1002 

GetTaskRequest in the REST protocol is a GET. GET does not have request body, and all fields in GET should be serializable to a query param. 

metadata when being a struct cannot be serialized to query param, and this is causing failure during API discovery doc generation `http: map field 'google.protobuf.Struct.fields' referred to by message 'a2a.v1.GetTaskRequest' cannot be mapped as an HTTP parameter.`

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -637,8 +637,6 @@ message GetTaskRequest {
   string name = 1 [(google.api.field_behavior) = REQUIRED];
   // The number of most recent messages from the task's history to retrieve.
   int32 history_length = 2;
-  // Optional metadata for the request.
-  google.protobuf.Struct metadata = 3;
 }
 
 message CancelTaskRequest {


pstephengoogle:
I think we should remove it from the TaskQueryParams as well. It doesn't make sense there

ToddSegal (approved):
Approved per TSC discord discussion. This change should not have been merged into main to begin with. So reverting that mistake. That being said, something like this should be supported by defining official serialization mechanism for struct -> queryParam in the spec(s). Need parity between all transports and it is not there for this method.

darrelmiller (approved):

pstephengoogle:
I believe that the jsonrpc method should not have the metadata field either. For GetTask it is unclear what value this provides, and as of now, none of the implementations utilize this

PR #1004, opened 2025-08-25
fix(proto): Adds metadata field to A2A DataPart proto

### LINKED ISSUE #1005: [Bug]: A2A DataPart proto missing metadata field
### What happened?

The Python DataPart class [has metadata](https://github.com/a2aproject/a2a-python/blob/e3a7207164503f64900feaa4ef470d37fb2bb145/src/a2a/types.py#L285), but the proto doesn't. This leads to loss of important information during Python->proto conversion; e.g. ADK embeds part types like "function_call" and "function_response" [in the metadata](https://github.com/google/adk-python/blob/main/src/google/adk/a2a/converters/part_converter.py#L207-L211).

### Relevant log output

```shell

```

### Code of Conduct

hoonji:
The Python DataPart class [has metadata](https://github.com/a2aproject/a2a-python/blob/e3a7207164503f64900feaa4ef470d37fb2bb145/src/a2a/types.py#L285), but the proto doesn't. This leads to loss of important information during Python->proto conversion; e.g. ADK embeds part types like "function_call" and "function_response" [in the metadata](https://github.com/google/adk-python/blob/main/src/google/adk/a2a/converters/part_converter.py#L207-L211).

Fixes #1005 🦕


specification/grpc/a2a.proto
```diff
@@ -216,6 +216,7 @@ message FilePart {
 // DataPart represents a structured blob. This is most commonly a JSON payload.
 message DataPart {
   google.protobuf.Struct data = 1;
+  google.protobuf.Struct metadata = 2;
```
mikeas1: Per my comment in #1005, can you move this to the `Part` message instead?

gukoff: @mikeas1 curious, why aren't Python classes autogenerated from the protos?
Would it complicate their compatibility with the other transports such as JSON-RPC?

hoonji: Done, added the field to `Part`.

Note that the [Python DataPart](https://github.com/a2aproject/a2a-python/blob/9193208aabac2655a197732ff826e3c2d76f11b5/src/a2a/types.py#L285) differs from the [TS DataPart](https://github.com/a2aproject/A2A/blob/64f1064598e3b33cf06c40981e47aba5fb82f1e3/types/src/types.ts#L828-L835) in that the former defines the metadata field per class (instead of inheriting from an interface/superclass)

mikeas1: @gukoff, the a2a-python SDK contains classes generated from both the JSON schema specification, and the gRPC definition.

JSON Schema generated types: https://github.com/a2aproject/a2a-python/blob/main/src/a2a/types.py
gRPC generated types: https://github.com/a2aproject/a2a-python/tree/main/src/a2a/grpc

We had to choose one set of types for our transport-agnostic interfaces (i.e. AgentExecutor in a2a-python). And since the gRPC types didn't exist when we originally created those interfaces, it would have been hard to use those.

@hoonji, that is a consequence of the TS -> JSON Schema -> Pydantic chain of conversions. JSON Schema doesn't have inheritance. https://github.com/a2aproject/A2A/blob/64f1064598e3b33cf06c40981e47aba5fb82f1e3/specification/json/a2a.json#L694

mikeas1 (approved):

mikeas1:
@darrelmiller @ToddSegal for TSC approval.

tl;dr: This is a backwards compatible change to the proto definitions to bring it in line with the JSON-RPC definitions. It adds a field that's present in JSON-RPC but missing in proto (Part.metadata)

darrelmiller:
This seems to be a duplicate of #1019

PR #1050, opened 2025-09-09
fix(spec): Added normative content to specification.md that lived in topics 

darrelmiller:
# Description

This PR attempts to add content to the specification.md file that currently is only mentioned in the topics section.

The text in this PR is primarily LLM generated and this PR is draft because it is only here for the purpose of getting TSC alignment on moving forward with this approach.

- [X] Follow the [`CONTRIBUTING` Guide](https://github.com/a2aproject/A2A/blob/main/CONTRIBUTING.md).
- [X] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [X] Ensure the tests and linter pass (Run `nox -s format` from the repository root to format)
- [X] Appropriate docs were updated (if necessary)

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -300,6 +300,17 @@ Agent Cards themselves might contain information that is considered sensitive.
 - If an Agent Card contains sensitive information, the endpoint serving the card **MUST** be protected by appropriate access controls (e.g., mTLS, network restrictions, authentication required to fetch the card).
 - It is generally **NOT RECOMMENDED** to include plaintext secrets (like static API keys) directly in an Agent Card. Prefer authentication schemes where clients obtain dynamic credentials out-of-band.
 
+#### 5.4.1. Conditional Disclosure (Selective Cards)
+
+Agents **MAY** vary the content of the Agent Card based on the authenticated caller identity. When doing so:
+
+- Any reduced (public) view **MUST NOT** misrepresent supported authentication requirements or transports.
+- Omitted skills or extensions **MUST NOT** be implicitly required for successful invocation of the public interface.
+- Sensitive fields (e.g., internal-only skills, extension parameters) **MUST** only appear for authorized identities.
+- Clients **MUST NOT** assume that a minimal card implies absence of additional capabilities unless the card sets `supportsAuthenticatedExtendedCard: false`.
+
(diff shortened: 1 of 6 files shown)


docs/topics/streaming-and-async.md
```diff
@@ -63,6 +65,10 @@ Push notifications are ideal for:
 - Clients that cannot or prefer not to maintain persistent connections, such as mobile applications or serverless functions.
 - Scenarios where clients only need to be notified of significant state changes rather than continuous updates.
 
+Normative security requirements are centralized in [Spec §16 Push Notification Security](/specification/#16-push-notification-security-expanded). This section provides explanatory background.
+
+Security is paramount for push notifications due to their asynchronous and server-initiated outbound nature. Both the A2A Server (sending the notification) and the client's webhook receiver have responsibilities.
+
```
holtskinner:
```suggestion
```

samuelsharaf:
Looks good Darrel!

docs/specification.md
```diff
@@ -1922,3 +1933,168 @@ Implementations **SHOULD** validate compliance through:
...
+
+Multiple concurrent tasks under a single `contextId` are allowed; each task lifecycle is independent.
+
+## 14. Streaming Semantics & Reconnection
+
+Additional norms; background: [Streaming & Asynchronous Operations](./topics/streaming-and-async.md).
+
+### 14.1. Ordering
+
+Servers **MUST** emit: initial `Task` → zero/more updates → single `final` status update → close.
```
darrelmiller, line 2017: This isn't the case if the client sends a message using message/stream and the agent decides to respond with a message.

pstephengoogle: Yes, cause if the server sends a message than the stream should be considered terminated, so no resubscribe is possible (there also does not exist a task id to resubscribe to)

pstephengoogle: But yes, the spec should make that clear 

Servers **MUST** **EITHER**
- Emit a single message and close the stream
- OR emit an initial `Task` -> zero or more updates. If there are any updates the last one **MUST** be a `final` status update

darrelmiller:
This issue is replaced by PR #1160

PR #1079, opened 2025-09-17
feat: Unique identifier for an agent in Agent Card

### LINKED ISSUE #1014: [Feat]: Unique identifier for an agent in Agent Card

Currently Agent Card does not have a unique identifier attribute for an agent. agent_card.name is not guaranteed unique and is prone to branding changes.

- Add a new`id` attribute to the agent card which must be a UUID.
- This `id` must remain same for all versions of an agent.
- Mark the id as optional for backward compatibility.
- Clients can reliably use `id` as the identifier if set in the agent card.

_No response_

### Additional context

_No response_

### Code of Conduct

kthota-g:
- Adds a new id attribute to the agent card which must be a randomly generated UUID.
- This id must remain same for all versions of an agent.
- id is optional to support backward compatibility.
- Clients can reliably use id as the identifier if set in the agent card.

Fixes #1014

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -2,6 +2,7 @@
 syntax = "proto3";
 package a2a.v1;
 
+import "buf/validate/validate.proto";
 import "google/api/annotations.proto";
 import "google/api/client.proto";
 import "google/api/field_behavior.proto";
@@ -457,6 +458,8 @@ message AgentCard {
   repeated AgentCardSignature signatures = 17;
   // An optional URL to an icon for the agent.
   optional string icon_url = 18;
+  // A stable unique identifier (UUID) for an agent, optional.
+  optional string id = 19 [(buf.validate.field).string.uuid = true];
 }
 // --8<-- [end:AgentCard]
 


darrelmiller (approved):

muscariello:
is it an RFC 9562 type?

darrelmiller:
> is it an RFC 9562 type?

Effectively yes.  In the registry of formats used in OpenAPI/JSON Schema RFC4122 is referenced https://spec.openapis.org/registry/format/uuid.html and RFC9562 replaces that.  We should probably update the OpenAPI format registry.

ToddSegal:
This seems to have overlap with the concept of an agent identity. I worry there are problems if a "malicious agent" reuses the uuid of a "good agent". Things like DID's or SPIFFE's are useful for preventing impersonation and guaranteeing uniqueness. Should we be waiting for that instead of opening up to random unsecure ids?

muscariello:
I think @ToddSegal's point is correct. If we use globally unique names they should be secure. In that case an RFC 9562 type must be computed over a security envelope which provides content integrity. This could be achieved by taking the the whole content  as security envelope. However this is in conflict with the requirement that the id should not change if the version is changed. 
The DID instead does not guarantee content integrity unless there is an additional signature to provide integrity and provenance. In this second case too, changing version would imply changing id.

The actual definition of id should be spelled out including examples of how to compute it.

kthota-g:
The intent of the `id` is to provide a stable identifier for an agent that remains same across versions of the same agent. Without an identifier, client applications have to rely on name or url which can change to identify an agent.

`id` is just another attribute part of the agent card, similar to any other attributes. The contents of an agent card including the `id` is verifiable for the client applications with [signatures](https://a2a-protocol.org/latest/specification/#556-agentcardsignature-object).

darrelmiller:
@ToddSegal @muscariello  I worry about trying to attach any kind of security characteristics to this "id".  An agent is a web resource (at least it was until we started talking about stdio :-) ).  Consumers of the agent should be paying close attention to the AgentURL as the thing that ensures authenticity.  We have an entire ecosystem of security capability built around HTTPS.  I would really like to avoid reinventing that.  As @kthota-g mentioned we have signatures to ensure that agent cards have not been tampered with.

Maybe the solution is to pick a better name than "id" to provide a way to group agent versions together.

muscariello:
@darrelmiller I am ok with that. I was just asking if we can say how we compute it. It is based on hashing. If hashing is mis-computed you may get frequent collision. It might be just a reference or additional text.  And I have no issue to call it id.  My comment about secure naming is about integrity not provenance.
For instance, take all fields of the card and zero all fields which are supposed to vary like the version. Hash using RFC 9562, maybe add if you have a default behavior (which hash, or if any). Make sense?

kthota-g:
@muscariello `id` attribute is just an UUIDv4 value. It's only meant to guarantee uniqueness and stay constant between versions of an agent.

muscariello:
@kthota-g it was mentioned UUID to provide uniqueness. As by RFC 9562, I was thinking about UUID5 for naming things and not UUID4 which acts as a nonce. If it is a nonce and not a unique name it should be clarified. But from the description of the use case it looks like UUID5. UUID5 is good as it is reproducible. Above,  I was describing a namespace as of RFC 9562 to do that.

kthota-g:
Sorry for the confusion. My intent was to specify the `id` to be a unique value that remains constant for an agent. The value need not depend on other attributes from the agent card and it can be a random uuid.

I have update the description to reflect this.

VadymKalin (approved):

VadymKalin (approved):

holtskinner:
@kthota-g Please rebase this to the latest version of the spec

kthota-g:
> @kthota-g Please rebase this to the latest version of the spec

@holtskinner just rebased with main.

chorghemaruti64-creator:
The challenge with any unique identifier system — whether UUID or structured — is making it *persistent* across the agent lifecycle, not just unique at registration time.

An agent's identifier needs to survive:
- Model swaps (same agent, different underlying LLM)
- Host migrations  
- Capability updates

This is the addressing vs. identity distinction: the UUID gives you a unique name, but discovery layers need a *stable, resolvable address* that maps to wherever the agent actually is right now.

We've been working on this at the discovery layer — an agent registry where the identifier anchors behavioral history and capability records, not just routing metadata. The RFC 9562 UUID as a base makes sense; the open question is what the resolution protocol looks like when the agent's endpoint changes.

cc @darrelmiller — curious whether the current Agent Card design assumes the identifier is immutable post-registration, or whether key rotation / migration paths are in scope here.

PR #1086, opened 2025-09-19
chore: Update MAINTAINERS.md

amye:
Adding role:maintain and role:admin for various repos as they were previously undiscoverable.
Note: this is meant as a quick fix, we can re-evaluate on format in the future.


ToddSegal (approved):
approved - this doesn't change any permissions, just reflects the permissions in the repo. Going forward, will use this file to track proposals/approvals before making permission changes.

Issue #1072, opened 2025-09-15
[Bug]: history length is not optional in .proto


yarolegovich:
### Description

Both fields are optional in [types.ts](https://github.com/yarolegovich/A2A/blob/main/types/src/types.ts#L637) and [jsonrpc schema](https://github.com/a2aproject/A2A/blob/main/specification/json/a2a.json#L1683).
For both cases it might be useful to differentiate 0 and unset. For example:
1. A client calls GetTask and doesn't want History to be fetched / included in the response.
2. A client calls SendMessage and doesn't want history to be persisted at all.

The change is backward-compatible and makes it possible for a2a servers to provide consistent behavior regardless of the selected transport.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

darrelmiller:
@yarolegovich 

> A client calls SendMessage and doesn't want history to be persisted at all.

I think there is some confusion as to what the history length parameter means for SendMessage.

According to the comments in the proto file, the history_length parameter  for MessageSendConfiguration is:

> // The maximum number of messages to include in the history. if 0, the
>  // history will be unlimited.
>  int32 history_length = 3;

According to the comments in the types.ts file:

> /** The number of most recent messages from the task's history to retrieve in the response. */
>   historyLength?: number;

Unfortunately, the specification document doesn't include the `MessageSendConfiguration` object definition at all, so these comments are all we have to go on.

I am assuming that comments in the proto file are just a poorly worded version of what is in the types.ts file.  However, the comment about the 0 value leaves me wondering how a user is supposed to say "don't send me any history values".

I think what you are proposing is that an unset history length would return all of the history items and a 0 value would return no history items.  

@kthota-g  @pstephengoogle   @mikeas1   Do you folks have any clarity on what was intended here?

yarolegovich:
> Unfortunately, the specification document doesn't include the MessageSendConfiguration object definition at all, so these comments are all we have to go on.

Indeed, I didn't notice that. 

> I am assuming that comments in the proto file are just a poorly worded version of what is in the types.ts file. However, the comment about the 0 value leaves me wondering how a user is supposed to say "don't send me any history values".
I think what you are proposing is that an unset history length would return all of the history items and a 0 value would return no history items.

That's correct.

mikeas1:
[...]

PR #1092, opened 2025-09-22
fix(grpc): Rename push_notification to push_notification_config (#1072)

powmax-commits:
Fix for #1072
Related conversation: https://github.com/a2aproject/a2a-java/pull/290


darrelmiller:
It also appears that `SendMessageConfiguration` is not actually documented in the specification.md file.  Could we also add object definition to the specification.md file as part of this PR?

powmax-commits:
> It also appears that `SendMessageConfiguration` is not actually documented in the specification.md file. Could we also add object definition to the specification.md file as part of this PR?

I think it's in there: https://github.com/a2aproject/A2A/blob/main/docs/specification.md?plain=1#L692

Although if you look at the rendered spec there seems to be a problem with the tab rendering macro: https://a2a-protocol.org/latest/specification/#711-messagesendparams-object

I'll poke at it a bit tomorrow and see if I can fix that.

powmax-commits:
Added subheading for `MessageSendConfiguration`. Tabs render correctly in documentation.

darrelmiller:
Fixed in #1160

Issue #1077, opened 2025-09-17
[Proposal]: Versioning & Release Strategy


holtskinner:
#### 1.1. Versioning: Semantic Versioning (SemVer)

To clearly communicate the nature of changes in each release, the A2A Protocol will adhere strictly to [Semantic Versioning (SemVer)](https://semver.org/) using a `MAJOR.MINOR.PATCH` format.

*   **MAJOR Version (e.g., `2.0.0`):** Reserved for incompatible, breaking changes to the protocol. Any modification that requires existing agents or SDKs to be updated to maintain interoperability will trigger a MAJOR version increase.
*   **MINOR Version (e.g., `1.1.0`):** Used for adding new features or capabilities in a backward-compatible manner. This includes the introduction of new, optional fields or new endpoints. Any change in behavior that SDK developers need to be aware of, but that does not break existing implementations, must be a MINOR release.
*   **PATCH Version (e.g., `1.0.1`):** Limited to backward-compatible bug fixes, typo corrections, and clarifications in the specification. Changes at this level can be safely ignored by SDK developers as they do not alter the protocol's behavior or surface.

This structured approach ensures that the impact of any given release is immediately understandable to the community.

#### 1.2. Branching & Tagging Strategy

We recommend a hybrid strategy that leverages a primary development branch with automated tagging to ensure both stability and clear release demarcation.

*   **`main` Branch:** The `main` branch will serve as the primary development branch, always reflecting the "next" state of the protocol. All new features and patches will be merged into `main`.
*   **[Git Tags (`vX.Y.Z`)](https://docs.github.com/en/repositories/releasing-projects-on-github/viewing-your-repositorys-releases-and-tags):** A permanent, annotated Git tag will be created for every official release (e.g., `v1.2.0`). Tags are immutable pointers to a specific commit, serving as the definitive marker for a release version. The creation of these tags and their corresponding GitHub Releases will be automated.
*   **Release Branches (`release/vX.Y.Z`):** For major, complex releases that require a longer stabilization period, a dedicated `release` branch may be created from `main`. This allows for final testing and bug fixing without halting new feature development. However, for most minor and patch releases, this will not be necessary due to the automated process.

#### 1.3. Commit and Release Automation

To streamline the release process, ensure consistency, and reduce manual overhead, we will adopt the following automated practices.

[...]

PR #1096, opened 2025-09-23
docs: Update Specification Documentation Release to only publish when a release is made

holtskinner:
### Issues with this approach

- Style updates or link changes relating to other documentation pages won't be propagated to the specification doc when other pages are updated.

### Alternatives

- Maintain  copies of the specification docs for each minor version `0.X.0`
  - Similar to MCP Docs https://github.com/modelcontextprotocol/modelcontextprotocol/tree/main/docs/specification
  - This would also allow us to not use mike for versioning (would get rid of `latest` in URLs and could simplify SEO)
  
Follow-up to #1077

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -14,17 +14,18 @@
     ```
 {% endmacro %}
 
-??? note "**Latest Version** [`0.3.0`](../0.3.0/specification.md)"
+??? note "**Latest Version** [`0.3.0`](https://a2a-protocol.org/v0.3.0/specification)"
 
     **Previous Versions**
 
-    - [`0.2.5`](../0.2.5/specification.md)
-    - [`0.2.4`](../0.2.4/specification.md)
-    - [`0.2.3`](../0.2.3/specification.md)
-    - [`0.2.2`](../0.2.2/specification.md)
-    - [`0.2.1`](../0.2.1/specification.md)
-    - [`0.2.0`](../0.2.0/specification.md)
-    - [`0.1.0`](../0.1.0/specification.md)
+    - [`0.2.6`](https://a2a-protocol.org/v0.2.6/specification)
+    - [`0.2.5`](https://a2a-protocol.org/v0.2.5/specification)
+    - [`0.2.4`](https://a2a-protocol.org/v0.2.4/specification)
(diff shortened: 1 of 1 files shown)


darrelmiller:
@holtskinner 

> Style updates or link changes relating to other documentation pages won't be propagated to the specification doc when other pages are updated.

Agreed, this was my concern too.  I would be ok with us publishing dev as the latest as long as we make it clear to users how to navigate back to v0.3.0.  At the moment, I couldn't find any links back to the older versions of the docs/specification. I had to hack the URL to get the old versions.

> Maintain copies of the specification docs for each minor version 0.X.0

These older copies already exist in gh-pages

https://a2a-protocol.org/v0.3.0/
https://a2a-protocol.org/v0.2.6/
https://a2a-protocol.org/v0.2.5/

holtskinner:
> @holtskinner
> 
> > Style updates or link changes relating to other documentation pages won't be propagated to the specification doc when other pages are updated.
> 
> Agreed, this was my concern too. I would be ok with us publishing dev as the latest as long as we make it clear to users how to navigate back to v0.3.0. At the moment, I couldn't find any links back to the older versions of the docs/specification. I had to hack the URL to get the old versions.
> 
> > Maintain copies of the specification docs for each minor version 0.X.0
> 
> These older copies already exist in gh-pages
> 
> https://a2a-protocol.org/v0.3.0/ https://a2a-protocol.org/v0.2.6/ https://a2a-protocol.org/v0.2.5/

Correct, those other pages are still available at those links, which I could update the navigation links to point to. I was specifically meaning to maintain a dedicated markdown file on the main branch for each minor release, so we wouldn't need to use mike for versioning.

darrelmiller:
@holtskinner   So far I like what I see of the mike tool.  I've never used it before but it seems quite effective.  I'm ok with also emitting past versions into main, but that would be easier if the normative content was all in one file.  As it is split across specifications.md, types.ts and a2a.proto it would require having a folder for each version and it feels like it is going to get messy because those files are going to be at different relative paths to where they are in the latest version.

I would be perfectly happy with just links to the previously released documentation from the main branch.  But you seem to be on the hook for managing this, so do what works for you. As long as there is a clear way for visitors to find the old versions of the docs, I am happy with whatever solution we choose.

PR #1103, opened 2025-09-24
Update partners.md

chitra-venkatesh:
Updating to remove intake form - page maintenance


amye:
If people want to be added here, do they just submit a pull request to this page?

PR #1120, opened 2025-10-01
feat!: Proposal for bidirectional streaming over gRPC.

mikeas1:
This adds a new `SendLiveMessage` RPC to the gRPC A2A service definition. The purpose of this RPC is to address cases where a client would like to maintain an active request stream with an agent, rather than requiring separate follow-up requests any time an agent needs further input. This bidirectional communication is ONLY supported via gRPC.

I have tried to lay out the expected semantics as a comment in the gRPC specification.

A2A is generally a turn-based protocol, where clients and servers take turns sending each other messages. This is not specifically encoded in our specification, however our SDK implementations do not directly support the idea of receiving a stream of messages from a client. Instead, each new message is treated as a new invocation of the agent implementation. The assumption here is that a client will only send messages to tasks that are in an interrupted state (i.e. input-required), and that agent executions will exit when they reach an interrupted state. The current behavior of SDKs for when a message is received for a task that is actively being processed could be categorized as "undefined".

A bidirectional endpoint opens the possibility of full-duplex communication between client and server, however it doesn't require it: bidirectionality is still useful even in a turn-based protocol. The benefit is that an ongoing connection can be maintained for sending responses to agents that enter interrupted states, which enables several convenient properties:

- Agent implementations can more easily "await" responses inline, rather  than needing to save all necessary state, exit, then reconstitute state when a response is received.
- Clients can perform less state tracking, particularly around specifying task and context IDs. These can be implicit to the connection, as a connection is only valid for a single task.
- Distributed agent implementations don't need to implement clever load-balancing/routing to achieve task processing locality. Since an active connection is maintained, all responses are received by the same server. This is another view of the first point in this list, but from the networking layer.

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -54,8 +54,8 @@ service A2AService {
   //
   // The initial SendMessageRequest from the client MAY specify a
   // SendMessageConfiguration. Subsequent requests from the client must either
-  // send the same SendMessageConfiguration or omit the
-  // SendMessageConfiguration.
+// send the same SendMessageConfiguration or omit it. If omitted, the agent
+// MUST continue to use the configuration from the initial request.
   //
   // The initial Message provided by the client MAY include non-empty task_id
   // and context_id values. All subsequent Messages MUST either specify the
@@ -71,7 +71,7 @@ service A2AService {
   // `NOT_FOUND` error. Otherwise, the agent should send the current state of
   // the Task and begin sending updates. If a separate client has already
   // established a bidirectional connection with the agent for the specified
-  // task, the agent MAY reject the request with `UNIMPLEMENTED` if multiple
+// task, the agent MAY reject the request with `FAILED_PRECONDITION` if multiple
   // active request streams are not supported.
   //
   // If the agent does not support this method, it returns `UNIMPLEMENTED`.


ToddSegal:
So how does this interact with Resubscribe or other Send{xyz} methods? Also, what is the behavior if the client calls this multiple times? Is that disallowed? Allowed with responses multiplexed among both response streams? What happens if the client calls this without specifying task_id but the agent infers the request is relevant to an existing task which is already running against an instance of this?

mikeas1:
> So how does this interact with Resubscribe or other Send{xyz} methods

Resubscribe should continue to work as-is. It's just a read-only view on the stream of updates, so it would receive the same set of messages as the bidi connected receiver does.

Send\* methods don't have a well-defined specification for how they should react when multiple messages are received simultaneously. I'd suggest we designate a status code that an agent developer MAY use to reject requests if there's an active agent execution for a task and the agent doesn't support multiple input streams. This situation applies in both the bidi case and non-bidi case (i.e. there's an active bidi stream and another bidi stream or non-bidi request is received, or there's an active server-side stream and another request is received). My suggestion is `FAILED_PRECONDITION` for gRPC. `409 Conflict` would make sense for HTTP, but this proposal is gRPC only.

> Also, what is the behavior if the client calls this multiple times? Is that disallowed? Allowed with responses multiplexed among both response streams?

I call this out a little in the comments, but let me try to clarify.

I'd propose the following rules:

1. If the agent doesn't support multiple active client connections, error with `FAILED_PRECONDITION`.
2. If the agent establishes (meaning: the first response from the agent on the stream is a Task) the same task for multiple client connections, the response stream MUST be treated as though both were connected via `resubscribe`, i.e. all streams get all responses. All input streams are essentially multiplexed together with the agent seeing one unified stream.
3. As always, an agent may ignore the requested task and return a new task. In that case, a client attempting to connect to a task multiple times may end up with connections to different tasks.

> What happens if the client calls this without specifying task_id but the agent infers the request is relevant to an existing task which is already running against an instance of this?

Fall through to the above and treat the request as though it specified that task_id. It's probably good to explain this situation specifically, but I see this as a chain of two rules that cause this result naturally: the agent gets ultimate authority on what task a request is interpreted against + how do agents handle multiple bidi connections for a single task.

HungYangChang:
Hi @mikeas1 

Thanks for the initiative, our team are building agent upon A2A, and we are looking forward to following up on this PR

Question: Bidirectional streaming (SendLiveMessage), the stream stays open after interruption. This might create new possibilities but also ambiguities; it would be better if the spec should explicitly include

- Interrupted (Waiting for client input, can be resumed) vs Failed (Error occurred, cannot be resumed (terminal state))
- `State Transition Rules`: When a Task reaches 'interrupted' state on a bidirectional stream, what should Agent and Client do?
  - client might send:
   1. A new Message (with the same task_id) to provide additional info
   2. A `CancelTask` request equivalent (or just close the stream)
  - Upon receiving additional messages, the agent MUST:
   1. Transition task to 'working' state
   2. Continue processing
   3. Eventually reach terminal state (completed/failed/cancelled)
- Timeout rule: What if the client never responds? Auto-cancel after timeout
- Multiple interruptions (book flight: need budget -> need date confirmation -> need final approval)
- clear and proper error code definition

lumeche:
This change is really useful for us. I hope it gets merged soon

HungYangChang:
Hi @mikeas1  @ToddSegal 

This PR is small but has been stale for a while.
It will be great to review in the upcoming TSC Review. 
Our team is happy to resolve the comments and submit a new PR if needed

cc @holtskinner

hkmo99:
**We strongly support merging this proposal.**

This change fills a real gap in the current gRPC binding: handling follow-up input while a task is processing is currently undefined. Bidirectional streaming provides a clean transport-level solution without changing A2A semantics.

Importantly, this remains fully backward compatible:
- Adds a new RPC only
- Explicitly capability-gated
- No changes to existing messages or flows

It unlocks natural interactive A2A use cases (live input, multimodal, voice).

This has been blocking our implementation since October 2025. Merging it upstream would allow us to proceed without maintaining a fork, and while keeping the ecosystem aligned.

Tehsmash:
I like this proposal as a starting point, my primary concern however is whether this encourages multi-turn interaction to stick to a specific Task which isn't the recommended behaviour in the protocol. Ideally interrupted states should only be used for human-in-the-loop interactions and not multi-turn:

BAD -> `SendMessage -> Task -> Input-required -> SendMessage -> Input-required -> SendMessage -> Complete`

vs 

Recommended -> `SendMessage -> Task -> Complete -> SendMessage (reuse context) -> Task -> Complete`

I wonder if its possible to approach BiDi streaming from the second perspective with a BiDi stream actually sticks to a "context" rather than Task. Maybe this is a different set of operations.

hkmo99:
Just to clarify: `input-required` in A2A is not inherently human-in-the-loop. It simply means the Task requires additional input before progressing; that input can come from another agent, a tool, an orchestrator, or a human.

Bidirectional streaming does not change the Task model; it enables live interaction with an in-flight Task (e.g., missing parameters, confirmation, interrupts, client-end tools) in a transport-efficient way. Multi-turn conversations that represent new units of work should still be modeled as new Tasks with context reuse.

darrelmiller:
Superseded by #1549

PR #1127, opened 2025-10-06
fix: Add reference_task_ids to Message proto.

### LINKED ISSUE #1107: [Bug]: Protobuf and Json drift
I have noticed that the spec allows for **referenceTaskIds** and **extensions**. Both exist in the json spec definition. 

However:
1. referenceTaskIds are missing in the protobuf
2. extensions are not populated in the python sdk

Am I missing something?

mikeas1:
This field is present in the JSON Schema but not in the proto definition.

Fixes #1107


darrelmiller:
Addressed in #1138

PR #1132, opened 2025-10-10
docs(spec): Recommend use of X25519MLKEM768 due to Harvest Now Decrypt later atta…

loganaden:
…cks.

https://datatracker.ietf.org/doc/draft-ietf-tls-ecdhe-mlkem/ has been implemented in Chrome, and other browsers are following.

sign-off-by: Devraj Appadu <kushal@cyberstorm.mu>
sign-off-by: Poshan Peeroo <poshanpeeroo@cyberstorm.mu>
sign-off-by: Atish Joottun <atish.joottun@cyberstorm.mu>
sign-off-by: Keshwarsingh Nadan <kn@cyberstorm.mu>

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -250,6 +250,7 @@ For a comprehensive guide on enterprise security aspects, see [Enterprise-Ready 
 ### 4.1. Transport Security
 
 As stated in section 3.1, production deployments **MUST** use HTTPS. Implementations **SHOULD** use modern [TLS](https://datatracker.ietf.org/doc/html/rfc8446) configurations (TLS 1.3+ recommended) with strong cipher suites.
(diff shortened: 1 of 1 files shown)


darrelmiller:
Thank you for the heads up on this.  We should definitely consider this for future inclusion, however as the draft states

> It is inappropriate to use Internet-Drafts as reference material or to cite them other than as "work in progress."

We cannot make a SHOULD recommendation to use an IETF draft.  Chrome gets to take a dependency on it because they control the source code and can adapt to changes in the draft.  

I am going to close this PR for how, just to try and keep things more maintainable for the reviewers, but please bring this back when the draft is published as an RFC.

PR #1143, opened 2025-10-13
refactor: Make `a2a.proto` the unique source of truth

guglielmo-san:
Generate the a2a.json from the a2a.proto. The a2a.proto is now the only source of truth, as a2a.ts is automatically generated from it.


specification/json/a2a.json
```diff
@@ -1,240 +1,286 @@
...
                 },
-                "in": {
-                    "description": "The location of the API key.",
-                    "enum": [
-                        "cookie",
-                        "header",
-                        "query"
-                    ],
+                "location": {
+                    "description": "Location of the API key, valid values are \"query\", \"header\", or \"cookie\"",
```
ToddSegal, line 136: This should still be an enum

specification/json/a2a.json
```diff
@@ -1,240 +1,286 @@
...
+                "TaskSubscription"
             ],
-            "description": "A discriminated union of all standard JSON-RPC and A2A-specific error types."
+            "type": "object"
         },
-        "A2ARequest": {
-            "anyOf": [
-                {
-                    "$ref": "#/definitions/SendMessageRequest"
+        "A2AServiceHandlers": {
```
ToddSegal, line 70: is this needed?

muscariello:
This is problematic to me as I see no reason to keep all those ts files. I think this needs to be done differently. There is a refactor plan to come. So hold on with this one.

Issue #1150, opened 2025-10-16
Single Source of Truth for Types/Schemas


muscariello:
Goals:

- Ensure consistency between protobuf and JSON schema.
- Establish a canonical source for type definitions.

Contents:

- Protobuf definitions for all A2A types.
- Script/tooling to auto-generate JSON Schema from protobuf.
- CI checks to ensure schema sync.
- Documentation update to clarify that protobuf is the authoritative source.

Optional Enhancements:

- Include examples of how SDKs should consume the schema.
- Add a changelog or versioning strategy for schema evolution.

darrelmiller:
@muscariello There are currently differences between the way the JSON-RPC JSON payload the HTTP JSON payload.  If we are going to do what you propose it will be a breaking json for the JSON-RPC.  Not suggesting we shouldn't do this, but just for awareness.

e.g. the `data` payload in the SSE event response to POST /message/stream is

```typescript
{
  message?: Message,
  task?: Task,
  statusUpdate?: TaskStatusUpdateEvent,
  artifactUpdate?: TaskArtifactUpdateEvent
}
```

whereas in JSON-RPC the `data` payload is described as

```typescript
export interface SendStreamingMessageSuccessResponse
  extends JSONRPCSuccessResponse {
  /** The result, which can be a Message, Task, or a streaming update event. */
  result: Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent;
}
```

which over the wire looks something like

```
data: { "jsonrpc": "2.0",  "id": "<someId>",  "result": { "kind": "message", ... } }
```

Note that the result object has a "kind" property and is not the "facet" pattern as defined in gRPC and HTTP.

muscariello:
@darrelmiller agreed. We need to fix data vs transport separation first.

Tehsmash:
This was addressed in v1.0.0 and a2a.proto became the normative definition for the data model, I'll close this issue.

PR #1160, opened 2025-10-19
docs(spec): Large refactor of specification to separate application protocol definition from mapping to transports.

darrelmiller:
Related to #1151 

Fixes https://github.com/a2aproject/A2A/issues/998
Fixes https://github.com/a2aproject/A2A/issues/1091
Fixes https://github.com/a2aproject/A2A/issues/1072
Fixes https://github.com/a2aproject/A2A/issues/1058
Fixes https://github.com/a2aproject/A2A/issues/975
Fixes https://github.com/a2aproject/A2A/issues/1105
Fixes https://github.com/a2aproject/A2A/issues/989
Fixes https://github.com/a2aproject/A2A/issues/1189
Fixes https://github.com/a2aproject/A2A/issues/1187
Fixes https://github.com/a2aproject/A2A/issues/1177
Fixes https://github.com/a2aproject/A2A/issues/1180
Fixes https://github.com/a2aproject/A2A/issues/1172

In order to more effectively describe the A2A protocol in a consistent way across different protocols, this PR refactors the specification to describe the application protocol in a transport agnostic way, and then have separate sections that describe how each of the transports map to the abstract application protocol.  To see a preview of what it looks like you can go here https://purple-moss-0838db51e.3.azurestaticapps.net/specification/

The goal of the TSC is to use the changes in this PR as the baseline for a v1.0 Release Candidate. This will give SDK maintainers an stable document to validate the changes in the specification through implementations before v1.0 release.

There are many pending feature requests for the A2A specifications that the TSC are supportive of.  However, getting a stable, consistent, unambiguous, multi-protocol specification is necessary before additional features can be added safely.

## Document Restructuring 

- [A2A Operations](https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#3-a2a-protocol-operations) are described independent of wire protocol 
- [Protocol Data Model](https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#4-protocol-data-model) is described using protobuf as source of truth. All other serializations are derived from that. 
- Protocol bindings (formerly described as transports) are now described in their own sections
  - [JSON-RPC](https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#9-json-rpc-protocol-binding)
  - [gRPC](https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#10-grpc-protocol-binding)
  - [HTTP](https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#11-httpjsonrest-protocol-binding)
[...]

PR #1168, opened 2025-10-21
fix: make `a2a.proto` the unique source of truth for Schema

guglielmo-san:
The goal of this PR is to propose a solution to this open task (https://github.com/a2aproject/A2A/issues/1150)
The current issue in the A2A project is the presence of `src/types/types.ts` as second source of truth other than the `a2a.proto`.
To fix this, the original `a2a.proto` has been splitted into two files:
1. `a2a_grpc.proto`: contains the gRPC Service definitions
2. `a2a_types.proto`: contains the message data defintion, common to gRPC ans JSONRpc

The `a2a_types.proto` is used to generate the `a2a_types.ts`, using the `generate-types` script command. `a2a_types.ts` will be generated in the folder `data\types`, which already hosts the predefined `types_jsonrpc.ts`. This file contains the message structure used by the JSONRpc, and will reference the types defined in the `a2a_types.ts`, generated from `a2a_types.proto`.
This structure will guarantee that the messages used by JSONRpc and gRPC transports are coming from a single source of truth.

The `a2a_types.ts` and `types_jsonrpc.ts` are then used to generate the `a2a.json` schema with the `generate` script command, used within the A2A language specific repositories.

The new folder structure will be:

```
src/
  data/
    a2a_types.proto -- Contains A2A message definitions.
    types/
       a2a_types.ts -- Contains Typescript definitions, generated from a2a_types.proto
       types_jsonrpc.ts -- Contains Typescript definitions for JSONRpc Request/Response/Error, referencing `a2a_types.ts`
  transports/
    json-rpc/
      a2a.json -- Contains JSON Schema representations of JSON-RPC objects.
    grpc/
      a2a_grpc.proto -- Contains A2A service definition, referencing `a2a_types.proto` values.

```


lkawka:
A few observations and suggestions:
1. Regarding the file structure, I recommend consolidating the two protobuf files into a single directory. Given that messages are integral to the gRPC specification, a unified location would improve clarity and ease comprehension.
1. It appears that documentation is generated from these definitions, and the new `a2a_core.ts` file currently lacks any documentation annotations.
1. The buf configurations seem to have been completely removed. Was this an intentional change? These configurations are likely beneficial to users experimenting with the repository.

darrelmiller:
This is addressed in #1160   a2a.proto is now the source of truth for data definitions

PR #1173, opened 2025-10-21
fix(proto): align APIKeySecurityScheme JSON field with OpenAPI 3.0 ("in" vs "location")

### LINKED ISSUE #1172: [Bug]: APIKeySecurityScheme field name inconsistency between JSON schema and gRPC proto
### Problem Description

There is a critical field name mismatch in the `APIKeySecurityScheme` message that breaks interoperability between JSON-RPC and gRPC transports:

- JSON Schema ([a2a.json](https://github.com/a2aproject/A2A/blob/main/specification/json/a2a.json)): Uses field name "in" (following OpenAPI 3.0 standard)
- gRPC Proto ([a2a.proto](https://github.com/a2aproject/A2A/blob/main/specification/grpc/a2a.proto)): Uses field name "location"

This inconsistency causes authentication configuration to fail when converting between JSON and gRPC formats, as clients implementing different transports cannot properly exchange security scheme information.

#### Current State

**gRPC Proto:**
```protobuf
message APIKeySecurityScheme {
  string description = 1;
  string location = 2;  // ❌ Incorrect field name
  string name = 3;
}
```

**JSON Schema:**
```json
"APIKeySecurityScheme": {
  // ...
  "in": {
                    "description": "The location of the API key.",
                    "enum": [
                        "cookie",
                        "header",
                        "query"
                    ],
                    "type": "string"
                }
// ...
}
```

#### Why This Breaks

1. **Proto serializes to JSON as `"location"`**, violating the JSON schema
2. **JSON clients send `"in"`**, which proto parsers don't recognize
3. **OpenAPI 3.0 standard mandates `"in"`** - the JSON schema is correct per spec
4. **Authentication configuration fails silently** - security schemes are ignored

#### Expected Behavior

Both specifications should use `"in"` to match the OpenAPI 3.0 Security Scheme specification.

#### Reproduction
```protobuf
// 1. Create security scheme in gRPC
security_schemes {
  key: "api_key"
  value {
    api_key_security_scheme {
      description: "API Key authentication"
      location: "header"  // Using proto field name
      name: "X-API-Key"
    }
  }
}

// 2. Serialize to JSON via HTTP/JSON transcoding
// Result: {"location": "header", ...}

// 3. Validate against JSON schema
// Error: Missing required field "in"
```

#### Proposed Solution

**Option A (Recommended): Rename proto field**
```protobuf
message APIKeySecurityScheme {
  string description = 1;
  string in = 2;  // CHANGED: Renamed from "location" to "in"
  string name = 3;
}
```

**Option B: Add json_name annotation**
```protobuf
message APIKeySecurityScheme {
  string description = 1;
  string location = 2 [json_name = "in"];  // Map to correct JSON field
  string name = 3;
}
```

**Recommendation:** Option A aligns the proto with OpenAPI 3.0 standard and maintains consistency. Field number 2 remains unchanged, so wire format compatibility is preserved.

### Relevant log output

```shell

```

### Code of Conduct

adarshnarayanhegde:
### Summary

This pull request fixes the critical field name mismatch in `APIKeySecurityScheme` that breaks interoperability between JSON-RPC and gRPC transports.  

The gRPC proto previously used `location` for field number 2, while the JSON/OpenAPI 3.0 schema uses `in`. This caused authentication configuration to fail during JSON/gRPC transcoding.

### Changes Made

- Renamed `APIKeySecurityScheme.location` → `APIKeySecurityScheme.in` in `specification/grpc/a2a.proto`.
- Field number `2` remains unchanged to preserve wire format compatibility.

Fixes #1172
Closes #1172

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,9 +13,9 @@ hide:
 
 ## What is A2A Protocol?
 
-The **Agent2Agent (A2A) Protocol** is an open standard developed by Google and donated to the Linux Foundation designed to enable seamless communication and collaboration between AI agents.
+Welcome to the **official documentation** for the **Agent2Agent (A2A) Protocol**, an open standard designed to enable seamless communication and collaboration between AI agents.
 
-In a world where agents are built using diverse frameworks and by different vendors, A2A provides a common language, breaking down silos and fostering interoperability.
+Originally developed by Google and now donated to the Linux Foundation, A2A provides the definitive common language for agent interoperability in a world where agents are built using diverse frameworks and by different vendors.
 
 !!! abstract ""
     Build with
@@ -25,7 +25,7 @@ In a world where agents are built using diverse frameworks and by different vend
     **![A2A Logo](./assets/a2a-logo-black.svg){class="twemoji sm middle"} A2A**,
     to remote agents, local agents, and humans.
 
-## Get started with Agent2Agent (A2A) Protocol
+## Get started with A2A
 
 <div class="grid cards" markdown>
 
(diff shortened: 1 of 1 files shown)


holtskinner:
@darrelmiller @muscariello Should this change be added into #1160
I believe this field name is currently set as is in the JSON-RPC version

darrelmiller:
This has been reconciled in #1160  We use `location` now because it is a better name than `in`.

PR #1175, opened 2025-10-23
feat(spec): add schema generation pipeline & docs integration (stacked)

muscariello:
Stacked tooling & documentation improvements atop transports without rebase.
Related to #1160 

Highlights:
- Introduces proto -> OpenAPI -> JSON Schema pipeline (protoc + grpc-gateway + jq extraction) integrated into build_docs.
- Pretty-printed multi-line JSON schema published to docs/spec-json/a2a.json (ephemeral, non-normative).
- Removes committed schema artifact; clarifies non-normative status in specification/json/README.md.
- Updates mkdocs.yml with redirect and includes generated spec-json path.
- Adds scripts: generate_openapi.sh, extract_json_schema.sh, build_docs.sh.
- CI workflow adjusted to regenerate swagger only on proto changes.
- Preserves legacy anchors and normative source principle already in transports branch.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -128,8 +128,8 @@ agentic
 agentskill
 agntcy
 agp
-aip
 ainvoke
+aip
 airbnb
 aldridge
 alloydb
@@ -225,7 +225,6 @@ konami
 kty
 langgraph
 linenums
-listtasks
 linkedin
 linting
 listtasks


.github/workflows/generate-a2a-json.yml
```diff
@@ -1,69 +1,83 @@
...
+            git commit -m "chore: update generated OpenAPI swagger"
+            git push
           else
-            echo "a2a.json is up-to-date."
-          fi
+            echo "No swagger changes detected."
 
-      - name: Commit and Push Changes (if on base repository)
-        if: steps.git_status.outputs.changes_detected == 'true' && github.event.pull_request.head.repo.full_name == github.repository
+      - name: Commit swagger changes on main
```
darrelmiller, line 75:
```suggestion
      - name: Commit OpenAPI changes on main
```

docs/specification.md
```diff
@@ -403,8 +407,29 @@ This specification defines three standard transport protocols: [JSON-RPC Transpo
 
 The A2A protocol defines a canonical data model using Protocol Buffers. All transport implementations **MUST** provide functionally equivalent representations of these data structures.
 
+"Normative Source" Principle:
+
+The file `specification/grpc/a2a.proto` is the single authoritative normative definition of all protocol data objects and request/response messages. A generated JSON artifact (`specification/json/a2a.json`, produced at build time and not committed) MAY be published for convenience to tooling and the website, but it is a non-normative build artifact. SDK language bindings, schemas, and any other derived forms **MUST** be regenerated from the proto (directly or via code generation) rather than edited manually.
+
+Change Control and Deprecation Lifecycle:
+
+- Introduction: When a proto message or field is renamed, the new name is added while existing published names remain available for at least one subsequent minor release.
```
darrelmiller: Should this not say until the next major release?  I had assumed we would attempt to confirm to semantic versioning.

mkdocs.yml
```diff
@@ -8,7 +8,8 @@ site_dir: site
 edit_uri: edit/main/docs/
 
 extra:
-  a2a_version: !ENV [MIKE_VERSION, 'dev']
+  a2a_version: !ENV [MIKE_VERSION, "dev"]
```
darrelmiller: I'm not sure how comfortable I am editing this file.  I'd prefer to keep TSC updates related to the specification.

mkdocs.yml
```diff
@@ -110,9 +111,7 @@ markdown_extensions:
   - attr_list
   - md_in_html
   - pymdownx.details
-  - pymdownx.emoji:
-      emoji_index: !!python/name:material.extensions.emoji.twemoji
-      emoji_generator: !!python/name:material.extensions.emoji.to_svg
```
darrelmiller, line 115: Why are these two lines removed?

mkdocs.yml
```diff
@@ -125,12 +124,8 @@ markdown_extensions:
       custom_fences:
         - name: mermaid
           class: mermaid
-          format: !!python/name:pymdownx.superfences.fence_code_format
   - pymdownx.tabbed:
       alternate_style: true
-      slugify: !!python/object/apply:pymdownx.slugs.slugify
-        kwds:
-          case: lower
```
darrelmiller, line 133: Why is this removed?

darrelmiller:
@muscariello These changes look good to merge into the transports branch, other than some of the changes to the mkdocs.yml file that I'm not clear why we are removing things.

I have some questions about our versioning plans but we can discuss that before we merge the transports branch into main.

Do you want me to merge into transports and we can discuss the mkdocs changes there?

muscariello:
@darrelmiller I can also address your comments here before merging to transport.

docs/specification.md
Tehsmash: If we are conforming to semantic versioning this should be next major indicating a breaking change. This is well understood by many package managers etc, but often does result in batching up large numbers of breaking changes, making moving from v1.x.x to v2.x.x a larger change than expected.

Alternatively, we define our own policy as this section seems to be doing, similar to k8s https://kubernetes.io/releases/version-skew-policy/, where we support backwards compatibility within a range of 2 minor versions. The downside here is that it pushes the requirement to the clients to track the features/deprecations they are using and can't upgrade "blindly" so easily.

muscariello: I have updated to next major. TBD in the future I'd say.

PR #1190, opened 2025-10-31
feat(spec): Add client-side protocol versioning in payloads

### LINKED ISSUE #1189: [Feat]: Add client-side protocol versioning in payloads

With the upcoming 1.0 spec release, we are expecting some breaking changes which will have an impact on SDKs and their users.

To ensure backwards compatibility, we can implement conversion layers within the SDKs. This approach requires the client to communicate its protocol version. The server would then be able to translate requests from older versions (e.g., 0.3) to the current version (e.g., 1.0), process them, and convert the responses back to the client's version, thus providing backwards compatibility.

The proposal is to add `protocol_version` fields on relevant requests, so that the server may know which version the client is on and act accordingly. 

Specifically for the 0.3 to 1.0 transition, absence of the value will indicate that the server is on the 0.3 version.

### No spec changes

SDKs could try to infer protocol version from request content but that would be imprecise and ineffective for response-only changes.

### Additional context

_No response_

### Code of Conduct

lkawka:
Added `protocol_version` fields to request objects.

Fixes #1189

### DIFF
--- a/types/src/types.ts
+++ b/types/src/types.ts
@@ -613,7 +613,7 @@ export interface GetTaskPushNotificationConfigParams extends TaskIdParams {
 /**
  * Defines parameters for listing all push notification configurations associated with a task.
  */
-export interface ListTaskPushNotificationConfigParams extends TaskIdParams { }
+export interface ListTaskPushNotificationConfigParams extends TaskIdParams {}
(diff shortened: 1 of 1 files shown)


darrelmiller:
Addressed by #1160  Please reopen the issue if you believe there are unaddressed issues.

PR #1206, opened 2025-11-17
feat(spec): Add last update time to `Task`

### LINKED ISSUE #1207: [Feat]: Add update_time to Task

The new task/list method's [specification](https://github.com/a2aproject/A2A/blob/main/docs/specification.md#74-taskslist) dictates that tasks must be ordered by the last update time. However, the current data model lacks a direct way to determine when a task was last updated.

To resolve this, an `update_time` field should be added to the `Task` message.

1. **The [`timestamp`](https://github.com/a2aproject/A2A/blob/0534ba6a9ea4e03b6c5fd538f8db17834d13e987/types/src/types.ts#L512) field in the `TaskStatus` message already exists**: There two issues with this field:
    1. Not all task updates are necessarily status updates.
    1. The field is optional, which can lead to unstable ordering when sorting by this value.
1. **Backends can just incorporate this information into their internal models**: While true, including this in the public API offers significant advantages by allowing clients to readily access and display this value to users.

### Additional context

_No response_

### Code of Conduct

lkawka:
The new `task/list` method's [specification](https://github.com/a2aproject/A2A/blob/main/docs/specification.md#74-taskslist) dictates that tasks must be ordered by the last update time. However, the current data model lacks a direct way to determine when a task was last updated. To resolve this, a new field with last update time is being added to the `Task` message.

Addressing potential concerns:
1. **The [`timestamp`](https://github.com/a2aproject/A2A/blob/0534ba6a9ea4e03b6c5fd538f8db17834d13e987/types/src/types.ts#L512) field in the `TaskStatus` message already exists**: There are two issues with this field:
    1. Not all task updates are necessarily status updates.
    1. The field is optional, which can lead to unstable ordering when sorting by this field alone.
1. **Backends can just incorporate this information into their internal data models**: While true, including this in the public API offers significant advantages by allowing clients to readily access and display this value to users.

Fixes #1207

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -233,6 +233,7 @@ jku
 jqlang
 jti
 jwks
+kadirpekel
 konami
 kty
 langgraph
--- a/docs/community.md
+++ b/docs/community.md
@@ -79,6 +79,7 @@ These agentic frameworks have built-in A2A integration, making it easy to get st
 - [Agent Development Kit (ADK)](https://google.github.io/adk-docs/a2a/)
 - [AG2](https://docs.ag2.ai/latest/docs/user-guide/a2a/)
 - [BeeAI Framework](https://framework.beeai.dev/integrations/a2a)
+- [Hector](https://github.com/kadirpekel/hector)
 - [LangGraph](https://docs.langchain.com/langgraph-platform/server-a2a)
 - [Pydantic AI](https://ai.pydantic.dev/a2a/)
 


amye:
Sigh, it really does need duration. Reverting!

He-Pin:
I think is is great, mcp just added this too.

geneknit:
Discussion moving to Discord and the main issue. Concern about having two that serve the same value. cc: @darrelmiller

Tehsmash:
After discussion in the TSC meeting on 2026/01/6, we agreed to leave the fields in Task as they are and change the filter for ListTasks to `status_timestamp_after` to indicate that its based on the task status only.

darrelmiller:
@Tehsmash I think that means we could close this PR and combine the proposed naming change with this PR to change the datatype https://github.com/a2aproject/A2A/pull/1288

Tehsmash:
> @Tehsmash I think that means we could close this PR and combine the proposed naming change with this PR to change the datatype #1288

👍  happy for it to be combined (whatever makes it turn around faster)

darrelmiller:
Closing this PR in favour of #1358

amye:
/cancel-vote 

#1358 has the correct one + gitvote has added Stephen already, so yay! We'll restart this.

PR #1211, opened 2025-11-20
fix(spec): draft-v1 specification update

darrelmiller:
This change contains a subset of the changes in PR #1160 .  These changes have been split out to enable merging of the document updates whilst we complete work on the scripting updates that will generate the a2a.json from a2a.proto

All changes in this PR were approved by the TSC in this vote https://github.com/a2aproject/A2A/pull/1160#issuecomment-3528520409

Note, due to the size of this PR change and nature of how it came about, we do not consider the TSC vote as final say on the changes in this document.  Further iterations will be done before a v1 release.


darrelmiller:
Change of plan. Just going to merge #1160 instead.

PR #1214, opened 2025-11-20
docs: Added adr template and protojson adr

darrelmiller:
Added ADR that captures decision to use ProtoJSON as the normative reference for JSON serialization


amye:
This vote actually passed, editing labels to be clear.

adrs/adr-001-protojson-serialization.md
ishymko, line 1: Should we also update the spec which currently documents a different format? https://github.com/a2aproject/A2A/blob/60f83c3faac4770b231f038406c9e02282887a25/docs/specification.md?plain=1#L1194-L1201

darrelmiller: I'll create a separate issue for that.

PR #1225, opened 2025-11-22
docs(spec): Updated specification to render data models as a table generated from the proto instead of a raw proto snippet

darrelmiller:
In the currently published specification the data model definitions include a proto snippet and then have a list of fields under the snippet.  This creates duplicated information.  To prevent this duplication while maintaining an easy to read presentation of the abstract data models, this PR uses a Python macro to convert the proto into a markdown table.

Also took the opportunity to do an editorial pass on the comments in the a2a.proto before the field descriptions were removed from the specification.md.  Some of this content came from the types.ts file that often had information that was not in a2a.proto.

A preview of the changes can be seen here https://purple-moss-0838db51e.3.azurestaticapps.net/specification/#41-core-objects

# Proto to Markdown Table Conversion

The macro system provides two main functions (`proto_to_table` and `proto_enum_to_table`) that convert Protocol Buffer definitions into formatted markdown tables:

1. **Locates Definitions**: Reads the `.proto` file and uses region markers (`--8<-- [start:Name]` and `--8<-- [end:Name]`) to extract specific message or enum definitions.

2. **Parses Structure**: Analyzes the proto content to separate three sections:
   - **Description**: Comments appearing before the message/enum declaration
   - **Fields/Values**: The actual field or enum value definitions with their inline comments
   - **Notes**: Comments appearing after the closing brace

3. **Extracts Field Details**: For each field, it captures:
   - Field name (converted from snake_case to camelCase, or using `json_name` annotation)
   - Type (mapped to readable formats like "string", "array of integer", "map of string")
   - Required status (based on `REQUIRED` annotation or `optional` keyword)
   - Description (from preceding comment lines)

4. **Generates Output**: Assembles the markdown output with:
   - Description text at the top
   - A formatted table with columns for Field/Type/Required/Description (or Value/Description for enums)
   - Special notes for `oneof` groups
   - Additional notes from comments after the definition

The conversion automatically handles multi-line field definitions, map types, repeated fields, and filters out protolint directives and region markers from the output.

### DIFF
--- a/.mkdocs/macros.py
+++ b/.mkdocs/macros.py
@@ -10,6 +10,15 @@ from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 
+def _is_primitive_type(proto_type: str) -> bool:
+    """Check if a proto type is a primitive type that shouldn't be linked."""
+    primitive_types = {
+        'string', 'int32', 'int64', 'bool', 'bytes',
+        'google.protobuf.Struct', 'google.protobuf.Timestamp'
+    }
+    return proto_type in primitive_types
+
+
 def define_env(env):
     """
     Define custom macros for MkDocs.
@@ -417,7 +426,8 @@ def _parse_proto_enum_full(content: str) -> Tuple[str, List[Dict[str, str]], str
 
 def _format_proto_type(proto_type: str, is_repeated: bool) -> str:
     """
-    Format proto type to a more readable format.
+    Format proto type to a more readable format with links to specification.
+    Non-primitive types are automatically linked to their anchor in specification.md.
     """
     # Map proto types to readable types
     type_map = {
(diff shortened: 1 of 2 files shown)


holtskinner:
Can you update the tables to have the types also be in `code font`?

Also, if the types that are part of the spec could be linked inline, that would be even better!

4Jts76hmiQQdzMU

darrelmiller:
@holtskinner Using code font for types is a great idea and should be easy.  I really want to do the linking thing.  However, we will need to enforce some kind of convention for bookmarks in specification.md in order to allow the python macro to generate them.  I will definitely try to do that.

darrelmiller:
@holtskinner   All I can say is thank goodness AI knows how to write python because I would have given up long before figuring this out.  But good news, both of your requests are addressed. I added anchor tags in specification.md so that it was easy to point to the type.  When I say "I", you know what I mean. :-)

ToddSegal (approved):

PR #1235, opened 2025-11-24
fix(spec): Remove metadata field from ListTasksRequest

### LINKED ISSUE #1236: [Bug]: ListTasks doesn't support passing 'google.protobuf.Struct' via GET
### What happened?

ListTasks doesn't support passing 'google.protobuf.Struct' fields via HTTP GET. This prevents importing the new version of the specification, because it breaks service config generation: `http: map field 'google.protobuf.Struct.fields' referred to by message 'a2a.v1.ListTasksRequest' cannot be mapped as an HTTP parameter`.

### Code of Conduct

iliesicatrinel:
The ListTasksRequest message includes a google.protobuf.Struct metadata field, which cannot be represented as simple URL query parameter required by HTTP GET requests. This causes service config generation to fail with `http: map field 'google.protobuf.Struct.fields' referred to by message 'a2a.v1.ListTasksRequest`.

Since it's unclear whether metadata will be useful for ListTasks, I propose to remove it at least for now to fix the specification.

Fixes #1236 🦕

### DIFF
--- a/.mkdocs/macros.py
+++ b/.mkdocs/macros.py
@@ -60,8 +60,9 @@ def define_env(env):
         # Extract description (before message), fields, and notes (after message)
         description, fields, notes, oneof_groups = _parse_proto_message_full(message_content)
 
+        # Handle the case of empty proto message
         if not fields:
-            return f"**Error:** No fields found in message '{message_name}'"
+            return "None"
 
         # Build the output
         output = []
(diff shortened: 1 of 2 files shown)


darrelmiller:
@herczyn Technically we are currently supposed to have two TSC reviews to sign off on small changes and a full vote for more significant decisions. Maybe @amye  can change the branch protection rules to enforce the two sign offs.

I think this change is fine.  We can always add it back once we figure out how to do the automatic translation.  We may have to declare it as a `string` in the proto and then define the serialization rules to create the metadata as a string.

Issue #1240, opened 2025-11-26
[Bug]: JSON-RPC protocol Bindings for  SendMessage Response is confusing


ehsavoie:
### What happened?

In the current specifications the Response format to the *SendMessage* operation is a bit confusing as it states:
Response:

{
  "jsonrpc": "2.0",
  "id": 1,
  "result": { /* Task or Message object */ }
}
while there is no way to be aware of the kind of the payload. The expected format should follow the SendMessageResponse type in the a2a.proto

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1241, opened 2025-11-26
fix(doc): Makes JSON-RPC SendMessage response clearer

ehsavoie:
Fixing issue #1240 by providing something more in line with what is expected.

Fixes https://github.com/a2aproject/A2A/issues/1240 🦕


ToddSegal:
lgtm. /vote

darrelmiller:
We have two TSC votes which is enough for an editorial change. Merging.

Issue #1104, opened 2025-09-25
[Feat]:


yarolegovich:
#### Extend AgentInterface

Make `protocolVersions` a part of `AgentInterface`.
```patch
additionalInterfaces?: AgentInterface[];
...
export interface AgentInterface {
  url: string;
  transport: TransportProtocol | string;
+ protocolVersions?: string[];
}
```
The structure was used for transport negotiation and this seems to extend naturally into "transport+protocol version negotiation".

We don't introduce any new implicit field coupling, but more like extend `(URL, PrefferedTransport) in the root + additional (url, transport) pairs` relation into `(URL, PreferredTransport, protocolVersion) in the root + additional (url, transport, versions) triples`. 
For better consistency we might want to keep protocolVersion singular, but this introduces redundancy if a server wants to support multiple versions on the same endpoint. 

If we make `protocolVersions` an optional field and treat its absence as "supports only the interface specified in the protocolVersion on the root level" all existing AgentCards stay compatible with the new format.

#### Extend AgentInterface

Make `protocolVersions` a part of `AgentInterface`.
```patch
additionalInterfaces?: AgentInterface[];
...
export interface AgentInterface {
  url: string;
  transport: TransportProtocol | string;
+ protocolVersions?: string[];
}
```
The structure was used for transport negotiation and this seems to extend naturally into "transport+protocol version negotiation".

A tuple (URL, PrefferedTransport) in the root + additional (url, transport) pairs 
also extend into 
(URL, PreferredTransport, protocolVersion) in the root + additional (url, transport, versions)
For better consistency we might want to keep protocolVersion singular, but this introduces redundancy if a server wants to support multiple versions on the same endpoint. 

If we make `protocolVersions` an optional field and treat its absence as "supports only the interface specified in the protocolVersion on the root level" all existing AgentCards stay compatible with the new format.

#### Root-level field changes

Considered different variations like:

Adding a new field with a different name:
```patch
  "protocolVersion": "0.4.0",
+ "supportedVersions": ["0.4.0", "1.0.0"],
```

Adding a new field with a similar name:
```patch
  "protocolVersion": "0.4.0",
+ "protocolVersions": ["0.4.0", "1.0.0"],
```

Changing the existing field:
```patch
- "protocolVersion": "0.4.0",
+ "protocolVersion": ["0.4.0", "1.0.0"],
```
The first two version introduce new field coupling semantics to the type (do they need to be in sync, which should take preference etc.). 
[...]

Issue #652, opened 2025-05-25
[Feat]: Add Ability for Server to Specify Supported Protocol Versions


izzymsft:
### The Need to Specify Protocol Versions Supported by the Server
With the recent update to the protocol version from 0.1.0 to 0.2.1, it is now possible for clients to expect certain behavior from a server that does not support the protocol version(s) the are expecting.

We need to update the Agent card to specify a list of protocol versions supported by the server.

The client should specify in the request header (X-A2A-Protocol-Version) what protocol version they are using to communicate with the server using a required request header. If this header is not present the server should reject the request with a 400 status code and a message indicating that the required request header is missing.

This will help keeps things simple for both the server and the client.

### Maintaining Original Protocol Version
Also, if a server and the client can support communication with multiple protocol versions, the protocol version used to initiate a task or session should be used until the task is completed. The protocol versions should not change in the middle of the interaction/session.

Push notifications from the server to client should always maintain the original protocol version used to estabilish the interaction and subsequent responses from the server and follow up requests from the client should also maintain the original protocol version.

The server or client should respond with a 400 or 409 status code when conflicting protocol versions are presented in the request header that differs from the original/current  protocol version

### Next Steps
I am working on a PR for this spec change

### Request Header from Client

X-A2A-Protocol-Version: 0.2.1

### Agent Card Showing Supported Protocol Version

````typescript
// An AgentCard conveys key information about an A2A Server:
// - Overall identity and descriptive details.
// - Service endpoint URL.
// - Supported A2A protocol capabilities (streaming, push notifications).
// - Authentication requirements.
// - Default input/output content types (MIME types).
// - A list of specific skills the agent offers.
interface AgentCard {
  // Human-readable name of the agent (e.g., "Recipe Advisor Agent").
  name: string;
  // A human-readable description of the agent and its general purpose.
  // [CommonMark](https://commonmark.org/) MAY be used for rich text formatting.
  // (e.g., "This agent helps users find recipes, plan meals, and get cooking instructions.")
  description: string;
  // The base URL endpoint for the agent's A2A service (where JSON-RPC requests are sent).
  // Must be an absolute HTTPS URL for production (e.g., `https://agent.example.com/a2a/api`).
  // HTTP MAY be used for local development/testing only.
  url: string;
  // Information about the organization or entity providing the agent.
  provider?: AgentProvider;
  // Version string for the agent or its A2A implementation
  // (format is defined by the provider, e.g., "1.0.0", "2023-10-26-beta").
[...]

PR #1259, opened 2025-11-30
fix(spec): Clarify versioning strategy and client responsibilities in protocol specification

### LINKED ISSUE #1258: [Bug]: Clarify client version fallback strategy
### What happened?

The specification needs to clarify what happens when a client application running v1.2 of the protocol talks to an agent running v1.1 of the protocol.  

### Relevant log output

```shell

```

### Code of Conduct

darrelmiller:
Fixes #1258

### High level summary of changes:

- protocolVersion in AgentCard is now an array called protocolVersions
- Servers are expected to return UnsupportedVersionError if they get a request with an A2A-Version header higher than they support.
- Clients are expected to fallback to a supported version if that is acceptable to them. Tooling should help here.
- Not every minor version needs to be enumerated in the protocolVersions array. 2.2 infers support for 2.1 and 2.0.


docs/specification.md
```diff
@@ -702,11 +702,25 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
 
 ### 3.6 Versioning
 
-The specific version of the A2A protocol in use is identified using the `Major.Minor` elements (e.g. `1.0`) of the corresponding A2A specification version. Patch version numbers do not affect protocol compatibility, SHOULD NOT be included in requests and responses, and MUST not be considered when clients and servers negotiate protocol versions.
+The specific version of the A2A protocol in use is identified using the `Major.Minor` elements (e.g. `1.0`) of the corresponding A2A specification version. Patch version numbers used by the specification, do not affect protocol compatibility. Patch version numbers SHOULD NOT be used in requests, responses and Agent Cards, and MUST not be considered when clients and servers negotiate protocol versions.
```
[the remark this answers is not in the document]
darrelmiller: This suggestion changes the meaning of the sentence.

docs/specification.md
```diff
@@ -702,11 +702,25 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
...
 
-Agents declare support for latest supported protocol version in the `protocolVersion` field in the Agent Card. Agents MAY also support earlier protocol versions. Clients SHOULD specify the desired protocol version in requests using the `A2A-Version` header. If the requested version is not supported by the agent, the agent MUST return a `VersionNotSupportedError`.
+Agents declare support for latest supported protocol version in the `protocolVersion` field in the Agent Card. Agents MAY also support earlier protocol versions.
```
holtskinner: Can you add an inline link to the definition for `protocolVersion`?

docs/specification.md
```diff
@@ -702,11 +702,25 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
...
 
-It is RECOMMENDED that clients send the `A2A-Version` header with each request to reduce the chances of being broken if an agent upgrades to a new version of the protocol. Sending the `A2A-Version` header provides visibility to agents about version usage in the ecosystem, which can help inform the risks of inplace version upgrades.
+#### 3.6.1 Client Responsibilities
+
+It is RECOMMENDED that clients send the `A2A-Version` header with each request to maintain compatibility after an agent upgrades to a new version of the protocol. Sending the `A2A-Version` header also provides visibility to agents about version usage in the ecosystem, which can help inform the risks of inplace version upgrades.
```
holtskinner, line 711: Can you add an example of this?

docs/specification.md
```diff
@@ -702,11 +702,25 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
...
+
+#### 3.6.2 Server Responsibilities
+
+Agents that receive requests with a supported `A2A-Version` MUST process the request according to the semantics defined in the latest patch version of the specification with the same `Major.Minor` values. If the requested version is not supported by the agent, the agent MUST return a `VersionNotSupportedError`.
```
holtskinner: Can you add an inline link to `VersionNotSupportedError`?

holtskinner (dismissed):

docs/specification.md
darrelmiller: Yep. will do.

Tehsmash: I think we mentioned in the TSC call that we should make protocolVersion a list to make it easier for older clients to still find newer Agents that they can talk to. Do we want to update this here, or merge this and follow up with making it a list?  (Related ticket https://github.com/a2aproject/A2A/issues/1104)

docs/specification.md
Tehsmash: One issue I see here: is how does the Client know that the Agent is compliant with the latest patch version of the specification? 

If I have an Agent which is running 1.2.0 and then tomorrow 1.2.1 is released from the A2A repo.
Client's asking for 1.2 are being told to assume that the Agent 1.2.1 but it is actually still being handled as 1.2.0 because I've not had a chance to upgrade my Agent yet.

I think we might need to word this like:

```
MUST process the request according to the semantics defined in the latest patch version with the same `Major.Minor` that they support.

Agent's should try to upgrade to the latest patch version as soon as possible after the specification is released.
```

I also wonder if the Agent should return the full specific version in the header, so that clients are able to see which specific version and potentially which security vulns they are exposed too by using this version.

docs/specification.md
```diff
@@ -702,11 +702,25 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
...
+
+#### 3.6.3 Client Fallback
+
+Clients that receive a `VersionNotSupportedError` can choose to retry the request with an earlier supported version, or fail the request. This explicit failure handling helps prevent unexpected behavior that could arise if an agent processes a request containing protocol features or fields it does not recognize.
```
Tehsmash, line 746: Its not described anywhere that Clients should try to use the latest version they support by default. This should probably be added to the Client responsibilities.

If a user of a client asked for an older version of the spec by default and they get `VersionNotSupportedError` can they then negotiate a higher version, if both client and server support it.

Ideally with the "protocolVersions" list we discussed in the TSC, clients will pre-empt this negotiation anyway and pick a version which they both understand based on the AgentCard details.

docs/specification.md
ToddSegal, line 707: Should we be noting that different Major versions are not guaranteed to be compatible with each other. But "minor version N is compatible with all minor versions N-1, N-2, ... for the same major version M". 

Or does that not belong to this PR?

amye: Yes, but probably not in this PR specifically. I think that's a separate call out, but I will defer to @darrelmiller.

darrelmiller: > Should we be noting that different Major versions are not guaranteed to be compatible with each other. But "minor version N is compatible with all minor versions N-1, N-2, ... for the same major version M".
> 
> Or does that not belong to this PR?

I wasn't sure what a major update means with this strategy.  Thinking more about this, I think what this means is when an agent upgrades 1.2 to 1.3, there is an expectation that a client will be able to fallback to 1.2 and the agent still support the 1.2 client.  Does this mean that when an agent releases 2.0, it doesn't have to support 1.9 anymore?  Or does it mean that the 2.0 release should be at a different URL?

Tehsmash: I think we said that we follow Semantic versioning so:

Major = breaking, deprecated fields should be removed, a client on 1.1 should not expect a 2.0 version to accept its request because fields might be missing etc.

Minor = additions only, older clients using old fields (1.1) should still continue to work on newer versions (1.2). However, this works today for clients sending requests, however there is a problem in responses because there could be unknown fields for the client and the strict validation of the response will fail. We could allow clients to ignore additional fields when parsing the response. 
 
Patch = bug fixes only

darrelmiller: @Tehsmash With the proposed text, even a v1.2 client will get a versionUnsupportedError from a v1.1 Agent and the client is expected to re-request with a v1.1 header.
It is not clear to me that this behaviour needs to change for major version bumps.

It is true that an agent that supports v1.2 should continue to allow v1.2 clients to use fields that were deprecated in v1.2. And major version bumps will allow removal of deprecated properties. I think that is mentioned in the document introduction. I will review and see if some reorg is required.

darrelmiller: Also, patch versions are only editorial and clarifications. The term "bug fix" is nebulous when it comes to spec updates because it raises the question "bug according to who?". The response of "well we meant to write this behaviour not that behaviour" in the spec is not very helpful to implementations.
So, patch versions should almost never create behaviour changes.

ToddSegal: 1.4 servers should be able to interact with ANY 1.X client. These servers should have full fidelity/100% support for all features in 1.0-1.4. Anything specific feature in 1.X where X > 4, will be ignored/not supported.

Tehsmash: I'm not sure about 1.5+ Client talking to 1.4 server without "downgrading" itself.
I'm concerned about the client sending parameters to the server and them getting silently ignored especially if there is a behaviour difference related to the sent options.
For example imagine if we added "blocking: true/false" in 1.5 version, a 1.4 server ignoring this would be a completely different behaviour, however a client 1.4 or below should work with a 1.4 server.

ToddSegal: "blocking: true/false" would be required to be an optional parameter with a well-specified default if omitted to be a "minor version" update. If it is required and omission creates a behavior change, it is breaking and therefore major.

docs/specification.md
darrelmiller: @Tehsmash Nobody should ever have to update anything based on patch versions. They are editorial only.  If someone deployed bad code because of bad wording that gets fixed in an edit, then there may be a compatibility issue, but we should consider that no different than if an implementer simply misread a specification.  Patch versions are only to indicate that we changed some text in the specification, not that we changed the protocol in any way, even "bug fixes".

docs/specification.md
darrelmiller: For the moment I am not going to add a recommendation for a client to use the latest version they support.  I don't want to infer that a client should do a negotiation if they already know based on protocolVersions that the agent doesn't support the client's latest version.  

I will add the list of protocolVersions to this PR if I don't find another one covering it.

docs/specification.md
darrelmiller: @ToddSegal Based on our discussions around ProtoJSON not being able to parse unrecognized properties, we realized that we would not be able to round-trip these unrecognized properties.
Consider we added an optional Skill_Id to the message object in v1.1.  A v1.0 agent would not be able to process that and store it in the Task Message History.  Using a default value would indicate that no skill was associated, but that would not be true.
The current proposal is to require the client to acknowledge that the agent does not support v1.1 and resend the request using a v1.0 compliant message if the client accepts that v1.1 behaviour is not essential.  This can be automated in SDK tooling if the client decides it is fine with auto-fallback.
I think this makes every agent's life easier because nobody has to deal with messages from different versions of the protocol that it doesn't understand. 

Related to your comment about an agent supporting all previous minor versions, this has an impact on the proposal to change "protocolVersion" into "protocolVersions" in the agentCard.  I was concerned about requiring agents to have
`protocolVersion: ["1.0","1.1","1.2","1.3","1.4","1.5"]`
which seems pretty redundant.  Would any agent not support just "1.2" for example?

Maybe the only reason for multiple protocol versions is so that we can do this
`protocolVersions: ["0.3","1.3"]`

This would indicate that "0.3", "1.1", "1.2" and "1.3" are supported.  There is an edge case that versions "0.2" and "0.1" are likely not supported because there were breaking changes there.
We could maybe add wording saying that "any version with a major value of 1 or greater automatically infers all prior minor versions are supported".

darrelmiller: > 1.4 servers should be able to interact with ANY 1.X client.

With the above strategy, if an agent really does want to explicitly support any 1.X client, then they could do the following:

`protocolVersions: ["1.99"]`

and attempt to handle unrecognized properties.

docs/specification.md
```diff
@@ -987,7 +1010,7 @@ Agents declare their supported extensions in the [`AgentCard`](#441-agentcard) u
 
 ```json
 {
-  "protocolVersion": "0.3.0",
+  "protocolVersions": ["0.3.0"],
```
Tehsmash:
```suggestion
  "protocolVersions": ["0.3"],
```

docs/specification.md
```diff
@@ -704,11 +704,50 @@ Push notifications are delivered via HTTP POST to client-registered webhook endp
 
 ### 3.6 Versioning
 
-The specific version of the A2A protocol in use is identified using the `Major.Minor` elements (e.g. `1.0`) of the corresponding A2A specification version. Patch version numbers do not affect protocol compatibility, SHOULD NOT be included in requests and responses, and MUST not be considered when clients and servers negotiate protocol versions.
+The specific version of the A2A protocol in use is identified using the `Major.Minor` elements (e.g. `1.0`) of the corresponding A2A specification version. Patch version numbers used by the specification, do not affect protocol compatibility. Patch version numbers SHOULD NOT be used in requests, responses and Agent Cards, and MUST not be considered when clients and servers negotiate protocol versions.
```
Tehsmash, line 707: > Patch version numbers SHOULD NOT be used in requests, responses and Agent Cards, and MUST not be considered when clients and servers negotiate protocol versions.

From a security perspective it could be useful for a client to know the exact patch version the server is running assuming that we use patch versions to include that kind of bug fix in the spec.

So perhaps:

```
Patch version numbers SHOULD NOT be used in requests and Agent Cards, and MUST not be considered when clients and servers negotiate protocol versions. Responses should include the complete negotiated version number providing clients information about the exact protocol version running on the Agent side.
```

Actually as a separate issue I don't want to derail this PR, this makes me wonder whether we should allow clients to specify a minimum patch version requirement to ensure they can say something like >=1.0.4 (where 1.0.4 contains a security fix).

docs/specification.md
darrelmiller: Thanks. I thought I had caught all of those.

docs/specification.md
darrelmiller: In previous discussions on this topic, we had said that SDKs would leverage the patch version to identify updates to the SDK.  e.g.  Python SDK might be at 1.0.34 and .NET at 1.0.11 and they both support the 1.0 protocol. 
If we allow using the specification version to indicate security fixes, then it is going to get confusing to say Python SDK 1.0.35 includes Specification 1.0.1 security fix.  I would guess most security fixes are going to be in the SDK rather than the specification, and if there is a security bug in the specification that needs closing, then I'm fine with us using a minor version for that.

Tehsmash: Does that mean that the spec should have no patch versions for any reason?

I'm not 100% sure keeping the SDK's version in sync with the spec's version is a good idea. In python for example the ecosystem uses semantic versioning pretty heavily to indicate bug fixes only, compatible and incompatible changes. Many people will pin to a Major.Minor but allow for patch releases through `~` operator in requirements. Using patch number for all releases breaks that assumption and also will prevent SDKs from making breaking API changes without a major spec release happening at the same time.

Also if SDKs start to support multiple versions of the spec so that clients can talk to agents on different versions, having the SDK version in sync with the spec adds to the confusion. i.e. Python SDK 1.0.34 supports 1.0 and 0.3 but Python SDK 2.0.3 supports 2.0 and 1.0 but not 0.3.

Tehsmash: As discussed in the TSC call, patch releases for the spec repo should only be for non-functional changes to the spec text / comments in the proto etc. As soon as functional changes to the a2a.proto are made this requires a minor release. Do we need to put this in the spec somewhere?

izzymsft:
It is great to see that [this feature request](https://github.com/a2aproject/A2A/issues/652) is now merged into the protocol specification. I think we can now close the original issue/feature request.

PR #1269, opened 2025-12-03
feat(spec): Remove v1s from a2a url http bindings.

### LINKED ISSUE #1268: [Feat]: Remove v1 from the urls

Currently the url has to contain v1 at the beginning. 

This is unnecessary and should be removed.

_No response_

### Additional context

_No response_

### Code of Conduct

herczyn:
these can still be part of the agent card.

fixes #1268

built on top of https://github.com/a2aproject/A2A/pull/1195

BEGIN_COMMIT_OVERRIDE
feat(spec)!: Remove v1s from a2a url http bindings

Release-As: 1.0.0
END_COMMIT_OVERRIDE

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -716,10 +716,6 @@ message SendMessageRequest {
     (google.api.field_behavior) = REQUIRED,
     json_name = "message"
   ];
-  Message request = 1 [
-    (google.api.field_behavior) = REQUIRED,
-    json_name = "message"
-  ];
   // Configuration for the send request.
   SendMessageConfiguration configuration = 2;
   // A flexible key-value map for passing additional context or parameters.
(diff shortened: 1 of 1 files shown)


specification/grpc/a2a.proto
```diff
@@ -366,6 +398,8 @@ message AgentInterface {
   // easily extended for other protocol bindings. The core ones officially
   // supported are `JSONRPC`, `GRPC` and `HTTP+JSON`.
   string protocol_binding = 2 [(google.api.field_behavior) = REQUIRED];
+  // Tenant to be set in the request when calling the agent.
+  string tenant = 3;
```
Tehsmash, line 402: This doesn't seem like this belongs here, AgentInterface isn't an input into any of the Operations.

specification/grpc/a2a.proto
```diff
@@ -675,9 +709,13 @@ message PasswordOAuthFlow {
...
 message SendMessageRequest {
+  // Optional tenant, provided as a path parameter.
+  string tenant = 4;
   // The message to send to the agent.
-  Message request = 1
-      [(google.api.field_behavior) = REQUIRED, json_name = "message"];
+  Message request = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    json_name = "message"
+  ];
```
Tehsmash, line 718: I've noticed this happening in a few PRs, it looks like different people have rules in their development environments for line length, this change probably shouldn't be in this PR.

specification/grpc/a2a.proto
herczyn: this was from the parent PR, fixed.

specification/grpc/a2a.proto
herczyn: fixed

PR #1279, opened 2025-12-05
docs(spec): patched security scheme and add anchor tag for Security

vinoo999:
Minor patch to `security_scheme` reference to `securityScheme` and added a missing anchor tag for `Security`. Another docs change should be added to clarify usage of `Security`. Opened https://github.com/a2aproject/A2A/discussions/1278 for clarification.

Fixes https://github.com/a2aproject/A2A/issues/1280


holtskinner:
This is a good question, It seems like `Security` has been mostly superseded by `SecurityScheme`

https://github.com/a2aproject/A2A/blob/6e65c7fbaf1de394f82ad21316202e0bcdd30dbe/specification/grpc/a2a.proto#L559-L580

Not sure if there's a need for both. Seems like we should deprecate/remove `Security`. Would be a breaking change, but a good cleanup to add to 1.0

What do you think @darrelmiller ?

vinoo999:
@muscariello @holtskinner want me to make the change in AgentSkill to remove and make the breaking change directly if vote passes?

+ discussion (in that case the anchor tag not needed and we do strict removal)
https://github.com/a2aproject/A2A/discussions/1278

PR #1288, opened 2025-12-08
fix(spec): update last_updated_after to be a timestamp

### LINKED ISSUE #1287: [Bug]: ListTasksRequest incorrect timestamp type
### What happened?

`ListTasksRequest` uses int64 for `last_updated_after`:
```proto
  // Filter tasks updated after this timestamp (milliseconds since epoch).
  // Only tasks with a last updated time greater than or equal to this value will be returned.
  int64 last_updated_after = 6;
```
Should be `google.protobuf.Timestamp` instead, for consistency with the rest of the protocol.
```proto
  // ISO 8601 Timestamp when the status was recorded.
  // Example: "2023-10-27T10:00:00Z"
  google.protobuf.Timestamp timestamp = 3;
```

### Code of Conduct

yarolegovich:
Update `ListTasksRequest.last_updated_after` to use `google.protobuf.Timestamp` for consistency with the rest of the protocol.

fixes #1287


yarolegovich:
> Be sure to make any fixes needed in the Documentation as well.

Double-checked, couldn't find field references anywhere.

darrelmiller:
This was addressed in #1358

Issue #1285, opened 2025-12-07
[Bug]: Is "supportedInterfaces" information that must be declared in the AgentCard?


MoonRendering:
### What happened?

In the following link (https://a2a-protocol.org/latest/specification/#441-agentcard), we found that the "Required" tag for "supportedInterfaces" is marked as "No."
In another link (https://a2a-protocol.org/latest/specification/#83-protocol-declaration-requirements), we found the following description: "The AgentCard MUST properly declare supported protocols."
1. Please help determine whether these two descriptions conflict with each other, or adjust the descriptions to avoid any apparent conflict.
2. Please clarify whether "supportedInterfaces" is information that must be declared in the AgentCard.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

darrelmiller:
@MoonRendering  This was probably not marked as required because previously the "url" and "preferredTransport" were still in the specification but marked as deprecated. Now that they are being removed, supportedInterfaces needs to be required.  I'll add a note to #1301

Tehsmash:
This got addressed as part of the v1.0.0 spec, closing this issue.

PR #1301, opened 2025-12-11
fix(spec)!: Remove deprecated fields from a2a.proto for v1.0 release

### LINKED ISSUE #1227: [Bug]: Remove deprecated fields from a2a.proto for v1.0 release
## Description

As part of the v1.0 release, deprecated fields should be removed from `a2a.proto` since this is a major version change and provides a clean break for backward compatibility.

## Deprecated Fields to Remove

The following fields in the `AgentCard` message are currently deprecated and should be removed:

1. `url` (field 3) - deprecated in favor of `supported_interfaces`
2. `preferred_transport` (field 14) - deprecated in favor of `supported_interfaces`
3. `additional_interfaces` (field 15) - deprecated in favor of `supported_interfaces`

All three fields were replaced by the `supported_interfaces` field (field 19).

## Location

`specification/grpc/a2a.proto`
Lines: 391-395

## Action Items

- [ ] Remove the three deprecated fields from the `AgentCard` message
- [ ] Update any documentation that references these deprecated fields
- [ ] Consider if any migration guide notes are needed in the changelog

## Context

The v0.3 to v1.0 update is a major version change, making this an appropriate time to clean up deprecated fields and simplify the protocol definition.

Tehsmash:
As v1.0 is a major release and may contain breaking changes this commit cleans up the deprecated fields in the a2a.proto and references to this in the spec doc.

Closes: #1227

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1926,22 +1926,13 @@ The AgentCard **MUST** properly declare supported protocols:
 - Each interface **MUST** accurately declare its transport protocol and URL
 - URLs **MAY** be reused if multiple transports are available at the same endpoint
 
-**Backward Compatibility:**
-
-For backward compatibility, agents **MAY** continue to populate the deprecated fields (`url`, `preferredTransport`, `additionalInterfaces`) alongside `supportedInterfaces`:
-
-- The `url` field **SHOULD** match the URL of the first entry in `supportedInterfaces`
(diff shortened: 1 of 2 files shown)


darrelmiller:
@Tehsmash Now that we are removing url and preferredTransport, we need to make `supportedInterfaces` required otherwise the client has no URL to find the agent. See #1285

darrelmiller (approved):

Issue #1230, opened 2025-11-24
[Bug] Inconsistent field naming for Message type (request/msg/message)


darrelmiller:
## Bug Description

There's an inconsistency in how `Message` fields are named across the proto file, due to `message` being a reserved keyword in Protocol Buffers.  It is best practice to not use reserved words for field names.

## Current State

The proto file uses different field names for `Message` objects in different contexts:

### 1. **`request`** - SendMessageRequest (line 679)
```protobuf
message SendMessageRequest {
  Message request = 1
      [(google.api.field_behavior) = REQUIRED, json_name = "message"];
  // ...
}
```
- Field name: `request`
- JSON name: `message`

### 2. **`msg`** - SendMessageResponse & StreamResponse (lines 810, 822)
```protobuf
message SendMessageResponse {
  oneof payload {
    Task task = 1;
    Message msg = 2 [json_name = "message"];
  }
}

message StreamResponse {
  oneof payload {
    Task task = 1;
    Message msg = 2 [json_name = "message"];
    // ...
  }
}
```
- Field name: `msg`
- JSON name: `message`

### 3. **`message`** - TaskStatus (line 182)
```protobuf
message TaskStatus {
  // ...
  Message message = 2;
  // ...
}
```
- Field name: `message` (no json_name override needed)

## Problem

The inconsistency has the potential to cause confusing.

### Option 1

Standardize all `Message` field names to use `message` consistently:

```protobuf
// Option 1: Use 'message' everywhere 
message SendMessageRequest {
  Message message = 1 [(google.api.field_behavior) = REQUIRED];
  // ...
}

message SendMessageResponse {
  oneof payload {
    Task task = 1;
    Message message = 2;
  }
}

message StreamResponse {
  oneof payload {
    Task task = 1;
    Message message = 2;
    // ...
  }
}
```

### Option 2

```protobuf
// Option 2: If 'message' truly is problematic in some contexts, 
// consistently use 'msg' everywhere with json_name override
message SendMessageRequest {
  Message msg = 1 
      [(google.api.field_behavior) = REQUIRED, json_name = "message"];
  // ...
}

message TaskStatus {
  TaskState state = 1 [(google.api.field_behavior) = REQUIRED];
  Message msg = 2 [json_name = "message"];
  // ...
}
```

## Impact

- **Breaking change** for gRPC clients (field name changes)
- **Non-breaking for JSON/HTTP** clients (if json_name is used consistently)
- **Improves consistency** and reduces confusion
[...]

PR #1302, opened 2025-12-11
fix(spec)!: Make "message" field name consistent between protocol bindings

Tehsmash:
While this is a reserved word for proto itself, best practise is to avoid field names that conflict with language (python etc) keywords, as "message" isn't a language keyword this improves the consistency and readability when using the types generated from the proto and avoids needing to translate the name for JSON encoding.

Closes: #1230

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -909,6 +909,7 @@ For detailed security guidance on push notifications, see [Section 13.2 Push Not
 
 ### 4.5. Security Objects
 
+<a id="Security"></a>
 <a id="SecurityScheme"></a>
 
 #### 4.5.1. SecurityScheme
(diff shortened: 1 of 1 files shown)


Tehsmash:
@darrelmiller @muscariello @geneknit I tested this through buf generate in the a2a-python repo and it generates correctly:

```
class SendMessageRequest(_message.Message):
    __slots__ = ()
    TENANT_FIELD_NUMBER: _ClassVar[int]
    MESSAGE_FIELD_NUMBER: _ClassVar[int]
    CONFIGURATION_FIELD_NUMBER: _ClassVar[int]
    METADATA_FIELD_NUMBER: _ClassVar[int]
    tenant: str
    message: Message
    configuration: SendMessageConfiguration
    metadata: _struct_pb2.Struct
    def __init__(self, tenant: _Optional[str] = ..., message: _Optional[_Union[Message, _Mapping]] = ..., configuration: _Optional[_Union[SendMessageConfiguration, _Mapping]] = ..., metadata: _Optional[_Union[_struct_pb2.Struct, _Mapping]] = ...) -> None: ...
```

Issue #1255, opened 2025-11-29
[Feat]: Move media type from FilePart to base Part to support all part types


darrelmiller:
## Feature Proposal

Move the `mediaType` field from `FilePart` to the base `Part` object, allowing all part types to specify a media type.  Currently if you need to provide a media type, you have to send it as a `FilePart` which requires base64 encoding if you want to include the content inline.  base64 encoding is not particularly efficient and is not great for debugging.

## Current Structure

Currently, only `FilePart` has a `mediaType` field:

```protobuf
message Part {
  oneof part {
    string text = 1;
    FilePart file = 2;
    DataPart data = 3;
  }
}

message FilePart {
  string name = 1;
  string uri = 2;
  string mediaType = 3;  // Only available on FilePart
}

message DataPart {
  google.protobuf.struct data = 1;
}
```

## Proposed Structure

Move `mediaType` to the base `Part` object and make data an `any` type to hold structured data that conforms to the JSON object model:

```protobuf
message Part {
  string mediaType = 4;  // Available for all part types
  oneof part {
    string text = 1;
    FilePart file = 2;
    google.protobuf.any data = 3;
  }
}

message FilePart {
  string name = 1;
  string uri = 2;
  // mediaType removed from here
}
```

### 1. **TextPart with Media Type**
Text parts can specify their format explicitly:
- `text/plain` - Plain text
- `text/markdown` - Markdown formatted text
- `text/html` - HTML content
- `text/csv` - CSV data

```json
{
  "mediaType": "text/markdown",
  "text": "# Hello\n\nThis is **formatted** markdown."
}
```

### 2. **DataPart with Media Type**
Data parts can indicate their content type:
- `application/json` - JSON data
- `application/city+json`  - Standardized vocabularies based on JSON e.g. https://www.cityjson.org/specs/2.0.1/

```json
{
  "mediaType": "application/json",
  "data": {
       "x": 24,
       "y": 32
   }
}
```

### 3. **Consistent API**
All parts have the same interface for specifying content type, making it easier for:
- Clients to handle content type detection uniformly
- Agents to specify content types for any part
- Middleware to process or transform content based on type

### 4. **Better Content Negotiation**
Clients and agents can negotiate content types for all message parts, not just files:
- Request text in markdown vs plain text
- Specify preferred data serialization format
- Handle multiple representations of the same content

### Use Case 1: Formatted Text Responses
```json
{
  "role": "agent",
  "parts": [
    {
[...]

Issue #1300, opened 2025-12-10
[Feat]: Incorporate a field of type Any in the gRPC spec


xui-will:
### Is your feature request related to a problem? Please describe.

Artifacts generated by agents already have several forms of representation:
- `text` for plain text
- `dataPart` for JSON-like data
- `filePart` for URLs, URIs, or inline binary content

This covers most scenarios. However, gRPC provides an additional powerful option for representing messages in proprietary formats: `google.protobuf.Any`. This format is suitable for conveying raw data in an efficient manner.

We suggest adding a field of this type to the `Part` model, in order to enable agents to use the `Any` type, and avoid resorting to redundant serializations or workarounds.

A corresponding Part can also be added for other transports, mimicking the structure of the `Any` message and allowing general-purpose raw data to be exchanged across all transports, not just gRPC.

### Describe the solution you'd like

The proposed solution is to add a new type of part to the `Part` message/model.  
The message would be of type `Any` for the gRPC transport, and the model would be a new type `RawPart` for HTTP & JSON.

No additional nesting is required, since `Any` and `RawPart` already contain a `type_url` that describes how to interpret the raw data.

#### gRPC

```protobuf
message Part {
  oneof part {
    // ---------- 1..3 unchanged ----------
    string text = 1;
    FilePart file = 2;
    DataPart data = 3;
    // ---------- Add field of type Any ----------
    google.protobuf.Any raw_data = 5;
  }
  google.protobuf.Struct metadata = 4;
}
```

#### JSON & HTTP

For non-gRPC transports, the protobuf `Any` type can be emulated with a simple model.

`RawPart`:
| name     | type   | required | description |
|----------|--------|----------|-------------|
| type_url | string | Yes      | A URL or resource name whose content describes the type of the serialized message. |
| value    | bytes  | Yes      | Must be valid serialized data of the type specified above. |

#### Workaround

- Serialize the data class to wire format using Protobuf's `SerializeToString()` helper.
- Then either:
  - Embed the resulting bytes as base64 inside a `dataPart` (not optimal),
  - Or: Smuggle it as-is, inside a filePart's `file_with_bytes` (which is both semantically incorrect and a hack).

#### Alternative solution

- Add a `file_with_any` field to `FilePart`. This would be a smaller change. However, it is too much of a shortcut, since objects conveyed in this field are likely to be anything but files.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

xui-will:
[...]

PR #1306, opened 2025-12-11
feat(spec)!: support arrays and primitives in `DataPart`

### LINKED ISSUE #1088: [Feat]: Allow DataPart.data to be a valid JSON data type (not just object)
https://github.com/a2aproject/A2A/blob/472c7362028150634f658c3dd0013a472090ead7/types/src/types.ts#L828-L835

Why must DataPart.data be an object? It would be particularly beneficial to accept data in list format like:

```json
{
  "kind": "data"
  "data": ["Apple", "Orange", "Banana"]
}
```

Thoughts?

Let me know if this should be opened as a PR or discussion instead. Thanks!

matoushavlena:
Changes `DataPart.data` from `google.protobuf.Struct` to `google.protobuf.Value` to support JSON arrays, primitives, and other non-object values.

**Breaking change:** The protobuf field type changes from `Struct` to `Value`. Code that directly accesses `DataPart.data` as a struct/map will need to be updated to handle the broader `Value` type, which can be an object, array, string, number, boolean, or null.

Resolves #1088
Related #1255

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -6,6 +6,7 @@ ACard
 AClient
 ACo
 ADK
+ADR
 AError
 AExecutor
 AGP
@@ -128,6 +129,8 @@ aboutasha
 achat
 aconnect
 adk
+adr
+adrs
 afet
 affef
 agentcard
@@ -284,6 +287,7 @@ postgresql
 pqr
 prefecthq
 protoc
+protojson
 protolint
 pyguide
 pylance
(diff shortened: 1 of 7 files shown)


xui-will:
This proposal introduces a breaking change at the gRPC/protobuf level, and the associated cost in terms of client updates, regressions, and debugging should not be overlooked. 

In protobuf, only field numbers are preserved on the wire, not the field definitions. Changing the type of an existing field creates a compatibility gap: a newer client interacting with an older agent (or vice versa) would be unable to reliably interpret DataPart, effectively breaking interoperability of one of the most heavily used field of the whole protocol. 

Alternative solutions should be considered. For instance,  and safer pattern would be to introduce a new field and deprecate the existing one rather than changing it in place.

Also, FWIW: We have made a different proposal to allow usage of the gRPC type `Any` which is even more powerful.
I would like to have our proposal considered in conjunction to this PR, as I feel it's either complementary or redundant. c.f. https://github.com/a2aproject/A2A/issues/1300

darrelmiller:
@xui-will The latest iteration of the thinking about changes to the Part type are documented in this comment here https://github.com/a2aproject/A2A/issues/1255#issuecomment-3679633802

We are moving to a v1.0, which is a breaking change from v0.3.  This is an opportunity we need to take before we release v1 and we will not want to make any breaking changes for hopefully a long time.

My understanding of `Any` is that it enables passing any gRPC type, but it expects the gRPC type identifier to be communicated. This is not simple to support with other protocol bindings.

Tehsmash:
@darrelmiller @muscariello I think we can close this in favor of Darrel's PR which refactors Part completely: https://github.com/a2aproject/A2A/pull/1363. This PR switches content.Data to a Value so that it supports arrays / other json primitives.

PR #1310, opened 2025-12-12
docs: fixing some typos in various files

didier-durand:
Hi,
Giving a second read to the repo, I found these additional typos: see commit diffs.
Didier


didier-durand:
@holtskinner :Hi, do you want me to also apply the change suggested by Gemini?

docs/topics/extensions.md
```diff
@@ -8,7 +8,7 @@ base protocol.
 
 Extensions allow for extending the A2A protocol with new data, requirements,
 RPC methods, and state machines. Agents declare their support for specific
-extensions in their Agent Card, and clients can then opt-in to the behavior
+extensions in their Agent Card, and clients can then opt in to the behavior
```
holtskinner, line 11: I think this one should stay as is

holtskinner, line 11:
```suggestion
extensions in their Agent Card, and clients can then opt-in to the behavior
```

didier-durand: Hi Holt, I believe that there is no hypen here since it's a verb.

image

didier-durand: No, it  should by without  hyphen: see the answer by Google that I pasted below

holtskinner: Interesting, didn't realize there was a distinction

PR #1323, opened 2025-12-23
feat: Add A2A CLI tool for Agent Card management and agent inter…

pascal910107:
Fixes https://github.com/a2aproject/A2A/issues/1325.
- Implement a comprehensive CLI tool for working with the A2A Protocol
- Add commands for generating, validating, testing, viewing, and simulating agent interactions
- Include JSON Schema validation, multi-format output, and shell completions


darrelmiller:
@pascal910107 Apologies if this comes across as unappreciative of the contribution, that isn't my intent. However, it isn't particularly helpful to create PRs without doing the appropriate due-diligence in the issue.  @zeroasterisk asked a question about the overlap between this tool and the existing tools.
Also, if we were to create a CLI, I don't think this is the right repo for it.

We are trying to close out PRs for the V1 release of the specification and having to spend time triaging unrelated PRs makes it just a little more difficult to achieve that goal.

We also need to be a little more rigid about this approach because with AI based code generation tools, it is now very easy to generate significant amounts of code to review.

We do appreciate your input, but please work with us to help us direct you so that your contributions deliver the most value.

I'm going to close this PR, not because I think it is a bad idea, or any judgement on the work you did. I just don't think this is the right place for it, and I don't want to other TSC members to have to spend time reviewing this PR.

PR #1358, opened 2026-01-13
fix(spec)!: Fixes for the last_updated_after field

Tehsmash:
This PR combines two fixes for the ListTaskRequest last_updated_after field one changes the field to a timestamp type and one renames it and clarifies the ordering of the task list result.

Original PR for the timestamp type fix was: https://github.com/a2aproject/A2A/pull/1288

Fixes: #1287
Fixes: #1206

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -6,6 +6,7 @@ ACard
 AClient
 ACo
 ADK
+ADR
 AError
 AExecutor
 AGP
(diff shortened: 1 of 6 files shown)


darrelmiller:
@a2aproject/a2a-tsc So, I slipped and merged this PR.  We should still vote.  If the vote doesn't pass, I will revert the change.

amye:
This has passed.

PR #1389, opened 2026-01-20
fix(spec)!: Switch to non-complex IDs in requests

Tehsmash:
Some operations like GetTask / SubscribeToTask still used complex / compound identifiers like `tasks/<uuid>` which requires the clients/servers to deconstruct/construct these IDs as they are normally using just the UUID to store the resource in the DB etc.

This commit refactors them so they are just the IDs, in some cases this meant that we needed to split one field into two i.e. parent + name.

Fixes: #1390

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -369,7 +369,7 @@ message AuthenticationInfo {
 // --8<-- [end:PushNotificationAuthenticationInfo]
 
 // --8<-- [start:AgentInterface]
-// Declares a combination of a target URL and a transport protocol for interacting with the agent.
+// Declares a combination of a target URL, transport and protocol version for interacting with the agent.
 // This allows agents to expose the same functionality over multiple protocol binding mechanisms.
 message AgentInterface {
   // The URL where this interface is available. Must be a valid absolute HTTPS URL in production.
@@ -381,6 +381,10 @@ message AgentInterface {
   string protocol_binding = 2 [(google.api.field_behavior) = REQUIRED];
   // Tenant to be set in the request when calling the agent.
   string tenant = 3;
+  // The version of the A2A protocol this interface exposes.
+  // Use the latest supported minor version per major version.
+  // Examples: "0.3", "1.0"
+  string protocol_version = 4 [(google.api.field_behavior) = REQUIRED];
 }
 // --8<-- [end:AgentInterface]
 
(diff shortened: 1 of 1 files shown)


amye:
@muscariello - still need a vote on this one if it's not 1.0?

specification/grpc/a2a.proto
```diff
@@ -822,22 +821,20 @@ message SetTaskPushNotificationConfigRequest {
...
   string tenant = 4;
-  // The parent task resource.
-  // Format: tasks/{task_id}
-  string parent = 1;
+  // The parent task resource id.
+  string parent = 1 [(google.api.field_behavior) = REQUIRED];
+
   // The maximum number of configurations to return.
   int32 page_size = 2;
-
```
Prakul95, line 841: Nit

specification/grpc/a2a.proto
```diff
@@ -125,12 +125,12 @@ service A2AService {
   // Delete a push notification config for a task.
   rpc DeleteTaskPushNotificationConfig(DeleteTaskPushNotificationConfigRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      delete: "/{name=tasks/*/pushNotificationConfigs/*}"
+      delete: "/tasks/{parent=*}/pushNotificationConfigs/{name=*}"
       additional_bindings: {
-        delete: "/{tenant}/{name=tasks/*/pushNotificationConfigs/*}"
+        delete: "/{tenant}/{parent=*}/pushNotificationConfigs/{name=*}"
```
Prakul95: Would it be- tasks/{parent=*}/?

Tehsmash: Great catch, thanks!

darrelmiller:
Can we use more readable names for these parameters like `taskId` and `configId` instead of `parent` and `name` ?

darrelmiller:
From TSC.  We will be renaming the parameters so as not to confuse AIP folks into thinking name is a complete resource name.

specification/grpc/a2a.proto
```diff
@@ -543,9 +543,8 @@ message AgentCardSignature {
 // A container associating a push notification configuration with a specific
 // task.
 message TaskPushNotificationConfig {
-  // The resource name of the config.
-  // Format: tasks/{task_id}/pushNotificationConfigs/{config_id}
-  string name = 1 [(google.api.field_behavior) = REQUIRED];
+  // The resource id of the config.
+  string id = 1 [(google.api.field_behavior) = REQUIRED];
```
mikeas1, line 525: I believe this message now needs tenant and task_id fields. Previously those would have been baked into the name field.

darrelmiller (approved):

muscariello:
Do we have to vote ? @darrelmiller @ToddSegal @amye ? I see PR review approvals but not as many votes.

specification/grpc/a2a.proto
Tehsmash: Thanks @mikeas1, I've added the task_id and tenant here, and it made me realise that the CreateTaskPushNotificationConfigRequest message had a bug of including the full TaskPushNotificationConfig message so I've fixed that too.

muscariello:
@geneknit @hughesthe1st @SivaNSAP @geneknit

PR #1396, opened 2026-01-21
fix: Remove unimplemented state_transition_history capability field

### LINKED ISSUE #1228: [Bug]: Evaluate removing state_transition_history capability from AgentCapabilities
## Description

The `state_transition_history` field in `AgentCapabilities` indicates whether an agent provides a history of state transitions for a task. However, there is currently no API method or mechanism to access this history.

## Current State

**Location:** `specification/grpc/a2a.proto`, line 451
```protobuf
message AgentCapabilities {
  // ...
  // Indicates if the agent provides a history of state transitions for a task.
  optional bool state_transition_history = 4;
}
```

## Problem

- Agents can advertise support for `state_transition_history` via their capabilities
- There is no corresponding RPC method or field in the `Task` message to retrieve this history
- Clients cannot access the state transition history even when an agent supports it
- This creates an incomplete feature that provides no value to users

## Questions to Consider

1. **Should we remove this capability?** Since there's no way to access the history, advertising support for it is misleading.

2. **Should we implement the feature instead?** If state transition history is valuable, we should add:
   - A field in the `Task` message to expose the history
   - Or a new RPC method like `GetTaskHistory` to retrieve it
   - Documentation on the expected format and use cases

3. **Is this planned for a future version?** If this is intentionally incomplete and planned for post-v1.0, we should document that clearly.

## Recommendation

For v1.0, I recommend **removing** this capability field since:
- It's a major version change, making this an appropriate time to remove incomplete features
- It creates confusion for implementers and users
- It can be re-introduced in a future version if/when the feature is fully implemented

If we do keep it, we should at minimum add a warning in the documentation that this is not yet accessible via the API.

## Related

See also: #1227 (removing deprecated fields)

madankumarpichamuthu:
Removes the `state_transition_history` field from `AgentCapabilities` message as it has no corresponding API implementation. This capability flag could mislead clients into expecting functionality that doesn't exist in v1.0.

The field can be reintroduced in a future version when a proper implementation is designed, including:
- Adding status_history field to Task message
- Adding include_status_history parameters to Get/List operations
- Defining behavior for history storage and retrieval

Fixes #1228

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -703,7 +703,7 @@ message DeviceCodeOAuthFlow {
 
 ///////////// Request Messages ///////////
 // --8<-- [start:SendMessageRequest]
-// Represents a request for the `message/send` method.
+// Represents a request for the `SendMessage` method.
 message SendMessageRequest {
   // Optional tenant, provided as a path parameter.
   string tenant = 4;
@@ -717,7 +717,7 @@ message SendMessageRequest {
 // --8<-- [end:SendMessageRequest]
 
 // --8<-- [start:GetTaskRequest]
-// Represents a request for the `tasks/get` method.
+// Represents a request for the `GetTask` method.
 message GetTaskRequest {
   // Optional tenant, provided as a path parameter.
   string tenant = 3;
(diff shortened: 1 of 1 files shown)


muscariello:
@amye @darrelmiller do we really need a vote for this one ? It looks like housekeeping.

amye:
I will advocate for +2 TSC and merge, this is housekeeping.

PR #1401, opened 2026-01-22
feat(spec): Provide ability for SDKs to be backwards compatible.

### LINKED ISSUE #1417: [Feat]: 1.0 Agents should be able to talk to 0.3 Agents and vice versa

Once 1.0 spec and SDKs are available, users should be able to upgrade without a risk of loosing ability to connect (both ways) to existing, non-updated Agents.

see #1416

see #1416

### Additional context

see #1416

### Code of Conduct

herczyn:
fixes #1417
for context see #1416

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,9 +6,13 @@
 
 *                       @a2aproject/a2a-tsc
 docs/                   @a2aproject/tech-writing
+scripts/                @a2aproject/tech-writing @a2aproject/google-a2a-eng
 specification/          @a2aproject/a2a-tsc
-types/                  @a2aproject/a2a-tsc
 docs/specification.md   @a2aproject/a2a-tsc
 GOVERNANCE.md           @a2aproject/a2a-tsc
+.mkdocs/                @a2aproject/tech-writing
 .mkdocs.yml             @a2aproject/tech-writing
-
+README.md               @a2aproject/tech-writing
+.gemini/                @a2aproject/tech-writing @a2aproject/google-a2a-eng
+.mkdocs/                @a2aproject/tech-writing
+.github/                @a2aproject/tech-writing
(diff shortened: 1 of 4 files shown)


specification/grpc/a2a.proto
```diff
@@ -474,6 +468,10 @@ message AgentProvider {
 // --8<-- [start:AgentCapabilities]
 // Defines optional capabilities supported by an agent.
 message AgentCapabilities {
+  // The security scheme details used for authenticating with this agent.
+  map<string, SecurityScheme> security_schemes = 6;
+  // Security requirements for contacting the agent.
+  repeated Security securities = 7;
```
darrelmiller: If we are going to change this name, I would recommend we take the opportunity to align the type name with [SecurityRequirement](https://spec.openapis.org/oas/v3.2.0.html#security-requirement-object) and use securityRequirements as the plural instead of securities.

herczyn: this actually helps us, because it means we don't have to move the field to another message if we don't want to. security_schemes can be handled in place because a one-of and union is easily distinguishable and does not break the unmarshalling in 0.3 (which Security does)

herczyn: moved the fields back with the name change, lmk if this is what you meant.

specification/grpc/a2a.proto
```diff
@@ -439,9 +438,8 @@ message AgentCard {
   AgentCapabilities capabilities = 7 [(google.api.field_behavior) = REQUIRED];
   // The security scheme details used for authenticating with this agent.
   map<string, SecurityScheme> security_schemes = 8;
-  // protolint:disable REPEATED_FIELD_NAMES_PLURALIZED
   // Security requirements for contacting the agent.
-  repeated Security security = 9;
+  repeated SecurityRequirement security_requirements = 19;
```
mikeas1, line 421: AgentSkill also needs this update.

herczyn: thanks, fixed!

herczyn:
> LGTM

i assume we still wait for the vote, right?

muscariello:
@ToddSegal  @geneknit @spetschulatSFDC @SivaNSAP @hughesthe1st

PR #1398, opened 2026-01-22
docs(spec): Correct wording around `blocking=true`

Tehsmash:
This corrects the wording so that blocking=True includes terminal and interrupted states as triggers for returning to the client.

Fixes: #1264

PR #1403, opened 2026-01-22
fix: Clarify blocking calls return on interrupted states

### LINKED ISSUE #1264: [Bug]: Handle blocking SendMessage calls needing user interactions
### What happened?

The SendMessageConfiguration of the spec [says](https://a2a-protocol.org/latest/specification/#322-sendmessageconfiguration) for `blocking` calls:

> The operation MUST wait until the task reaches a terminal state (completed, failed, cancelled, rejected) before returning. The response MUST include the final task state with all artifacts and status information.

However, I think we need some clarification that if the state is `INPUT_REQUIRED` or `AUTH_REQUIRED` that we need to return? Or if that is not the case, and explanation of how this is meant to work.

### Relevant log output

```shell

```

### Code of Conduct

madankumarpichamuthu:
Updates the specification to clarify that blocking SendMessage calls must return when tasks reach interrupted states (input_required, auth_required) in addition to terminal states. This prevents deadlock scenarios where tasks waiting for user input would cause blocking calls to hang indefinitely.

Changes:
- specification/grpc/a2a.proto: Updated blocking field comment to include "terminal or interrupted state"
- specification/grpc/a2a.proto: Fixed AUTH_REQUIRED classification to be an "interrupted state" (consistent with INPUT_REQUIRED)
- docs/specification.md: Updated blocking behavior description to include interrupted states and changed "final task state" to "current task state"

This enables proper client workflows where blocking calls can return on INPUT_REQUIRED or AUTH_REQUIRED, allowing clients to provide the needed information and resume with subsequent SendMessage calls.

Fixes #1264


muscariello:
@madankumarpichamuthu this is already covered by #1398

PR #1363, opened 2026-01-14
fix(spec): Refactor Part object to remove the need for FilePart and DataPart

darrelmiller:
- Removed FilePart and DataPart
- Moved all content properties into a oneOf
- Moved filename and meda_type into part so that they are optional properties for all content
- Changed data property to be value instead of struct to support all kinds of JSON structures.

PR #1411, opened 2026-01-25
fix: Simplify Part message structure by flattening FilePart and DataPart

darrelmiller:
This change simplifies the Part message design by:
- Renaming oneof 'part' to 'content' for clarity
- Flattening FilePart fields (file_with_uri -> url, file_with_bytes -> bytes) directly into Part
- Flattening DataPart by using google.protobuf.Value instead of Struct
- Moving media_type and filename to Part level (available for all content types)
- Removing separate FilePart and DataPart messages

Also updated migration guidance documentation to reference ancillary documents instead of requiring an in-spec appendix.

Related to issue #1255

Replaces PR #1363

### DIFF
--- a/specification/grpc/a2a.proto
+++ b/specification/grpc/a2a.proto
@@ -353,7 +353,7 @@ message PushNotificationConfig {
   string url = 2 [(google.api.field_behavior) = REQUIRED];
   // Token unique for this task/session
   string token = 3;
-  // Information about the authentication to sent with the notification
+  // Authentication information required to send the notification
   AuthenticationInfo authentication = 4;
 }
 // --8<-- [end:PushNotificationConfig]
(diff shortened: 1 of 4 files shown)


specification/grpc/a2a.proto
```diff
@@ -226,45 +226,25 @@ message TaskStatus {
...
-
-// --8<-- [start:FilePart]
-// FilePart represents the different ways files can be provided. If files are
-// small, directly feeding the bytes is supported via file_with_bytes. If the
-// file is large, the agent should read the content as appropriate directly
-// from the file_with_uri source.
-message FilePart {
-  oneof file {
+    // The raw byte content of a file. In JSON serialization, this is encoded as a base64 string.
+    bytes bytes = 2;
```
darrelmiller:
```suggestion
    bytes raw = 2;
```

PR #1427, opened 2026-01-29
docs(spec): Convert Proto Table display in specification to use existing libraries

holtskinner:
Convert bespoke `macros.py` to use `proto-schema-parser` and `tabulate` to display tables


darrelmiller:
This looks very promising. It's nice that we can remove all the extra magic comments in the proto file.

Does this also handle things like putting the note below the table for oneOfs?  There were a bunch of exception cases that I had to handle.  

Is there any chance you could share the rendered HTML so we can see how it renders the tables?

holtskinner:
[site.zip](https://github.com/user-attachments/files/24968943/site.zip)
Here is the generated site

holtskinner:
@gemini-code-assist review

holtskinner:
> This looks very promising. It's nice that we can remove all the extra magic comments in the proto file.
> 
> Does this also handle things like putting the note below the table for oneOfs? There were a bunch of exception cases that I had to handle.
> 
> Is there any chance you could share the rendered HTML so we can see how it renders the tables?

I've also gone through and checked the tables. As far as I can tell, all of the values match and are displayed as expected, including the `oneOf` sections.

holtskinner:
@gemini-code-assist review

holtskinner:
@darrelmiller This is ready for final review

darrelmiller:
@holtskinner  LGTM. Thank you!

PR #1436, opened 2026-01-30
docs(spec): What changed from v0.3.0 to v1.0

darrelmiller:

PR #1460, opened 2026-02-09
docs(spec): Fix whats-new-v1.md

ishymko:
Changes below as they appear in the file.

- `blocking` flag existed in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#712-messagesendconfiguration-object)
- `SendMessage`: remove "MAY return Message"/"MUST return Task" which are not in the spec, remove service parameters mention as they are covered above
- streaming was specified for gRPC in proto definition as `returns (stream StreamResponse)`
- `SubscribeToTask` is not a new operation
- blocking has no effect for streaming operations: [source](https://a2a-protocol.org/latest/specification/#322-sendmessageconfiguration)
- `TaskNotCancelableError` was mentioned in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#82-a2a-specific-errors)
- RFC 7515 was mentioned for Agent Card signatures in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#556-agentcardsignature-object), kept only RFC 8785 for canonicalization which wasn't
- `agentId` field does not exist and never did
- extension declarations (AgentExtension object) did not change between [0.3](https://a2a-protocol.org/v0.3.0/specification/#5521-agentextension-object) and [1.0](https://a2a-protocol.org/latest/specification/#444-agentextension)
- Mutual TLS existed in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#553-securityscheme-object)
- subscribing to tasks not originally created via streaming is not a behavior change and is not explicitly mentioned in 1.0
- `token` is not removed from `PushNotificationConfig`: [source](https://a2a-protocol.org/latest/specification/#431-pushnotificationconfig)
- `configId` does not exist in `PushNotificationConfig` ([source](https://a2a-protocol.org/latest/specification/#431-pushnotificationconfig)) and `id` was there in 0.3 ([source](https://a2a-protocol.org/v0.3.0/specification/#68-pushnotificationconfig-object))
- `PushNotificationNotSupportedError` existed in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#82-a2a-specific-errors)
- `SendMessageRequest` example fixed to match the spec
- `createdAt` and `lastModified` do not exist in `Task`: [source](https://a2a-protocol.org/latest/specification/#411-task)
- changes to nested `Task` objects removed as they are covered in respective sections (`TaskStatus`, `Message`, `Artifact`)
- `timestamp` was not added to `Message`: [source](https://a2a-protocol.org/latest/specification/#414-message)
- there is no "stricter typing" for `metadata`
- "retained" should not be in the changelog
- `createdAt` was not added to `Artifact`: [source](https://a2a-protocol.org/latest/specification/#417-artifact)
- `signatures` field was there in 0.3: [source](https://a2a-protocol.org/v0.3.0/specification/#55-agentcard-object-structure)

Re #1436.


docs/whats-new-v1.md
```diff
@@ -559,20 +464,6 @@ This capability may be reintroduced in a future version with proper implementati
...
 
 - ✅ `extendedAgentCard`: Moved from top-level `supportsAuthenticatedExtendedCard` field
-- ✅ `pushNotifications`: More formally specified with enhanced configuration options
-- ✅ `streaming`: Enhanced with clearer semantics for SSE support
-
-### SecurityScheme Objects
-
-**Added:**
-
-- ✅ `MutualTLSSecurityScheme`: Support for mutual TLS authentication
```
darrelmiller, line 569: Not sure what happened here.  It is in the spec https://a2a-protocol.org/latest/specification/#456-mutualtlssecurityscheme but not in the proto

darrelmiller: Also, deviceCodeFlow was added and implicit was supposed to be removed but it still shows in the proto https://a2a-protocol.org/latest/specification/#457-oauthflows

ishymko: > MutualTLSSecurityScheme

Right, good catch, proto embedding is broken on the website now for MTLS, I believe it may be because of case sensitivity. But it's a separate problem, either way MTLS [existed in 0.3](https://github.com/a2aproject/A2A/blob/v0.3.0/specification/grpc/a2a.proto#L508).

> deviceCodeFlow was added and implicit was supposed to be removed

Implicit and password flows were [marked as deprecated](https://github.com/a2aproject/A2A/blob/741ccaa3af19fbe6f9ba26a47d8781fe68af475f/specification/a2a.proto#L632-L633) in #1401.

ishymko: MTLS fix: #1462

holtskinner:
While we're at it, should we add an entry in `mkdocs.yml` to add this page to the TOC?

ishymko:
Apologize @holtskinner auto-merge merged the PR with your comment. I see you covered it in #1489, thank you!

PR #1470, opened 2026-02-12
fix(spec): Adjust field number for `ListTasksRequest.tenant` to prevent missing number

holtskinner:
Introduced in https://github.com/a2aproject/A2A/pull/1195

- Breaking change should be fine because `ListTasksRequest` is new for 1.0.0

Release-As: 1.0.0

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -175,6 +175,7 @@ The primary operation for initiating agent interactions. Clients send a message 
 
 - [`ContentTypeNotSupportedError`](#332-error-handling): A Media Type provided in the request's message parts is not supported by the agent.
 - [`UnsupportedOperationError`](#332-error-handling): Messages sent to Tasks that are in a terminal state (e.g., completed, canceled, rejected) cannot accept further messages.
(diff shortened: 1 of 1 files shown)


specification/a2a.proto
```diff
@@ -673,15 +673,15 @@ message GetTaskRequest {
 // Parameters for listing tasks with optional filtering criteria.
 message ListTasksRequest {
   // Optional tenant, provided as a path parameter.
-  string tenant = 9;
+  string tenant = 8 [(google.api.field_behavior) = OPTIONAL];
```
darrelmiller: It is not clear to me why we are marking this as OPTIONAL.  This is a request parameter.  It does not really have any persistent storage, nor any default value.  So, I'm not sure why we need to distinguish between "unset" and "set to default", which is what I understand OPTIONAL to do.

holtskinner: We don't have to put the OPTIONAL field behavior. I can remove that part.

holtskinner:
```suggestion
  string tenant = 8;
```

PR #1474, opened 2026-02-13
fix(spec)!: Add LF prefix to the package.

### LINKED ISSUE #1473: [Bug]: V1 Spec has the same package as 0.3
### What happened?

This will make it very hard for any usecase where both need to coexist - compatibility layer, migrations etc.

### Relevant log output

```shell

```

### Code of Conduct

herczyn:
Fixes #1473

Release-As: 1.0.0


holtskinner:
@herczyn Does this need to be considered a breaking change?

amye:
Adding to TSC agenda for 17 Feb.

herczyn:
> @herczyn Does this need to be considered a breaking change?

good question, but i think yes. it impacts how classes are generate and how grpc connections are made. for http/json rpc it's not breaking on the wire.

Issue #1451, opened 2026-02-05
[Bug]: 1.0-rc inconsistency corrections


yarolegovich:
### What happened?

1. TaskNotFoundError is missing form non-streaming message [errors list](https://a2a-protocol.org/latest/specification/#311-send-message) (addressed in #1482).

2. Capability validation typo in [extensions section](https://a2a-protocol.org/latest/specification/#334-capability-validation), has to be server (addressed in #1482):
> When a client requests use of an extension marked as required: true in the Agent Card but **the client does** not declare support for it, the agent MUST return ExtensionSupportRequiredError.

3. Migration and legacy [compatibility examples]([Legacy](https://a2a-protocol.org/latest/specification/#appendix-a-migration-legacy-compatibility)) are wrong / outdated (addressed in #1482):
```json
{
  "file": {
    "mediaType": "image/png",
    "name": "diagram.png",
    "fileWithBytes": "iVBORw0KGgo..."
  }
}
```
must be:
```json
{
  "mediaType": "image/png",
   "filename": "diagram.png",
   "raw": "iVBORw0KGgo..."
}
```

4. A mismatch of ID naming in Create vs (Delete|List|Get) RPCs operating on push configurations. The field is called ConfigID in CreateRequest, but just ID in the rest. In CreateRequest the field is actually duplicated because we have ConfigID on the root level in addition to Config.ID in the actual config.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nahapetyan-serob:
https://a2a-protocol.org/latest/specification/#62-streaming-task-execution
Here we also have "final" in SSE Response Stream statusUpdate example, which was removed.

yarolegovich:
Should ListTaskPushNotificationConfig be pluralized?

yarolegovich:
addressed in #1482 #1486 #1487

PR #1487, opened 2026-02-17
fix(spec)!: remove duplicated ID from the create task push config request

yarolegovich:
ID is already present in the config object.
It is not used in path segment bindings and also uses inconsistent naming with other push-related rpc payloads.

fixes https://github.com/a2aproject/A2A/issues/1451

Follow-up to #1389

### DIFF
--- a/specification/a2a.proto
+++ b/specification/a2a.proto
@@ -107,8 +107,8 @@ service A2AService {
     };
     option (google.api.method_signature) = "task_id,id";
   }
-  // Lists push notification configs for a task.
-  rpc ListTaskPushNotificationConfig(ListTaskPushNotificationConfigRequest) returns (ListTaskPushNotificationConfigResponse) {
+  // Get a list of push notifications configured for a task.
+  rpc ListTaskPushNotificationConfigs(ListTaskPushNotificationConfigsRequest) returns (ListTaskPushNotificationConfigsResponse) {
     option (google.api.http) = {
       get: "/tasks/{task_id=*}/pushNotificationConfigs"
       additional_bindings: {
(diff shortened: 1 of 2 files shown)


holtskinner:
Good catch.

Just to clarify, this doesn't need to be marked as a breaking change because this should have been added in https://github.com/a2aproject/A2A/pull/1389 (already marked as breaking)

specification/a2a.proto
holtskinner, line 743:
```suggestion
  reserved 2;
  reserved 3;
```

holtskinner: Should we add this since this field is no longer used?

Or since there's already breaking changes, should we go ahead and shift the numbers up?

specification/a2a.proto
```diff
@@ -739,8 +739,6 @@ message CreateTaskPushNotificationConfigRequest {
   string tenant = 4;
   // The parent task resource id.
   string task_id = 1 [(google.api.field_behavior) = REQUIRED];
-  // The ID for the new config.
-  string config_id = 2 [(google.api.field_behavior) = REQUIRED];
```
darrelmiller, line 751: Remove id from line 481 also.   TaskPushNotificationConfig contains an id for the config that is redundant with the id in PushNotificationConfig.

specification/a2a.proto
herczyn: i will do a separate cl for that

specification/a2a.proto
```diff
@@ -477,8 +477,6 @@ message AgentCardSignature {
 message TaskPushNotificationConfig {
   // Optional. Tenant ID.
   string tenant = 4;
-  // The ID of the configuration.
```
holtskinner, line 480: Should we mark field as reserved or just adjust numbers later?

PR #1500, opened 2026-02-18
refactor(spec)!: Combine `TaskPushNotificationConfig` and `PushNotificationConfig`

herczyn:
**this PR:**
- removes CreateTaskPushNotificationConfigRequest and replace it with TaskPushNotificationConfig (same fields).
- flattend TaskPushNotificationConfig and PushNotificationConfig removing the latter
- puts TaskPushNotificationConfig in SendMessageConfig instead of the removed PushNotificationConfig

**why:** more readable structure, with less messages but with the same possibilities

Release-As: 1.0.0


Tehsmash (commented):
LGTM, but its hitting an AIP linter issue. We do have the correct comment in place for ignoring AIP-133 but it doesn't seem to be applying it properly maybe because the extra comment line inbetween?

holtskinner:
@gemini-code-assist review

amye:
@a2aproject/a2a-tsc - ping for review

specification/a2a.proto
```diff
@@ -472,10 +462,17 @@ message AgentCardSignature {
...
   // Optional. Tenant ID.
   string tenant = 1;
-  // The ID of the task this configuration is associated with.
-  string task_id = 2 [(google.api.field_behavior) = REQUIRED];
   // The push notification configuration details.
-  PushNotificationConfig push_notification_config = 3 [(google.api.field_behavior) = REQUIRED];
+  // A unique identifier (e.g. UUID) for this push notification configuration.
+  string id = 2;
+  // The ID of the task this configuration is associated with.
+  string task_id = 3 [(google.api.field_behavior) = REQUIRED];
```
darrelmiller: How can this be marked as required, when it is explicitly not required when using it with send message?  And what is expected when you call send message with a taskId and you provide the a TaskPushNotificationConfig?

darrelmiller:
I'm for this change in principle, I just want to get clarity on the issue around the "required" for taskId

specification/a2a.proto
herczyn: good point, removed required. i think if the task id is set for send message we should throw an error.

PR #1507, opened 2026-02-23
refactor(spec): rename `SendMessageConfiguration.blocking` to `SendMessageConfiguration.return_immediately `

### LINKED ISSUE #1453: [Feat]: Rename blocking to make the default behavior blocking

The spec now explicitly specifies the `blocking` flag in `SendMessageConfiguration` defaults to false:
>  If true, the operation waits until the task reaches a terminal or interrupted state before returning. Default is false.

This was not specified before and many SDKs (python, Go, JS, haven't checked the other ones) implemented blocking=true default behavior for the case when client have not explicitly provided this flag. 

My concern is that:

> Non-Blocking (blocking: false): The operation MUST return immediately after creating the task, even if processing is still in progress

Will be counterintuitive for clients which do not pass any value explicitly expecting a completed task a result.

Rename `blocking` in [SendMessageConfiguration](https://a2a-protocol.org/latest/specification/#322-sendmessageconfiguration) to `polling`, `nonblocking` or `async` such that the `false` default gives a more predictable default behavior. 

_No response_

### Additional context

_No response_

### Code of Conduct

ishymko:
As mentioned in the linked issue, defaulting to non-blocking behavior is counterintuitive and is implemented the opposite way in the SDKs.

Renaming it to `return_immediately` makes default behavior simpler while advanced scenario requires an extra flag.

---
Previos proposals:
1. `polling` has inaccurate meaning in this context as pointed out in the comments.
2. `non_blocking` is a negative boolean which leads to confusing conditional expressions (`not non_blocking`).

Fixes #1453

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -443,7 +443,7 @@ This section defines common parameter objects used across multiple operations.
 
 **Execution Mode:**
 
-The `return_immediately` field in [`SendMessageConfiguration`](#322-sendmessageconfiguration) controls whether the operation returns immediately or waits for task completion. By default, operations are blocking:
+The `return_immediately` field in [`SendMessageConfiguration`](#322-sendmessageconfiguration) controls whether the operation returns immediately or waits for task completion. Operations are blocking by default:
 
 - **Blocking (`return_immediately: false` or unset)**: The operation MUST wait until the task reaches a terminal state (`COMPLETED`, `FAILED`, `CANCELED`, `REJECTED`) or an interrupted state (`INPUT_REQUIRED`, `AUTH_REQUIRED`) before returning. The response MUST include the latest task state with all artifacts and status information. This is the default behavior.
 


Tehsmash:
Hey @ishymko, Thanks for the PR, I'm not sure about the wording of "polling" when its a server side flag. i.e. Now it sounds like your asking the server to poll when in fact you're asking it to block/not-block. Polling is a client side operation and is just one reason to send blocking=False for example if you register a webhook callback you aren't doing polling but you still want non-blocking.

Even if blocking=True you may still need to "poll" the Task to get it to completion if it goes through an interrupted state, so blocking=True is really an implementation of long polling which returns when the state changes. Perhaps we should lean further into that design.

ishymko:
Hey @Tehsmash, thank you for your input, makes sense, what's your proposal for something which is going to be `false` by default?

Options mentioned in #1453: `polling`, `nonblocking` or `async`.

Tehsmash:
Out of the proposed options I prefer `async`.
Other options could be: `nowait` or `return_immediately`.

Another option that we make `blocking` an `optional boolean` in the proto, this would allow us to have three values: unset, true and false. Then we can decide what the "unset" behaviour is separately.

darrelmiller:
I added a comment on the issue, but adding here too.  I agree with @Tehsmash that "polling" is not the right term.  I'd rather not use "async" because it means something different in programming languages.  I'm ok with "nonBlocking", or "noWait" or "returnImmediately".  I think the last suggestion is probably the most explicit description of the desired behavior.

ishymko:
@Tehsmash @darrelmiller thank you for your input. Renamed to `non_blocking` as it's close to original. I believe that `return_immediately` still doesn't eliminate the need of reading the description.

~On a side note: CI is failing due to some unrelated (I hope) issue.~ Fixed via #1509

ishymko:
CI is going to be fixed via #1509.

ishymko:
@holtskinner so how about `return_immediately`?

Changing it to proto `optional` and treating unset as `true` is also confusing IMO.

ishymko:
@holtskinner @darrelmiller renamed to `return_immediately`

chorghemaruti64-creator:
`return_immediately` is much clearer — aligns with how SDKs typically name this (cf. HTTP's long-polling vs immediate-return distinction).

One addition worth considering: **discovery-time declaration**. If an agent's Agent Card could advertise its default execution model (blocking/streaming/async), callers could make the `return_immediately` decision based on peer capabilities rather than guessing.

Example: `preferred_execution: ["streaming", "async"]` in Agent Card capabilities. Then `return_immediately` at call time becomes a negotiated behavior, not a bilateral assumption. Reduces cases where caller defaults to blocking against an agent that's optimized for async.

chorghemaruti64-creator:
The naming matters beyond runtime behavior — it has implications at discovery time too.

When an agent publishes its AgentCard, the execution model it advertises becomes a filter for orchestrators *before* they attempt a connection. An orchestrator coordinating a time-sensitive pipeline needs to know: "does this agent return immediately, or does it hold the connection?" This decision should happen at agent selection, not after the task handoff fails.

That's why I'd lean toward `return_immediately` over `async` or `nonBlocking`: it's self-describing from the caller's perspective. `async` is ambiguous (all HTTP calls are async at some level), and `nonBlocking` is a double negative in boolean terms as @darrelmiller noted.

A field named `return_immediately: bool` in the AgentCard makes it unambiguous both at discovery time ("I need an agent that returns immediately") and at runtime ("should I wait or stream?"). The semantics are caller-centric, which is the right frame for a protocol field.

One related thought: this field probably deserves a default value stated explicitly in the spec, since discovery layers need to make assumptions when the field is absent. If agents don't declare it, orchestrators have no way to pre-filter — they fail at execution instead of at discovery.

muscariello (approved):
LGTM

holtskinner:
@gemini-code-assist review

chorghemaruti64-creator:
Worth noting: the `return_immediately` framing does something useful beyond just readability — it signals execution intent at the *call site*, which matters for discovery filtering.

An orchestrator selecting an agent for a workflow needs to know before task assignment whether the agent will block the calling context. If that information lives in the field name (`return_immediately: false` = "I will block until done"), a discovery layer can surface it as a capability filter — "find agents that support long-running non-blocking tasks" vs "find agents optimized for synchronous interactions."

`blocking: true/false` is ambiguous about *whose* blocking perspective it represents (caller? agent? connection?). `return_immediately` is unambiguous: it describes what happens to the caller's context.

One edge case to consider: in multi-agent pipelines, an orchestrator might want `return_immediately: false` for agents near the end of the chain (where it can afford to wait) and `return_immediately: true` for intermediate agents (where it needs to continue coordinating). The field-level clarity here actually helps pipeline composition, not just individual invocations.

The rename is the right call.

ishymko:
@chorghemaruti64-creator - it's clear at this point that comments are AI-generated and are added on any change in the PR even when content is not affected (no-op merge from the main branch, approval, etc.), please stop.

amye:
>  - it's clear at this point that comments are AI-generated and are added on any change in the PR even when content is not affected (no-op merge from the main branch, approval, etc.), please stop.

Blocked. Carry on.

PR #1585, opened 2026-03-04
docs(spec): clarify contextId optional generation and client side semantics

### LINKED ISSUE #1581: [Bug]: Context ID Client Side Generation / Optional Semantics
### What happened?

Thread tracking follow-up to https://github.com/a2aproject/A2A/issues/1406
- Can contextId be optional?
- Can contextId be generated by the client?

vinoo999:
Clients MAY provide a contextId in a Message to associate it with an existing context. Server MUST generate contextId when creating a Task but MAY omit it when responding with a Message only (no task created).

Keeps Task.context_id as REQUIRED in [a2a.proto](https://github.com/a2aproject/A2A/blob/d1ed0da587d2d634ba0b81a40d082cee0850b81b/specification/a2a.proto) — no breaking change.

Fixes #1581

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -624,10 +624,11 @@ The A2A protocol supports several patterns for multi-turn interactions. Agents t
 **Context Continuity:**
 
 - [`Task`](#411-task) objects maintain conversation context through the `contextId` field
-- Clients **MAY** include a `contextId` in any [`Message`](#414-message) to associate it with an existing context or to indicate continuation of a previous interaction
+- Clients **MAY** include the `contextId` in subsequent messages to indicate continuation of a previous interaction
 - Clients **MAY** use `taskId` (with or without `contextId`) to continue or refine a specific task
-- Clients **MAY** use `contextId` without `taskId` to start a new task within an existing conversation context
+- Clients **MAY** use `contextId` without `taskId` to start a new task within an existing or non-existing conversation context provided validations pass
 - Agents **MUST** infer `contextId` from the task if only `taskId` is provided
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -589,11 +589,14 @@ A `contextId` is an identifier that logically groups multiple related [`Task`](#
 
 **Generation and Assignment:**
 
-- Agents **MUST** generate a new `contextId` when processing a [`Message`](#414-message) that does not include a `contextId` field
-- The generated `contextId` **MUST** be included in the response (either [`Task`](#411-task) or [`Message`](#414-message))
-- Agents **MUST** accept and preserve client-provided `contextId` values if validations pass (i.e., it doesn't conflict with provided `taskId`)
+- When a client includes a `contextId` in a [`Message`](#414-message), it **MAY** reference an existing context on the server
```
darrelmiller, line 592: If a context is not recognized by the server, what is expected to happen? A new one created and propagated? A error message? Ignored?

vinoo999: Agents **MUST** provide a Validation Error if a context id cannot be accepted?

darrelmiller: If an agent doesn't care about contextIds, because it isn't going to use it, do you consider that as "cannot be accepted".  Or are you ok with an agent choosing to ignore the contextId and just pass it back blindly in the task?

vinoo999: If the agent doesn't care about context ids I would still expect the client context id to be preserved per spec and returned accordingly.

vinoo999:
Closing in favor of https://github.com/a2aproject/A2A/pull/1588

PR #1586, opened 2026-03-04
docs(spec)!: make contextId fully optional on server side

### LINKED ISSUE #1581: [Bug]: Context ID Client Side Generation / Optional Semantics

vinoo999:
BREAKING CHANGE: Task.context_id is now OPTIONAL in a2a.proto (previously [REQUIRED](https://github.com/a2aproject/A2A/blob/d1ed0da587d2d634ba0b81a40d082cee0850b81b/specification/a2a.proto)). Agents MAY generate a new contextId and include it in the response, but are no longer required to do so.

Clients MAY provide a contextId in a Message. Agents MUST accept and preserve client-provided contextId values if validations pass as before.

This aligns more concretely with TSC direction in #1581 that contextId should be optional as the agent need not provide a tying identifier.

Fixes #1581

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -623,10 +623,11 @@ The A2A protocol supports several patterns for multi-turn interactions. Agents t
 **Context Continuity:**
 
 - [`Task`](#411-task) objects **MAY** maintain conversation context through the optional `contextId` field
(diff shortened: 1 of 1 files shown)


vinoo999:
Closing in favor of https://github.com/a2aproject/A2A/pull/1588

PR #1587, opened 2026-03-04
fix(spec): Remove config from binding.

herczyn:
Fix error caused by #1500:

× Failed to build `a2a-sdk @ file:///home/runner/work/a2a-python/a2a-python`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `hatchling.build.build_editable` failed (exit status: 1)

      [stderr]
      Failure: failed_precondition: plugin
      "buf.build/grpc-ecosystem/openapiv2:v2.28.0" failed: no field "config"
      found in TaskPushNotificationConfig


herczyn:
the bug causing issues in python import, and it's confirmed that the change fixes it.

PR #1588, opened 2026-03-04
fix(docs): Clearer wording around context id.

herczyn:
client side context id passing can be supported, server can choose to reject, client's should know how to fallback accordingly
if server accepts, server must respond with the context id
if server cannot accept client side context id it MUST respond with an error (i.e. it SHOULD not respond with a task with a different context id?)
servers may not accept a client side context id for a number of reasons: clashes, validation errors, not supporting context ids

### DIFF
--- a/specification/a2a.proto
+++ b/specification/a2a.proto
@@ -168,7 +168,7 @@ message Task {
   string id = 1 [(google.api.field_behavior) = REQUIRED];
   // Unique identifier (e.g. UUID) for the contextual collection of interactions
   // (tasks and messages).
-  string context_id = 2 [(google.api.field_behavior) = OPTIONAL];
+  string context_id = 2;
   // The current status of a `Task`, including `state` and a `message`.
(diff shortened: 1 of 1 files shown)


specification/a2a.proto
```diff
@@ -167,8 +167,8 @@ message Task {
   // new task.
   string id = 1 [(google.api.field_behavior) = REQUIRED];
   // Unique identifier (e.g. UUID) for the contextual collection of interactions
-  // (tasks and messages). Created by the A2A server.
-  string context_id = 2 [(google.api.field_behavior) = REQUIRED];
+  // (tasks and messages).
+  string context_id = 2 [(google.api.field_behavior) = OPTIONAL];
```
darrelmiller: Why are we putting the `OPTIONAL` attribute here? We don't use it anywhere else in the proto. Why is this special?

herczyn: good point, removed

PR #1600, opened 2026-03-06
refactor(spec): Update HTTP+JSON error mapping to use `google.rpc.Status` AIP-193

### LINKED ISSUE #1596: [Bug]: Error format for HTTP+JSON Binding causes friction with gRPC transcoding
### What happened?

The [specification states](https://a2a-protocol.org/latest/specification/#116-error-handling) that errors in the HTTP+JSON binding should use RFC 9457 Problem Details format for error response bodies.

The [specification for the gRPC binding](https://a2a-protocol.org/latest/specification/#106-error-handling) states to use google.rpc.Status for representing errors. The use of google.rpc.Status for errors is the [official model for gRPC](https://grpc.io/docs/guides/error/).

The difference between these two error formats makes it difficult to expose an HTTP+JSON endpoint that is based on gRPC HTTP transcoding. Ensuring the HTTP+JSON specification is compatible with gRPC HTTP transcoding has been an explicit goal.

Developers that wish to write their agent in gRPC but expose both a gRPC endpoint and an HTTP+JSON endpoint via gRPC transcoding are left with a difficult problem to solve. gRPC transcoding markup does not support custom error payloads, and so this is not an off-the-shelf feature of transcoders.

I am aware of 4 gRPC transcoding systems:

- [Envoy](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_json_transcoder_filter)
- [gRPC Gateway](https://grpc-ecosystem.github.io/grpc-gateway/)
- [Cloud Endpoints](https://docs.cloud.google.com/endpoints/docs/grpc/transcoding)
- [ASP.NET Core gRPC transcoding](https://learn.microsoft.com/en-us/aspnet/core/grpc/json-transcoding?view=aspnetcore-10.0)

None of these systems offer custom error payload transcoding.

This means that it is not possible to adhere to the HTTP+JSON error specification with standard gRPC transcoding. Developers would need to write and maintain their own custom transcoding layer.

If we instead align the HTTP+JSON error output with the transcoded format of google.rpc.Status, this becomes much simpler. 

### Relevant log output

```shell

```

### Code of Conduct

mikeas1:
Updates the specification to require use of `google.rpc.Status` formatted JSON objects for error response bodies in HTTP+JSON mapping.

See related issue for reasoning.

Fixes #1596

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -51,7 +51,7 @@ jobs:
           python-version: 3.13
 
       - name: Setup uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v7
 
       - name: Restore pip cache
         uses: actions/cache@v5


darrelmiller:
I will add a note to align the data property of JSONRPC errors with the details property of status code.

Need to confirm with @mikeas1  that removing the outer "error" member that AIP uses will not break transcoding.

darrelmiller:
@mikeas1  I made a PR into your PR https://github.com/mikeas1/A2A/pull/2

mikeas1:
> Need to confirm with @mikeas1  that removing the outer "error" member that AIP uses will not break transcoding.

I will double check this today.

mikeas1:
I did a quick review of gRPC Gateway, Envoy, and .NET transcoding. All of these seem to directly transcode the google.rpc.Status value into the HTTP response body. gRPC Gateway supports registering a custom error transcoder, but I did not find an equivalent for Envoy. 

Google's transcoding system adheres to AIP-193, which puts it at odds with the others. 

I'm torn. Google is almost certainly going to be the primary user of gRPC transcoding, so aligning errors to match how that works is very useful. But this will make it harder for anyone else using transcoding to be adherent to the spec. However, it should be possible to be adherent, and with additional SDK support we can paper over this gap. So, I'm inclined to align the spec with AIP-193.

PR #1605, opened 2026-03-09
docs: Create RELEASES.md

amye:
Draft release checklist that may be way too heavy!


amye:
> Some of the parts in the process are automated with release-please already, so we might not need to spell them out directly, but overall good steps.

I knew they were already here just generally, but I'm documenting nearly everything that I can think of! For example, every time we push a release, we should update /docs/roadmap.md to help keep things clean and give ourselves a good boundary to make sure we're on top of our communications. 

I'll hold off committing this until we're through the release process for 1.0 and we can update subsequently from here.

PR #1549, opened 2026-02-26
feat!: Proposal for bidirectional streaming over gRPC

HungYangChang:
Resolves #656

# Follow on PR #1120 - BiDi Streaming Discussion Responses

**PR**: [feat!: Proposal for bidirectional streaming over gRPC](https://github.com/a2aproject/A2A/pull/1120)
**Original Author**: mikeas1| **Author**: hungyangchang| **Status**: Open, `TSC Review`

## Summary

Adds `SendLiveMessage` bidirectional streaming RPC to A2A gRPC. Clients maintain an active stream with agents instead of separate follow-up requests. Benefits: inline await, less client state tracking, simpler distributed routing.

## Upstream File Structure Changes

The original PR modified 3 files that no longer exist in upstream `main`:

| File | Old (PR #1120) | Current upstream | Action |
|------|---------------|-----------------|--------|
| `specification/grpc/a2a.proto` | Modified | Moved to `specification/a2a.proto` | Changes applied to new location |
| `specification/json/a2a.json` | Modified | Auto-generated from proto, not in git | None needed |
| `types/src/types.ts` | Modified | Removed entirely | None needed |

`a2a.proto` is now the **single source of truth**. The JSON schema is generated at build time via `scripts/proto_to_json_schema.sh`, and `docs/specification.md` uses `proto_to_table()` macros that auto-render from proto definitions.

---

### 1. gemini-code-assist (bot) - Oct 1, 2025

**Field ordering**: Fixed. `bidi_streaming` now follows `extensions` by field number. In new schema it's field 5 (field 4 = `extended_agent_card`).

**`FAILED_PRECONDITION` vs `UNIMPLEMENTED`**: Both are used correctly for distinct cases. `FAILED_PRECONDITION` = duplicate active stream; `UNIMPLEMENTED` = method not supported.

**Omitted `SendMessageConfiguration`**: Already specified: "If omitted, the agent MUST continue to use the configuration from the initial request."

### 2. ToddSegal - Oct 7, 2025

**Resubscribe interaction**: Continues as-is. `SendLiveMessage` is a separate RPC.

**Multiple client calls**: Three rules (per mikeas1):

1. No multi-connection support -> `FAILED_PRECONDITION`
2. Multi-connection support -> treat as resubscribe with multiplexed input
3. Agent has full authority over task lifecycle

**Missing task_id**: Empty values are interpreted as the established stream's task_id/context_id.

### 3. HungYangChang - Dec 11, 2025

[...]

PR #1619, opened 2026-03-12
docs: Add custom protocol bindings documentation

Tehsmash:
## Summary

- Adds `docs/topics/custom-protocol-bindings.md`, a dedicated page documenting custom protocol bindings as a concept distinct from extensions; covers declaration in the Agent Card, specification requirements, key areas to address (data type mappings, service parameters, error mapping, streaming, authentication), and interoperability testing
- Adds `docs/topics/extension-and-binding-governance.md`, merging the governance content previously in `extensions.md` with the equivalent custom protocol binding governance; deduplicates the tiers section into a single section with a naming table; notes that SDKs SHOULD implement official custom protocol bindings vs MAY for extensions
- Adds spec section 5.8 defining URI-based custom binding identification as a SHOULD requirement and updates the section 12.7 Agent Card example accordingly
- Replaces the verbose inline governance sections in `extensions.md` and `custom-protocol-bindings.md` with short summaries linking to the new governance page
- Adds Custom Protocol Bindings and Extension & Binding Governance to the mkdocs.yml nav

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1297,7 +1297,7 @@ The `protocolBinding` field in the Agent Card's `supportedInterfaces` entry
 When a breaking change is introduced to a binding, a new URI **MUST** be used
 so that clients can distinguish between incompatible versions:
 
-```
+```text
 https://example.com/bindings/websocket/v1  →  https://example.com/bindings/websocket/v2
 ```
 
--- a/docs/topics/extension-and-binding-governance.md
+++ b/docs/topics/extension-and-binding-governance.md
@@ -14,11 +14,11 @@ those hosted under the `a2aproject` GitHub organization.
 Both extensions and custom protocol bindings use a two-tier system within the
 `a2aproject` organization. Repository naming and URI prefixes differ by type:
 
-|                        | Extensions                                      | Custom Protocol Bindings                       |
-| :--------------------- | :---------------------------------------------- | :--------------------------------------------- |
-| Official repo prefix   | `ext-{name}`                                    | `cpb-{name}`                                   |
-| Experimental repo prefix | `experimental-ext-{name}`                     | `experimental-cpb-{name}`                      |
-| Official URI prefix    | `https://a2a-protocol.org/extensions/`          | `https://a2a-protocol.org/bindings/`           |
+|                          | Extensions                                      | Custom Protocol Bindings                       |
+| :----------------------- | :---------------------------------------------- | :--------------------------------------------- |
+| Official repo prefix     | `ext-{name}`                                    | `cpb-{name}`                                   |
+| Experimental repo prefix | `experimental-ext-{name}`                       | `experimental-cpb-{name}`                      |
+| Official URI prefix      | `https://a2a-protocol.org/extensions/`          | `https://a2a-protocol.org/bindings/`           |
 
 ### Official
 


holtskinner:
@gemini-code-assist review

Tehsmash:
Hey @vinoo999 thanks for reviewing. 

The way I see it protocol bindings implement the A2A protocol over top of another protocol, defining how A2A is bound to that protocol. i.e. how A2A can be carried over HTTP + JSON, or how A2A can be carried over gRPC.

Extensions are extensions to the A2A protocol specification, either with additional data or even operations/methods.

If an Extension is simply an extension of a existing data models and is purely behavioural, the example extension in the spec adds geolocation metadata for example. Then this can be easily supported by existing protocol bindings without additional work.

If an Extension adds additional methods / operations to the protocol, then I believe it is up to the protocol binding author to choose to support that extension or not, it also may not be possible for all protocol bindings to support all operations. BiDi streaming is a great example of this.

To put together a concrete example I can use the BiDi streaming proposal we already have a PR for (https://github.com/a2aproject/A2A/pull/1549): 

The BiDi protocol "extension" defines some new methods/operations:
	* SendMessageLive, and ReconnectToTask as well as the data models etc associated with them.
	* The expected behaviour when these methods are called
	* Today the PR adds a new capability in the AgentCard but in an extension world this could be exposed as an available extension

There is also a part of the PR which adds a new section to the gRPC protocol binding to declare it supports these methods. This would not be part of the protocol extension, but instead would be the protocol binding "supporting" the extension.

If I have a custom Websocket protocol binding outside of the spec. Initially there would be not be support for the new BiDi methods, but I can choose to define how those new methods get mapped into my binding, and then implementations can add support.

It also may be that a custom protocol binding is required to support your extension until the extension becomes official and support is added to the official protocol bindings.

As we said for AgentCard capabilities, I believe all declared extensions should be supported by all declared AgentInterfaces in the AgentCard to avoid the client needing to handle a matrix of support.

This is my opinion, but hopefully it makes sense and we can get agreement on the definition.

vinoo999:
Thanks, this helps clarify and I think the model mostly makes sense, but just want to be explicit because I still have a gap in this capability negotiation:

* Some Extensions may only be supported by a subset of protocol bindings
* AgentCards should avoid exposing a matrix, i.e. all declared extensions are supported by all AgentInterfaces

From the BiDi example:
* The extension defines new methods + behavior
* Each protocol binding independently decides whether/how to support them (right now the AgentCard has a capability that is "gRPC only" but feasibly in extension adoption world could be in the extension spec

Given that, it seems like we end up with the implication that if an AgentCard declares an extension, then all advertised protocol bindings must support it.

Which would mean one of the following must be true:

1. AgentCards are effectively binding-constrained when declaring extensions + capabilities (right now okay but maybe not going forward)
   - If only gRPC supports BiDi, the AgentCard can only expose gRPC; alternate agent cards must be shown if there is different capabilities and other agent interfaces supported

2. We allow interface-level capability differences (this feels like a non-starter based on what you mentioned)

3. Extensions must be universally supportable across bindings (which rules out certain classes of extensions?)

Tehsmash:
@vinoo999 

> AgentCards are effectively binding-constrained when declaring extensions + capabilities (right now okay but maybe not going forward)

This is the current state declared in the specification: https://github.com/a2aproject/A2A/blob/main/docs/specification.md#51-functional-equivalence-requirements

This would mean exactly as you say, if you state you support BiDi streaming, you can only declare gRPC endpoints, and no HTTP+JSON ones.

> We allow interface-level capability differences (this feels like a non-starter based on what you mentioned)

This was something I proposed for capabilities like streaming (its deep in a thread somewhere) but @darrelmiller was rightly concerned about a negotiation explosion on the client side.

Now there may be workarounds using GetExtendedAgentCard for example if you make a GetExtendedAgentCard request with an `A2A-Extensions: ...` list in the headers it could return a card with a reduced set of AgentInterfaces. This needs more thought though.

> Extensions must be universally supportable across bindings (which rules out certain classes of extensions?)

The way operations / protocol bindings are implemented today this would essentially rule out all extensions which modify the data model or add operations. Now it could be possible to add a "special" endpoint/operation in the A2A protocol on top of which all "extension" operations are implemented as "meta" operations. Again this needs more thought.

vinoo999:
Makes sense. Nothing in this proposal prevents us from following up on how operation extension points + custom protocol bindings can evolve so lets approve and create an issue where we can follow up discussion on it?

PR #1627, opened 2026-03-13
fix(spec): recent transcoding-related error changes

yarolegovich:
Fix error code mappings table so that http codes correspond to [`google.rpc.Code`](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L32)-defined mappings.

Undo changes made in https://github.com/a2aproject/A2A/pull/1600 to jsonrpc error handling section. The changes made were meant to address http-grpc transcoding compatibility. I don't see a convincing reason for forcing jsonrpc error data format, so I assume JSONRPC section was updated by mistake.

Fixes https://github.com/a2aproject/A2A/issues/1596

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1182,12 +1182,12 @@ All A2A-specific errors defined in [Section 3.3.2](#332-error-handling) **MUST**
 | `TaskNotFoundError`                   | `-32001`      | `NOT_FOUND`           | `404 Not Found`              |
 | `TaskNotCancelableError`              | `-32002`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
 | `PushNotificationNotSupportedError`   | `-32003`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
-| `UnsupportedOperationError`           | `-32004`      | `UNIMPLEMENTED`       | `501 Not Implemented`        |
+| `UnsupportedOperationError`           | `-32004`      | `FAILED_PRECONDITION`       | `400 Bad Request`        |
 | `ContentTypeNotSupportedError`        | `-32005`      | `INVALID_ARGUMENT`    | `400 Bad Request`            |
 | `InvalidAgentResponseError`           | `-32006`      | `INTERNAL`            | `500 Internal Server Error`  |
 | `ExtendedAgentCardNotConfiguredError` | `-32007`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
 | `ExtensionSupportRequiredError`       | `-32008`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
-| `VersionNotSupportedError`            | `-32009`      | `UNIMPLEMENTED`       | `501 Not Implemented`        |
+| `VersionNotSupportedError`            | `-32009`      | `FAILED_PRECONDITION`       | `400 Bad Request`        |
 
 **Custom Binding Requirements:**
 


docs/specification.md
```diff
@@ -1180,14 +1180,14 @@ All A2A-specific errors defined in [Section 3.3.2](#332-error-handling) **MUST**
 | A2A Error Type                        | JSON-RPC Code | gRPC Status           | HTTP Status                  |
 | :------------------------------------ | :------------ | :-------------------- | :--------------------------- |
 | `TaskNotFoundError`                   | `-32001`      | `NOT_FOUND`           | `404 Not Found`              |
-| `TaskNotCancelableError`              | `-32002`      | `FAILED_PRECONDITION` | `409 Conflict`               |
-| `PushNotificationNotSupportedError`   | `-32003`      | `UNIMPLEMENTED`       | `400 Bad Request`            |
-| `UnsupportedOperationError`           | `-32004`      | `UNIMPLEMENTED`       | `400 Bad Request`            |
-| `ContentTypeNotSupportedError`        | `-32005`      | `INVALID_ARGUMENT`    | `415 Unsupported Media Type` |
-| `InvalidAgentResponseError`           | `-32006`      | `INTERNAL`            | `502 Bad Gateway`            |
+| `TaskNotCancelableError`              | `-32002`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
```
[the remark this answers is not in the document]
yarolegovich: The stated goal is compatibility with gRPC transcoding systems.  A transcoder will use the canonical mapping, producing different status codes than the spec prescribes

Orkhan647:
❤️

mikeas1:
Changes to the JSON-RPC section were made to ensure a common error detail format is shared across all transports. That ends up being `google.rpc.ErrorInfo` since it's what's used on gRPC, which was then subsequently pushed into JSON+HTTP due to ensuring transcoding is compatible.

The HTTP Status code mapping change makes sense, I should have included that in the original PR. If you split this PR into two, we should be able to get the HTTP Status code changes submitted. We can discuss the JSON-RPC changes further.

docs/specification.md
```diff
@@ -1180,14 +1180,14 @@ All A2A-specific errors defined in [Section 3.3.2](#332-error-handling) **MUST**
...
+| `PushNotificationNotSupportedError`   | `-32003`      | `UNIMPLEMENTED`       | `501 Not Implemented`        |
```
darrelmiller: @yarolegovich 501 is not the right status code here. Is this what gRPC transcoding uses?

darrelmiller:
@yarolegovich Thanks for this.  Based on the conversation in last week's TDC meeting, we are all onboard with this change to the JSON RPC data element. 
My only issue is the 501s.

docs/specification.md
mikeas1: For the `UNIMPLEMENTED` gRPC Code, yes: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L161

yarolegovich:
> My only issue is the 501s.

I can see the semantic mismatch here. Should we update gRPC code to be `FAILED_PRECONDITION` instead of `UNIMPLEMENTED`? In this case the matching code would be 400.

docs/specification.md
```diff
@@ -2424,17 +2424,6 @@ JSON-RPC error responses use the standard [JSON-RPC 2.0 error object](https://ww
...
-
-For A2A-specific errors, implementations **MUST** include a `google.rpc.ErrorInfo` message in the `data` array with:
-
-- `@type`: Set to `"type.googleapis.com/google.rpc.ErrorInfo"`
-- `reason`: The A2A error type in UPPER_SNAKE_CASE without the "Error" suffix (e.g., `TASK_NOT_FOUND`)
-- `domain`: Set to `"a2a-protocol.org"`
-- `metadata`: Optional map of additional error context
-
-Additional error context **MAY** be included in the `data` array.
-
```
Tehsmash, line 2464: This was added to align the error metadata for all transports, i.e. if you get a TaskNotFound error then the data body should be aligned to make it easier for the SDKs convert to common Exceptions/Errors.  According to the JSONRPC schema data can be any JSON type, object or list. gRPC and HTTP are aligned on a list so we did the same for JSONRPC, this allows for multiple types of metadata to be carried per error, you can see any example of that here: https://google.aip.dev/193#http11json-representation

yarolegovich: I don't mind data being an array. but if you're talking about grpc.ErrorInfo-shaped payload in it I don't think its existence makes it any easier for SDKs to convert to a runtime error representation. my point is it's the opposite, because it becomes the second source of truth. in grpc it's needed for error code disambiguation because the code set size is small, in HTTP - for transcoder compatibility, in JSONRPC a code can be unambiguously mapped to an error type.

yarolegovich:
> My only issue is the 501s.

@darrelmiller pushed `FAILED_PRECONDITION` & `400` use for `PushNotificationNotSupportedError` instead of `UNIMPLEMENTED` and `501`

docs/specification.md
Tehsmash: My point was that if the error data/details bodies are aligned then the handling can be deduplicated and separated from the protocol wrapper into a separate helper something like:

For JSONRPC:
```python
if error["code"] >= 32000 and error["code"] < 32100:
	raise_exception_from_a2a_error_data(error["data"])
```

For HTTP:
```
if status_code ...:
	raise_exception_from_a2a_error_data(body["error"])
```

For gRPC:
```
except RpcError as e:
	raise_exception_from_a2a_error_data(e.details)
```

This `raise_exception_from_a2a_error_data` would be a central place to map "reason" to a specific Exception class, and also to be able to pull out any other specific details from the data.

On the server side it would be easy to have a similar helper: 
```
try:
	call_request_handler(request)
except A2AException as e:
	details = exception_to_error_details(exc)
	return JSONRPCError(data=details)
...
```

We could handle it special for JSON RPC to map codes to Exceptions and have its own functions to pull out the details, and the reverse for caught exceptions on the server side but it seems redundant to do that.

yarolegovich: sorry, I don't understand. user code should not deal with transport level error representations regardless of how JSONRPC binding represents errors.

wire format mapping logic should be localized. it will exist anyway: eg. to parse a standard [jsonrpc error](https://github.com/a2aproject/a2a-go/blob/main/internal/jsonrpc/jsonrpc.go#L85) payload or get status details [in gRPC](https://github.com/a2aproject/a2a-go/blob/main/internal/grpcutil/errors.go#L123). that's the stage at which all formats should map the wire error representation to [runtime error representation](https://github.com/a2aproject/a2a-go/blob/main/a2a/errors.go#L79).

now, if we force JSONRPC to have ErrorInfo containing the exact same information as the code already carries - how will it simplify the code to which the first link in this message point to?

Tehsmash: I understand your point, there are two aspects to my argument for standardising on the AIP / gRPC ErrorDetails structure (not ErrorInfo specifically):

* If we make "details"/"data" a list so that we can return different types of details then we need an identifier to separate the types of details from the other types of details in the list. In this case "ErrorInfo" is just the standard "type" for a generic application error with metadata.

* There is nothing in the spec today which defines the expected content for each type of error, there is some recommendation that details should contain contextual information about the error, but nothing solid. AIP / Google provides a set of standard detail payloads for things like BadRequest with violations. If we standardise on these then the user code knows what data it's expecting to receive for different errors.  We could actually create A2A specific types too for A2A specific errors and define the payloads explicitly instead of using ErrorInfo.

ehsavoie:
Could we also clarify the payload format: 
```json
  "error": {
    "code": -32001,
    "message": "Task not found",
    "data": {
      "taskId": "nonexistent-task-id",
      "timestamp": "2025-11-09T10:30:00.000Z"
    }
  }
}
```
while in the specs for HTTP+JSON/REST:
```
HTTP/1.1 404 Not Found
Content-Type: application/json

{
  "error": {
    "code": 404,
    "status": "NOT_FOUND",
    "message": "The specified task ID does not exist or is not accessible",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "TASK_NOT_FOUND",
        "domain": "a2a-protocol.org",
        "metadata": {
          "taskId": "task-123",
          "timestamp": "2025-11-09T10:30:00.000Z"
        }
      }
    ]
  }
}
```
Should the field be **data** or **details** ? 
Do we have to support the ErrorInfo there as well ?

darrelmiller:
@yarolegovich We discussed this at length in the TSC on Tuesday. 

We would like to enable both the HTTP+JSON and the JSON-RPC to support returning the same array of ErrorInfo objects. We understand there is some redundancy in the JSON-RPC case due to the A2A specific error numbers.  However, we think there is value in being able to keep a type attribute in the type, so that servers can return secondary errors.  This could become very helpful for implementing extensions, that have their own specific errors.

So, `data` in JSON-RPC should be an array of ErrorInfo in the same way that `details` is in the HTTP+JSON payload.

Would you be open to making this specification change?

@ehsavoie In HTTP+JSON the member should be `details` because it maps to the gRPC error object that is required for transcoding.  However, in JSON-RPC, the extension member is called `data`.


/cc @mikeas1 @Tehsmash @muscariello @herczyn

yarolegovich:
> So, data in JSON-RPC should be an array of ErrorInfo in the same way that details is in the HTTP+JSON payload.

Maybe that's what you meant, but error data should be an an array of structured values one of which must be of type `ErrorInfo`. That's what AIP-193 says:

> Each type of detail payload must be included at most once. For example, there must not be more than one BadRequest message in the details, but there may be a BadRequest and a PreconditionFailure. All error responses must include an ErrorInfo within details. 

Do we want to force all values in the array to be an object with `@type` key?

docs/specification.md
yarolegovich: > standardising on the AIP / gRPC ErrorDetails structure (not ErrorInfo specifically):

Are you proposing to limit the shape of objects which can go into error details to:
```
BadRequest
DebugInfo
ErrorInfo
Help
LocalizedMessage
PreconditionFailure
QuotaFailure
RequestInfo
ResourceInfo
RetryInfo
```
Or just provide some kind of "first-class" support for those?

ehsavoie:
@darrelmiller why not called the jsonrpc field **details** then ? I think it is confusing to have different payload formats. The the data can be a map/object and if we want transcoding then so be it.

darrelmiller:
@ehsavoie The JSON-RPC specification says the member must be named `data` https://www.jsonrpc.org/specification#error_object

@yarolegovich I had missed that `details` was defined as `any` here https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto

That is going to be annoying for anyone not using gRPC and a strongly typed language. Parsing types where the type discriminator is inside the object is a pain. I'd rather this be an array of ErrorInfo so it can be typed collection and metadata can be defined by whatever appears in @type.  That would still comply with AIP-193 and just be easier to deal with.

darrelmiller:
Update from Tuesday's TSC call.  `details` in HTTP+JSON and `data` in JSON-RPC are going to map to the ProtoJSON equivalent to an array of `any` .  This means that each object in the array MUST have at least a @type attribute.  To further refine the error reporting in the surrounding object,  `errorInfo` type should be used as an object in the array.  To attach structured data to generic errors like bad request, the `badRequest` type can be used.

yarolegovich:
@darrelmiller thanks for the summary. This was a busy week, but managed to add the changes to the PR before todays meeting

docs/specification.md
```diff
@@ -1180,14 +1180,14 @@ All A2A-specific errors defined in [Section 3.3.2](#332-error-handling) **MUST**
...
 | :------------------------------------ | :------------ | :-------------------- | :--------------------------- |
 | `TaskNotFoundError`                   | `-32001`      | `NOT_FOUND`           | `404 Not Found`              |
-| `TaskNotCancelableError`              | `-32002`      | `FAILED_PRECONDITION` | `409 Conflict`               |
-| `PushNotificationNotSupportedError`   | `-32003`      | `UNIMPLEMENTED`       | `400 Bad Request`            |
-| `UnsupportedOperationError`           | `-32004`      | `UNIMPLEMENTED`       | `400 Bad Request`            |
-| `ContentTypeNotSupportedError`        | `-32005`      | `INVALID_ARGUMENT`    | `415 Unsupported Media Type` |
-| `InvalidAgentResponseError`           | `-32006`      | `INTERNAL`            | `502 Bad Gateway`            |
+| `TaskNotCancelableError`              | `-32002`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
+| `PushNotificationNotSupportedError`   | `-32003`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
+| `UnsupportedOperationError`           | `-32004`      | `UNIMPLEMENTED`       | `501 Not Implemented`        |
```
darrelmiller:
```suggestion
| `UnsupportedOperationError`           | `-32004`      | `FAILED_PRECONDITION`       | `400 Bad Request`        |
```
5XX errors indicate a server error. A 4XX error more accurately indicates a client error.

docs/specification.md
```diff
@@ -1180,14 +1180,14 @@ All A2A-specific errors defined in [Section 3.3.2](#332-error-handling) **MUST**
...
+| `ContentTypeNotSupportedError`        | `-32005`      | `INVALID_ARGUMENT`    | `400 Bad Request`            |
+| `InvalidAgentResponseError`           | `-32006`      | `INTERNAL`            | `500 Internal Server Error`  |
 | `ExtendedAgentCardNotConfiguredError` | `-32007`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
 | `ExtensionSupportRequiredError`       | `-32008`      | `FAILED_PRECONDITION` | `400 Bad Request`            |
-| `VersionNotSupportedError`            | `-32009`      | `UNIMPLEMENTED`       | `400 Bad Request`            |
+| `VersionNotSupportedError`            | `-32009`      | `UNIMPLEMENTED`       | `501 Not Implemented`        |
```
darrelmiller:
```suggestion
| `VersionNotSupportedError`            | `-32009`      | `FAILED_PRECONDITION`       | `400 Bad Request`        |
```
As above.

darrelmiller:
Merging this PR even with the Markdown linting error because it is not actually a linting error it is failing to call the linter.
Also, this is changing the official v1.0.0 specification without bumping versions or patch.  We do not want to set precedent for using patches to fix errors in a published specification.  However, we do believe that this is early enough in the process while we are still stabilizing SDKs to justify making this correction to the v1.0.0.
We appreciate your forgiveness for breaking our own rules.

Issue #883, opened 2025-07-17
[Feat]: Documenting Best Practices for Exposing Multi-Skill CrewAI Agents as Unified External Services


nandrews729:
### Is your feature request related to a problem? Please describe.

Organizations adopting agentic agents for building powerful multi-agent workflows often need to integrate these internal "agents" into a broader ecosystem of external agent frameworks, such as the Agent-to-Agent (A2A) Protocol. A common challenge arises when a single external A2A Agent is designed to represent a common functional domain (e.g., "Customer Operations" or "Content Management") but needs to expose multiple distinct "skills" or "services" within that domain (e.g., "Order Tracking," "Refund Processing," "Product Information" under "Customer Operations").

The A2A framework itself focuses on routing requests to a specific A2A Agent's endpoint, rather than directly specifying an internal "skill" in the routing path. This means that any differentiation between the various skills/services offered by a single A2A Agent must be handled internally by the A2A Agent's logic, typically by inferring the desired "skill" from the incoming request's payload (params field).   The examples show a "one-app-per-agent" model.

### Describe the solution you'd like

It would be good to document the best practices for this as I can see there are a number of approaches which could be adopted, but you want it to scale and be easily implemented and maintained.  
The current approach seems to focus around a "one-app-per-agent" approach.
I guess there are a number of approaches to solve, but each comes with pros/cons. 

1. Where you define multiple agents/skills on the same port with different urls/rpc for invoking. with different agent executors.
2.  Where you can have multiple skills per agent, but then the intent and routing of the request needs to be handled within the a2a framework to invoke different agent executors for example.  (this would mean the client passing the agent name or skills as part of the a2a request and then the executor can map the intent to invoke different internal agents.
3. You can ensure the client passes this in the internal agent payload and then the internal agent uses the skill/name (or has to do more intent analysis to dynamically determine the internal routing or assignment logic.   e.g. what tools to assign an agent, or which internal agents should be used or which crews/flows (if part of a crew or larger process flow) - this approach would mean that if the a2a framework could be replaced if needed as the routing logic is still held internally.

Without clear guidance, developers may resort to less optimal integration patterns, such as deploying separate A2A Agents for each individual "skill," leading to increased complexity in discovery, management, and orchestration for both the A2A framework and the client agents (and port proliferation as these agentic agents grow in deployment).

Also I think the examples should be written in a way which the a2a wrapper code is contained within the a2a framework rather than polluting the native agentic code  (I've only looked at the crewai example btw).

[...]

Issue #773, opened 2025-06-20
How to avoid LLM-based orchestration logic for task routing within an A2A agent?


prokolyvakis:
Hello,

I'm working on integrating the A2A protocol into a service that supports multiple tool invocations. The orchestration logic (e.g., deciding whether to trigger "Tool A" vs "Tool B") is quite simple and deterministic.

I would like to avoid using an LLM (_as shown in the [a2a_mcp](https://github.com/google-a2a/a2a-samples/tree/main/samples/python/agents/a2a_mcp) sample_) to make this decision for every query, as that introduces latency and unnecessary cost for a fixed logic path.

Instead, I'm exploring the option of having the user message include a `"kind": "data"` payload with a `task_type` field (e.g., `"task_type": "tool_a"`), which can map directly to a known AgentSkill in the `AgentCard`.

To wrap up, my question is:

> **Is this an idiomatic or recommended A2A-compliant way to structure task routing logic without requiring a dedicated orchestrator agent or LLMAgent?**

Any thoughts or design recommendations would be appreciated.

holtskinner:
A2A Protocol does not depend on any usage of LLMs/Generative AI models, it only defines the communication between agents. As long as the protocol is followed for all external agent communication, it doesn't matter how agent routing/decisions are made. For example, the routing logic could be done with plain conditionals (`if`/`else`)

PR #1655, opened 2026-03-18
feat: Add QuerySkill() operation for runtime skill capability introspection

Srinivasoo7:
# Description
This PR addresses the need for dynamic capability discovery in the A2A protocol. Currently, orchestrators rely on the static `AgentCard`, which cannot represent multidimensional or runtime-specific constraints (e.g., "Can you process a 500MB audio file right now?"). 

Several active discussions confirm real community pain - #883 , #773 , #921,  #166  all describing the same root cause: no protocol-level mechanism exists for a client to probe whether a remote agent can handle a specific skill at runtime.

This PR introduces the `QuerySkill` RPC to establish a proactive capability declaration standard prior to task execution.

## Changes:
- **Added `QuerySkill` RPC** to `A2AService` in `a2a.proto` for dynamic skill introspection.
- **Added `skill_query` boolean flag** to `AgentCapabilities` so agents can explicitly declare support for this feature.
- **Added `SkillQueryRequest` and `SkillQueryResult` messages**, including a `SkillSupportLevel` enum (`SUPPORTED`, `PARTIAL`, `UNSUPPORTED`).
- **Added ADR-002 (`adr-002-queryskill.md`)** detailing the context, decision, consequences, and rejected alternatives.

## Details:
- Supports negotiation via a `constraints` map for `PARTIAL` support levels.
- Includes a `confidence` float and `reason` string for probabilistic routing.
- Returns a `cache_ttl_seconds` to allow orchestrators to cache results (and utilize `stale-if-error` semantics), preventing the `QuerySkill` RPC from becoming a performance bottleneck.

### DIFF
--- a/adrs/adr-002-queryskill.md
+++ b/adrs/adr-002-queryskill.md
@@ -13,7 +13,7 @@ The result is that orchestrators either over-delegate (sending tasks to agents t
 We will add a `QuerySkill` RPC to `a2a.proto` as a capability declaration, not a task execution.
 
 The design has three layers:
-1.  **Proto definition**: Add `SkillQueryRequest` and `SkillQueryResult` messages, and the `SkillSupportLevel` enum with values `SUPPORTED`, `PARTIAL`, and `UNSUPPORTED`.
+1.  **Proto definition**: Add `SkillQueryRequest` and `SkillQueryResult` messages, and the `SkillSupportLevel` enum with values `SKILL_SUPPORT_LEVEL_SUPPORTED`, `SKILL_SUPPORT_LEVEL_PARTIAL`, and `SKILL_SUPPORT_LEVEL_UNSUPPORTED`.
 2.  **Protocol binding**: Map to JSON-RPC as `skills/query`, to gRPC as `QuerySkill` on the existing `A2AService`, and to HTTP/REST as `POST /skills/{skill_id}/query`. All bindings must return `UnsupportedOperationError` if the server hasn't declared `capabilities.skill_query: true` in its `AgentCard`.
 3.  **Caching and staleness**: `SkillQueryResult` includes a `cache_ttl_seconds` field. Furthermore, we define `stale-if-error` semantics: if a `QuerySkill` call fails transiently, the orchestrator MAY use a cached `SUPPORTED` result for up to 2 × `cache_ttl_seconds`.
 
--- a/adrs/adr-002-queryskill.md
+++ b/adrs/adr-002-queryskill.md
@@ -1,0 +1,28 @@
+# ADR 002: Dynamic Skill Introspection (`QuerySkill`)
+
+## Status
+Proposed
+
+## Context
+The A2A protocol today relies entirely on the `AgentCard` for capability discovery. That card is a static, cached document published at `/.well-known/agent.json` or accessed via `GetExtendedAgentCard`.
+The problem is that agent capabilities are not truly static. They vary based on load, configuration, runtime environment, available models, feature flags, and context.
+An orchestrator agent has no way to ask: "Can you handle this specific task right now, with these parameters?"
+The result is that orchestrators either over-delegate (sending tasks to agents that will fail) or under-delegate (avoiding agents whose card looks uncertain). There is also no concept of partial capability — an agent that can do 80% of a requested skill but not the full scope cannot express that today.
+
+## Decision
+We will add a `QuerySkill` RPC to `a2a.proto` as a capability declaration, not a task execution.
+
+The design has three layers:
+1.  **Proto definition**: Add `SkillQueryRequest` and `SkillQueryResult` messages, and the `SkillSupportLevel` enum with values `SUPPORTED`, `PARTIAL`, and `UNSUPPORTED`.
+2.  **Protocol binding**: Map to JSON-RPC as `skills/query`, to gRPC as `QuerySkill` on the existing `A2AService`, and to HTTP/REST as `POST /skills/{skill_id}/query`. All bindings must return `UnsupportedOperationError` if the server hasn't declared `capabilities.skill_query: true` in its `AgentCard`.
+3.  **Caching and staleness**: `SkillQueryResult` includes a `cache_ttl_seconds` field. Furthermore, we define `stale-if-error` semantics: if a `QuerySkill` call fails transiently, the orchestrator MAY use a cached `SUPPORTED` result for up to 2 × `cache_ttl_seconds`.
+
+## Consequences
+*   **Proactive Capability Checking:** Clients can avoid costly `SendMessage` calls when the agent cannot support the specific request right now.
+*   **Partial Negotiation:** Agents can use the `constraints` map to request changes from the client (e.g., chunking smaller files).
+*   **Explicit declaration:** Agents must opt into this by setting `skill_query = true` in their capabilities, maintaining compatibility.
+
+## Rejected Alternatives
+*   **Using `GetExtendedAgentCard`:** This is too coarse and cannot carry the varied runtime context without combinatorial explosion.
+*   **Adding a `supports` field to `SendMessage`:** This requires the client to fully constuct the message payload (potentially including large artifacts) before discovering unsupportability, defeating the core benefit.
+*   **Polling `GetTask` for failures:** This is reactive, not proactive, and still incurs the initial transmission and processing cost. The chosen design is proactive and non-destructive.
--- a/specification/a2a.proto
+++ b/specification/a2a.proto
@@ -812,17 +812,17 @@ message ListTaskPushNotificationConfigsResponse {
 // Support level for a requested skill.
 enum SkillSupportLevel {
   SKILL_SUPPORT_LEVEL_UNSPECIFIED = 0;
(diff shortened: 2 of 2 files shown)


citriac:
Coming at this from the implementation side — I've been running an agent registry ([Agent Exchange Hub](https://clavis.citriac.deno.net)) where agents declare both static capabilities (AgentCard) and runtime availability (/signals), so I have a data point on where the static/dynamic split breaks down.

The `QuerySkill` RPC direction is right, but I'd push back on one design assumption: **the confidence float + PARTIAL support level may be adding negotiation complexity that most orchestrators won't use.**

From what I've observed in practice, the cases that actually matter are simpler:

1. **Hard no** — "this file is 800MB and the agent caps at 100MB" → UNSUPPORTED, route elsewhere
2. **Latency mismatch** — "this task needs <5min and the agent wakes every 4h" → also a hard no, just a different axis
3. **Soft limit** — "I can do this but it'll take longer than usual" → PARTIAL, but the caller just needs `estimatedSeconds`, not a negotiation round-trip

The `constraints` map in the proposal handles case 3, but it requires a synchronous pre-check call before every task. For heartbeat/polling agents (discussed in #1667), that pre-check call will often time out because the agent isn't listening.

**Alternative that might cover 80% of cases with less protocol surface:**

Instead of a new RPC, extend the `extensions[]` array in `AgentCapabilities` to allow agents to declare capability constraints inline:

```json
"capabilities": {
  "extensions": [
    {
      "uri": "https://a2a-protocol.org/extensions/skill-constraints/v1",
      "params": {
        "skillId": "audio-transcription",
        "maxFileSizeBytes": 104857600,
        "supportedFormats": ["audio/mpeg", "audio/wav"],
        "estimatedSecondsPerMB": 2
      }
    }
  ]
}
```

This is discoverable without a connection (just fetch the AgentCard), composable with the existing extensions mechanism, and doesn't require the agent to be live when the orchestrator is making routing decisions.

The `QuerySkill` RPC still makes sense for the cases that genuinely need runtime introspection (e.g. "do you have GPU available right now?"), but that's a narrower surface than the proposal suggests.

Curious if anyone has data on how often the runtime-specific introspection is actually needed vs. "this would have been fine as a static declaration."

Srinivasoo7:
Hi @citriac
 
Genuinely valid and useful data points from someone running a live registry. Let me address the specific concern

1. **On confidence float:** Thanks for pointing this out, that this adds orchestrator complexity for minimal gain. I'll work on it. The SUPPORTED / PARTIAL / UNSUPPORTED enum with the constraints map and reason string is sufficient for routing decisions. No orchestrator needs a float to branch.

2. **On the AgentCard extensions alternative:** I like this pattern and it's a valid complement, but it solves a different problem. Static constraints (max file size, supported formats) belong exactly where you suggest: declared once in the AgentCard, discoverable without a live connection. I'll make a note of this in the ADR as the recommended pattern for static constraints.
However, QuerySkill exists for the cases that genuinely can't be static: current GPU availability, live queue depth, API rate limit headroom, feature flag state. Your own closing question — "do you have GPU available right now?"  is precisely the use case that I see where AgentCard extension cannot answer. A cached document cannot express runtime state.

3. **On heartbeat/polling agents timing out:** The cache_ttl_seconds field in SkillQueryResult is specifically designed for this. An agent that wakes every 4 hours returns UNSUPPORTED with cache_ttl_seconds: 14400. The orchestrator caches that result and never calls again until expiry. No pre-check timeout on every task — just one cached result per TTL window. 


**To add on regarding the question with data:** I have as well mentioned in the description of the PR regarding the related discussions (#883, #773, #921) all discuss cases where static AgentCard is insufficient, file size constraints that change with load, skills that depend on external API availability, latency guarantees that vary with queue depth. 
If there is any contratry findings/data that I missed on the same from implementation, would like to better understand to reshape my scope of this PR.

Srinivasoo7:
Hi @balthazar-bot 
Thanks for sharing your views, it helps!

1. On the TTL: cache_ttl_seconds is already in SkillQueryResult, this was specifically designed so agents returning UNSUPPORTED can tell orchestrators "don't call me again for N seconds." A heartbeat agent waking every 4 hours returns cache_ttl_seconds: 14400 and the orchestrator never hammers the endpoint. Happy to rename it validForSeconds otherwise 

2. On DEGRADED as a fourth state: I'll incorporate this one. PARTIAL would be more towards structural constraint and DEGRADED will be transient runtime impairment if I understand your suggestions correctly. 

3. On Batch Query Mode: For this PR, adding a batch variant means adding a second message type, a second RPC, and second binding documentation for all three transports. I'll address that in the subsequent PR referencing this one. 

For the current scope of the PR, I want to merge QuerySkill (singular) now as the minimal viable surface and file a follow-on issue for QuerySkills (batch) that references this PR. The batch mode is a pure extension of the same semantics, so the follow-on PR will be straightforward once the singular RPC is in.

Srinivasoo7:
/gemini review what's the process to approve this PR?

kevinkaylie:
QuerySkill() for runtime skill capability introspection is a natural fit for Enclave permission discovery.

**How it works:**
- Enclave `role → permissions` defines what each agent can do at the enclave level
- A `QuerySkill()` call against an Enclave member returns its permission set — including runtime constraints like scope limits and spend limits
- This tells the orchestrator: can this agent handle a 500MB audio file right now, given its current role and active constraints?

**The constraint picture at query time:**
```json
{
  "agent": "did:agentnexus:z...",
  "role": "member",
  "permissions": ["docs:read", "search:invoke"],
  "active_constraints": {
    "scope": ["legal-docs"],
    "spend_limit_remaining": 45.00
  },
  "can_handle": false
}
```

**On the multi-dimensional constraint problem:** Enclave scopes are orthogonal — scope, spend, time, depth, tools are all evaluated independently and combined. This avoids the single-dimensional collapse problem.

Repo: https://github.com/kevinkaylie/AgentNexus

Srinivasoo7:
Hi @kevinkaylie 

Thanks for sharing this, this is an interesting framing, and it highlights a real design tension worth making explicit.
I'd draw a distinction between two separate questions that QuerySkill could answer:
From what I understand in regards to your comment, 
1) **Capability check:** "Can this agent physically handle this task right now?" (resources, file size, latency, queue depth)
2) **Authorization check:** "Is this agent permitted to handle this task given its current role and constraints?" (scope, spend limits, access controls)

This PR current proposal is scoped to (1). Folding authorization state into SkillQueryResult would couple capability routing to identity management, which I think belongs in the auth layer (the A2A spec already has SecurityScheme and in-task authorization for this).
That said, your point about orthogonal constraint dimensions is the most useful design signal in this comment. The current flat constraints map<string, string> loses the axis information, an orchestrator receiving {"scope": "legal-docs", "spend_remaining": "45.00", "max_file_mb": "100"} can't tell which dimension caused PARTIAL or what the recovery action is for each independently.
A typed constraint structure, even something simple like a repeated SkillConstraint with a dimension enum and value string would preserve that signal without requiring the full permission model.

I'll work on this to whether replacing the flat map with a typed constraint list is worth the added surface for this PR, or whether it's better left for a follow-on. Will check out AgentNexus the multi-dimensional constraint evaluation model looks worth understanding in more depth.


Thanks

Srinivasoo7:
@holtskinner could you support on this or help me point to the right reviewer?

Appreciate your support 
Thanks

Tehsmash:
Hi @Srinivasoo7 thanks for the PR its interesting. Something I'm struggling with though is why GetExtendedAgentCard isn't enough for this. GetExtendedAgentCard was specifically designed with dynamic capabilities / skills in mind so that different users of the same agent might see different skills available at different times. I see this proposal adds additional semantics but I'm struggling with the idea of a partially supported skill.

The current spec would expect that the AgentCard in the catalog / well-known would essentially be empty no-skills available by default, the client should not assume anything is available and should always check what they can do/whats available at this point in time for this client through GetExtendedAgentCard.

What seems to be missing from this story however is related to discovery i.e. we need to find the Agent in the first place most likely based on the skill's its capable of doing. So we want Agents to be able to declare what they *might* be able to do assuming that you have the correct permissions and the Agent is able to do it etc, then to tell you what you can actually do when you authenticate etc.

Srinivasoo7:
Hi @Tehsmash 

Your comments lines up with some feedback I got from previous reviews on the implementation side, I will work that helped narrow the scope.

For example, one point raised was those static constraints (e.g., file size limits, supported formats) should live in AgentCard extensions, since they need to be discoverable without requiring a live call. That suggests any new mechanism only makes sense if it represents truly runtime-only state that can’t be captured in a cached document.

Another concern was that something like QuerySkill could be interpreted as an authorization check, rather than a capability check. I agree that would be problematic, and I’d want to keep those concerns clearly separated capability as “can the agent handle this right now,” while authorization remains in the existing auth layer (SecurityScheme).

That said, combining these constraints runtime-only and capability-only significantly narrows the surface area where a QuerySkill-style RPC would actually add value.

Given your point about GetExtendedAgentCard, I’m starting to think this may not justify a new RPC in the core spec after all.

The more concrete gap I’m seeing is still around how a client expresses routing intent once it already knows what’s available (e.g., after calling GetExtendedAgentCard), rather than further capability introspection.

Curious if you see that as the more relevant direction to focus on, instead of introducing another RPC.

PR #1640, opened 2026-03-16
docs(spec): use `application/json` content type in Section 6 examples

matanweks:
## Summary

- Fixes all 18 HTTP examples in Section 6 that incorrectly use `application/a2a+json` as the `Content-Type` header
- Changes them to `application/json`, consistent with:
  - **Section 9.1** (JSON-RPC Protocol Requirements): `Content-Type: application/json for requests and responses`
  - **Section 11.1** (HTTP+JSON/REST Protocol Requirements): `Content-Type: application/json for requests and responses`
  - All official SDKs (Python, Java) and the TCK, which all use `application/json`
- The IANA media type registration in Section 14.1.1 is intentionally left unchanged

Fixes #1639

## Test plan

- [x] All `Content-Type` headers in Section 6 examples now read `application/json`
- [x] Section 14.1.1 IANA registration still references `application/a2a+json`
- [x] Consistent with Section 9.1 and 11.1 protocol requirements
- [x] Consistent with all official SDK implementations and TCK tests

PR #1689, opened 2026-03-26
docs: clarify HTTP+JSON error content type

msaleme:
## Summary
- explicitly call out that the JSON-RPC and HTTP+JSON bindings use `application/json` for both success and error bodies per AIP-193
- fix the two sample error responses that still showed `application/problem+json`
- update the Section 6.2 SSE example so `task`, `artifactUpdate`, and `statusUpdate` events include `contextId`

Fixes #1685
Fixes #1683

## Testing
- docs change only

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1373,11 +1373,11 @@ Authorization: Bearer token
 HTTP/1.1 200 OK
 Content-Type: text/event-stream
 
-data: {"task": {"id": "task-uuid", "status": {"state": "TASK_STATE_WORKING"}}}
+data: {"task": {"id": "task-uuid", "contextId": "context-uuid", "status": {"state": "TASK_STATE_WORKING"}}}
 
-data: {"artifactUpdate": {"taskId": "task-uuid", "artifact": {"parts": [{"text": "# Climate Change Report\n\n"}]}}}
+data: {"artifactUpdate": {"taskId": "task-uuid", "contextId": "context-uuid", "artifact": {"artifactId": "artifact-uuid", "parts": [{"text": "# Climate Change Report\n\n"}]}}}
 
-data: {"statusUpdate": {"taskId": "task-uuid", "status": {"state": "TASK_STATE_COMPLETED"}}}
+data: {"statusUpdate": {"taskId": "task-uuid", "contextId": "context-uuid", "status": {"state": "TASK_STATE_COMPLETED"}}}
 ```
 
 ### 6.3. Multi-Turn Interaction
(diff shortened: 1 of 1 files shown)


msaleme:
Added artifactId to the SSE artifact update example in cf67005 so it matches the proto definition.

jmesnil:
/lgtm for the error examples. 

@msaleme While you're looking at `Content-Type`, there is also #1640 that fixes some weird `application/a2a+json`.
This PR should also be merged so that content-type is consistent across the spec.

msaleme:
Thanks Jeff. #1640 covers the broader Section 6 sweep so once that lands the whole chapter will read `application/json`. This PR stays scoped to the new SSE example + the protocol requirement bullets, so no conflicts today. Happy to rebase if needed after 1640 merges.

docs/specification.md
```diff
@@ -1595,7 +1595,7 @@ Authorization: Bearer token
 
 ```http
 HTTP/1.1 400 Bad Request
-Content-Type: application/problem+json
+Content-Type: application/json
```
darrelmiller, line 1598: While this change to the content-type is correct, there are also other changes to the error payloads that need to be updated based on the most recent PR https://github.com/a2aproject/A2A/pull/1627 relating to error payloads.

msampathkumar:
@msaleme - Please check the latest changes in specifications.

msaleme:
@msampathkumar thanks for the pointer, and apologies for the slow turnaround. I rebased onto current `main` and re-verified against the present spec text. Checking the latest changes was the right call, because one part of this PR had become wrong.

**What changed upstream:** section 11.1 now reads `application/a2a+json` **SHOULD** be used for requests and responses. The original version of this PR proposed `application/json` for HTTP+JSON error bodies, which is no longer the right remedy for that binding.

**What is still open on `main`:** the inconsistency itself, and it is a little sharper than I first described it. In 6.4 the request is sent as `application/a2a+json` and the error three lines below answers with `application/problem+json`. The same pattern appears in the 6.5 validation example. Those two are now the only `problem+json` occurrences in the document, against 27 uses of `a2a+json`.

**Revised change, 7 lines in one commit:**

- 6.4 and 6.5 error responses use `application/a2a+json`, matching the request Content-Type directly above them.
- 11.1 keeps `application/a2a+json` and states that it covers error bodies, since an error is a response. No media-type change.
- 9.1 is unchanged in media type. The JSON-RPC binding still uses `application/json`; only the error-body clarification is added there.
- The 6.2 SSE example carries `contextId` on all three events and `artifactId` on the artifact, which is still missing on `main`.

If the intent is instead that A2A errors keep RFC 9457 `problem+json` and 11.1 covers success responses only, then the fix belongs in 11.1 rather than in the examples, and I am happy to invert it. That is the one judgment call here and it is yours, not mine.

Issue #1497, opened 2026-02-18
Proposal: Agent Identity Verification and Trust Framework


thebenignhacker:
## Problem

The A2A protocol recommends TLS certificate validation and allows optional JWS signatures on AgentCards for agent identity. These mechanisms verify the domain hosting an agent but do not verify the agent itself. This creates security gaps in multi-agent ecosystems:

1. No verified agent identity - AgentProvider fields are self-asserted with no verification
2. No trust evaluation framework for agents
3. No trust propagation or authorization boundaries in delegation chains
4. No AgentCard revocation mechanism
5. No application-layer message integrity

## Proposal

An extension-based identity framework using existing AgentExtension and metadata fields, with three verification levels (self-asserted, domain-verified, organization-verified), standardized trust signals, mandatory AgentCard signing for production, revocation endpoints, delegation chain security, and message signing.

Full proposal: docs/proposals/agent-identity-trust-framework.md
Design rationale: docs/proposals/agent-identity-trust-framework-design-rationale.md

chorghemaruti64-creator:
Relevant context from our implementation: we've built and deployed an agent identity + trust system at [Agenium](https://chat.agenium.net) that addresses several points in this proposal.

**What we've shipped:**

1. **Platform-verified identity** — agents get identity from their owner's platform login (Telegram OAuth → `username.telegram` address). No separate verification flow needed — the platform login IS the proof of ownership.

2. **DNS-resolvable agent addresses** — each agent gets an `agent://` URI that resolves via our DNS system. The address persists across endpoint changes.

3. **Trust scoring** — star ratings, review system, and verification badges. Verified agents get search ranking boost (1.5x). Trust data is queryable via API.

4. **Claim & verify flow** — agents can prove ownership via DNS TXT records or email verification. This gives you cryptographic-ish verification without requiring PKI infrastructure.

**Key lesson learned:** the biggest barrier to agent identity adoption isn't the crypto — it's the UX. Requiring agents to manage keypairs or certificates creates massive friction. Leveraging existing platform identity (Telegram, GitHub) and DNS infrastructure gets you 80% of the trust value with 10% of the complexity.

The proposal's "DID-based identity" approach is solid for the long term, but for practical adoption today, platform-verified identity + DNS resolution has been our fastest path.

Live system: [chat.agenium.net](https://chat.agenium.net) — agents respond via permanent DNS addresses.

The-Nexus-Guard:
[...]

PR #1696, opened 2026-03-28
docs: add identity verification guidance for agent discovery

FransDevelopment:
## Summary

- Adds a new **Verifying Agent Identity** section to `agent-discovery.md` documenting how identity verification complements A2A's existing transport-layer security model
- Expands the **Future Considerations** section with concrete areas the community is exploring (registry standardization, capability + identity composition, delegation chains)
- References open-source ecosystem projects and ratified specs that address the identity verification gap using the same cryptographic primitives A2A already uses (JWS [RFC 7515], JSON canonicalization [RFC 8785])

## Context

This addresses the identity verification gap discussed extensively in #1672 (73+ comments), #1497, #1575, and #1511. The A2A specification establishes trust at the transport layer (TLS, OAuth), but does not currently address cryptographic verification of the agent runtime itself. This is a docs-only addition that documents ecosystem approaches without proposing schema changes.

## What this does NOT do

- No changes to `a2a.proto` or the Agent Card schema
- No new normative requirements
- No breaking changes

## References

- Issue: #1672 (Agent Identity Verification for Agent Cards)
- Paper: [Composable Trust Infrastructure for the Agent Internet](https://zenodo.org/records/19263547) (Zenodo, CC-BY 4.0)
- Open Agent Trust Registry: https://github.com/FransDevelopment/open-agent-trust-registry (MIT)
- agent.json: https://github.com/FransDevelopment/agent-json (MIT)
- Agent Identity Working Group: https://github.com/corpollc/qntm/issues/5

## Test plan

- [ ] Verify all external links resolve correctly
- [ ] Confirm markdown passes markdownlint with project config
- [ ] Review that new content matches the tone and style of existing A2A documentation

### DIFF
--- a/docs/topics/agent-discovery.md
+++ b/docs/topics/agent-discovery.md
@@ -91,6 +91,56 @@ To mitigate risks, the following protection mechanisms should be considered:
 
 Any Agent Card containing sensitive data must be protected with authentication and authorization mechanisms. The A2A specification strongly recommends the use of out-of-band dynamic credentials rather than embedding static secrets within the Agent Card.
 
+## Verifying Agent Identity
+
+Discovery and security tell a client *where* an agent is and *how* to authenticate the connection. They do not answer a different question: *is this agent runtime legitimate?*
+
+Transport-layer mechanisms (TLS, OAuth, mTLS) verify the server's domain identity, but they do not provide cryptographic proof of the agent's own identity, its registration status within a trust framework, or whether its credentials have been revoked. In open agent ecosystems where agents from different organizations interact, clients may need to verify these properties before routing a task.
+
+### The Identity Verification Gap
+
+A2A's current security model establishes trust at the transport layer: TLS verifies the server certificate, and OAuth/OpenID Connect handles authorization. This is sufficient when both agents operate within the same trust boundary (for example, agents within a single enterprise). However, when agents operate across organizational boundaries, additional verification questions arise:
+
+- Is this agent's runtime registered with a trusted authority?
+- Has this agent's identity been revoked or suspended?
+- Can the agent's public key be resolved to a verifiable decentralized identifier (DID)?
+- Was this agent authorized to act on behalf of a specific principal, and within what scope?
+
+These are identity verification questions, distinct from the authorization questions that OAuth answers.
+
+### Ecosystem Approaches
+
+The agent identity ecosystem is actively developing open protocols to address this gap. Several approaches are emerging that complement A2A's existing security model:
+
+#### Trust Registries
+
+Federated trust registries maintain cryptographically signed lists of verified agent runtimes. A client discovering an Agent Card can optionally verify the agent's runtime against such a registry before establishing communication. This verification can be performed locally against a cached registry manifest, without requiring network calls at verification time.
+
+Example open-source implementations include the [Open Agent Trust Registry](https://github.com/FransDevelopment/open-agent-trust-registry) (OATR), which provides permissionless registration via Ed25519 proof-of-key-ownership, domain verification, and a [14-step local verification protocol](https://github.com/FransDevelopment/open-agent-trust-registry/blob/main/spec/03-verification.md) using JWS ([RFC 7515](https://tools.ietf.org/html/rfc7515)) attestations with JSON canonicalization ([RFC 8785](https://tools.ietf.org/html/rfc8785)), the same primitives A2A uses for [Agent Card Signing](../specification.md#84-agent-card-signing).
+
+#### Decentralized Identifiers (DIDs)
+
+[W3C Decentralized Identifiers](https://www.w3.org/TR/did-core/) (DIDs) provide a standard way to resolve an agent's identity to a cryptographic public key without depending on a central identity provider. DID methods such as `did:web` (resolved via `/.well-known/did.json`) and `did:key` (self-contained in the identifier) allow agents to advertise verifiable identities that can be independently resolved.
+
+An [Agent Identity Working Group](https://github.com/corpollc/qntm/issues/5) comprising multiple independent projects has ratified three specs for DID-based agent identity: [DID Resolution v1.0](https://github.com/corpollc/qntm/blob/main/specs/working-group/did-resolution.md), [Entity Verification v1.0](https://github.com/corpollc/qntm/blob/main/specs/working-group/entity-verification.md), and [QSP-1 v1.0](https://github.com/corpollc/qntm/blob/main/specs/working-group/qsp1-envelope.md) (encrypted transport). Cross-implementation [test vectors](https://github.com/corpollc/qntm/tree/main/specs/test-vectors) are available.
+
+#### Capability Discovery
+
+Separate from the Agent Card, capability manifests such as [agent.json](https://github.com/FransDevelopment/agent-json) (hosted at `/.well-known/agent.json`) can declare structured capabilities alongside identity metadata including DIDs and public keys. Because A2A Agent Cards (`/.well-known/agent-card.json`) and capability manifests (`/.well-known/agent.json`) use different well-known paths, they can coexist on the same domain without conflict, allowing services to serve both formats for different consumers.
+
+### Composing Identity Verification with A2A
+
+Identity verification is designed to complement, not replace, A2A's existing security model. A typical verification flow alongside A2A:
+
+1. **Discover** the Agent Card via `/.well-known/agent-card.json` (standard A2A)
+2. **Authenticate** using the Agent Card's declared security schemes (OAuth, API key, mTLS)
+3. **Optionally verify** the agent's runtime identity against a trust registry or DID resolution
+4. **Route the task** only if both authentication and identity verification succeed
+
+Steps 1-2 are existing A2A behavior. Steps 3-4 add an optional identity verification layer that services can adopt incrementally based on their trust requirements.
+
+For a comprehensive treatment of composable trust architecture for agent identity, see [Moore, 2026](https://zenodo.org/records/19263547).
+
 ## Caching Considerations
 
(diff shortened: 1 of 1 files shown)


aeoess:
Strong PR. The gap this fills is real — A2A's agent-discovery docs cover transport security but say nothing about identity verification at the application layer. This adds exactly the guidance needed for implementers who want agents to prove who they are, not just that the connection is encrypted.

The approach is right: document the pattern without mandating a specific implementation. Reference the ecosystem work (DID methods, cryptographic primitives, ratified interop specs) as concrete examples without requiring any single one.

One suggestion: the "Future Considerations" section could mention delegation verification alongside identity verification. Knowing who an agent is matters, but knowing who authorized it and with what scope is equally important for trust decisions in multi-agent workflows. The WG just ratified specs covering both ([Entity Verification](https://github.com/corpollc/qntm/issues/5), [Execution Attestation](https://github.com/corpollc/qntm/issues/6)).

msaleme:
Strong contribution. Identity verification is the foundation that makes everything else trustworthy — capability claims, security posture, even the `limitations` field discussion in #1694 all depend on knowing who is making the assertion.

Two things this connects to well:

**Attestation as the identity-capability bridge.** Once agents can prove identity, the next question is "what can I trust this agent to do safely?" We have been working on a [machine-readable attestation schema](https://github.com/msaleme/red-team-blue-team-agent-fabric/blob/main/schemas/attestation-report.json) that binds security test results to a verified identity. The JWS signing approach you reference here (RFC 7515) is the same primitive we use for attestation signing — the two could compose naturally.

**Practical gap today.** The `/.well-known/agent` convention some registries are adopting needs identity verification to be meaningful. Without it, anyone can claim a well-known endpoint. The cryptographic verification guidance here closes that gap.

This would pair well with the conformance testing discussion happening on the MCP side (modelcontextprotocol/specification#2484) — both are moving toward "prove it, do not just claim it" as a protocol-level expectation.

aeoess:
@msaleme — the attestation-as-bridge framing is exactly right. We shipped a 4-tier attestation model yesterday that connects to what you're building.

Our passport grades (0-3) are evidence-based posture scores: Grade 0 is a bare keypair. Grade 1 is issuer countersigned. Grade 2 is runtime-bound with challenge-response attestation from the infrastructure. Grade 3 is runtime plus verified human principal. The grade travels with the passport — any agent or registry can read it without understanding the scoring internals.

The composition point with your attestation report schema: your security test results (red team/blue team findings bound to a verified identity via JWS) could feed directly into our Tier 2 (infrastructure-attested) evidence. If a sandbox or test harness signs an attestation report for an agent, that's exactly the kind of signal our grade computation consumes. The signed report becomes evidence in the `IssuanceEvidenceRecord`, the grade reflects it, and downstream consumers (registries, other agents, A2A discovery) see the grade without parsing the raw report.

On the `/.well-known/agent` gap — agreed. Our `aps.txt` convention (analogous to robots.txt but signed) addresses the same problem for governance terms. Identity verification + attestation + well-known discovery are three layers of the same trust stack.

SDK: [agent-passport-system](https://www.npmjs.com/package/agent-passport-system) v1.29.1. The attestation types are in `src/types/attestation.ts` if you want to see the evidence model.

internet-dot:
The attestation-based identity verification approach here is solid. One gap this guidance (and the broader #1672 discussion) doesn't address: **how agents route to each other after identity verification succeeds**.

Verification proves *who* an agent is, but doesn't tell you *where* it is or *how to talk to it*. The Agent Card provides endpoint URLs, but those are bound to a specific deployment — if an agent moves between providers or exposes multiple protocol endpoints (A2A, MCP, REST), the card can't express that.

We addressed this in our HCS-14 standard by making identifiers **self-describing** — the UAID embeds routing metadata directly in the ID:

```
uaid:aid:{hash};proto=a2a;nativeId=agent.example.com;registry=example;domain=example.com
```

This means after verifying an agent's identity, you already know:
- **What protocol to use** (`proto=a2a`)
- **Where to route** (`nativeId` + `domain`)
- **What registry attests to its identity** (`registry`)

No separate lookup needed. The verified identity and the routing information are in the same string.

Additionally, HCS-14 encodes agent capabilities as numeric enums (core 0-19, protocol-specific 20-39) that are part of the deterministic hash. This addresses the skill introspection gap discussed in #1655 — an agent's capabilities are provably part of its identity, not just mutable metadata in the Agent Card.

Spec: https://hol.org/docs/standards/hcs-14

aeoess:
@internet-dot — routing after identity verification is a real gap. The question is whether to solve it by coupling routing into the identifier or by keeping them separate.

APS deliberately separates identity from routing. The Ed25519 public key is the stable identity. Routing is a mutable property — agents move between providers, expose new endpoints, go offline. If routing metadata is embedded in the identifier, the ID changes when the agent moves, and every delegation, receipt, and attestation referencing the old ID becomes orphaned.

Our approach: identity is the passport (stable, Ed25519). Discovery is the Agora (signed registry, queryable). Routing is the MCP endpoint or Agent Card (mutable, updatable without changing identity). An agent can move from `provider-a.com` to `provider-b.com` and keep the same passport, same delegation chain, same trust history. The delegation says "agent with public key X can do Y" — not "agent at endpoint Z can do Y."

That said, the capability encoding in HCS-14 is interesting. Capabilities as part of the identity hash means you can verify what an agent *claims* to do is cryptographically bound to its identity. We do something similar with scope declarations in delegation chains — the delegation specifies what capabilities are authorized, and `processToolCall()` enforces them at runtime. The difference: we treat capabilities as *granted authority* (what you're allowed to do), not *inherent properties* (what you can do). An agent might be capable of `web_search` but its delegation only authorizes `code_execution`.

The practical bridge between the two approaches: after attestation verification succeeds (proving who), query the agent's trust profile for routing + capability data. We shipped a public endpoint for exactly this — `GET /api/v1/public/trust/:agentId` returns grade, trust level, and capability signals without requiring the querier to be in the same organization. No separate registry lookup needed beyond the one you already make for verification.

internet-dot:
@aeoess — fair critique on the routing-as-mutable-property argument. You're right that if an agent moves providers, a UAID with baked-in `nativeId` would change.

Two clarifications on how HCS-14 actually handles this:

**1. Dual targeting.** The `uaid:aid:` path (deterministic, registry-generated) is designed for stable agents where the routing IS the identity (a registry agent doesn't move). But `uaid:did:` wraps an existing DID (Ed25519, did:web, whatever) and adds routing parameters on top. The DID stays stable. The routing params in `uaid:did:` are advisory metadata, not part of the hash. So:

```
uaid:did:ed25519:abc123;proto=a2a;nativeId=provider-a.com;registry=aps;domain=example.com
```

The Ed25519 key is the stable identity (same as your passport). The routing params (`nativeId=provider-a.com`) are updateable without changing the underlying DID. Delegations reference the DID, not the routing wrapper.

**2. Skills as claims vs granted authority.** You make a good distinction. HCS-14 skills in the `uaid:aid:` hash are *inherent capabilities* (what the agent can do). Your delegation scope is *granted authority* (what it's allowed to do). These are complementary, not conflicting. An agent's inherent capabilities can be verified independently of any delegation, which is useful for discovery before any trust relationship exists.

The practical bridge you describe (attestation verification → trust profile query → routing + capability data) is clean. The question is whether the routing data lives in the identifier or in a separate lookup. Different tradeoffs — identifier coupling vs query dependency. Both valid.

The public endpoint `GET /api/v1/public/trust/:agentId` is a nice pattern. Worth mentioning in the discovery docs as a post-verification routing approach.

internet-dot:
@aeoess — the orphaning concern is valid if routing is purely operational metadata. But HCS-14 takes the position that **routing changes can constitute an identity change**.

If an agent moves from `provider-a.com` to `provider-b.com`, the operator, infrastructure, security posture, and trust relationships all potentially change. A delegation chain tied to `provider-a.com` should not silently transfer to `provider-b.com` without re-attestation. Embedding routing in the ID makes this explicit — the ID changes, and downstream consumers know to re-verify.

That said, HCS-14 has two targeting modes precisely to handle the tension you describe:

- **`uaid:aid:`** — deterministic, generated from canonical fields (registry, name, version, protocol, nativeId, skills). Use this when the routing *is* part of the identity contract. If the agent moves, the ID changes, and that's intentional — it signals a different deployment with potentially different trust properties.

- **`uaid:did:`** — wraps an existing DID (Ed25519 key, `did:web`, `did:key`, anything). The underlying identity stays stable. The routing parameters (`proto`, `nativeId`, `domain`) are added on top but the DID itself doesn't change. This is the self-sovereign mode that preserves your delegation chains.

So the architecture supports both philosophies: mutable routing with identity continuity via `uaid:did:`, and routing-as-identity via `uaid:aid:`. The choice depends on whether you trust the agent to be the same entity across provider moves.

On the capabilities vs authority distinction — good call. HCS-14's skills encoding describes what an agent *can do* (inherent). Your delegation scopes describe what it's *allowed to do* (granted). These are complementary layers, not competing. An agent's `uaid:aid:` hash proves its inherent capabilities are part of its identity. Your APS delegation then restricts what it's authorized to exercise.

aeoess:
@internet-dot — the dual targeting clarification is useful. `uaid:did:` wrapping an existing Ed25519 key with mutable routing params on top is architecturally clean. The underlying identity stays stable, routing is advisory. That's the same separation we make, just expressed differently.

The capabilities-as-claims vs capabilities-as-authority distinction is worth keeping explicit. `uaid:aid:` skills in the hash prove what the agent CAN do. APS delegation scopes restrict what it's ALLOWED to do. These compose: discover via inherent capabilities, authorize via delegation. Neither replaces the other.

On the public trust endpoint as a post-verification routing approach: agreed, that's the intended use. Verify identity first (Ed25519 signature check), then query trust profile for grade + capability signals. Two calls, no coupling between identity and routing in the protocol layer.

PR #1750, opened 2026-04-15
fix: redact push notification credentials in read APIs

madankumarpichamuthu:
## Summary
- Clarifies that `GetTaskPushNotificationConfig` and `ListTaskPushNotificationConfigs` MUST redact `authentication.credentials`.
- Updates the proto contract and normative specification language to prevent credential leakage when reading stored push notification configs.
- Aligns supporting docs and changelog with the security behavior.

## Why
Issue #1681 reports that querying existing push notification configs can expose sensitive credentials. This change makes non-disclosure explicit and consistent across the spec and docs.

Fixes #1681

## Changes
- Updated `specification/a2a.proto`:
  - Added security notes on `GetTaskPushNotificationConfig` and `ListTaskPushNotificationConfigs`.
  - Documented `AuthenticationInfo.credentials` as write-only for read operations.
  - Added response-level redaction notes for list/config structures.
- Updated `docs/specification.md`:
  - Added MUST-level redaction requirements in operation behavior and security sections.
  - Added explicit redaction expectations in gRPC binding response sections.
- Updated `docs/topics/streaming-and-async.md`:
  - Clarified that config readback redacts `authentication.credentials`.
- Updated `docs/whats-new-v1.md`:
  - Added security clarification note for Get/List redaction.
- Updated `CHANGELOG.md`:
  - Added unreleased bug-fix entry for #1681.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased](https://github.com/a2aproject/A2A/compare/v1.0.0...HEAD)
+
+### Bug Fixes
+
+* **spec:** redact push notification credentials in `GetTaskPushNotificationConfig` and `ListTaskPushNotificationConfigs` responses ([#1681](https://github.com/a2aproject/A2A/issues/1681))
+
 ## [1.0.0](https://github.com/a2aproject/A2A/compare/v0.3.0...v1.0.0) (2026-03-12)
 
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -357,7 +357,7 @@ Retrieves an existing push notification configuration for a task.
 
 **Behavior:**
 
-The operation MUST return configuration details including webhook URL and notification settings. For security, the operation MUST redact `authentication.credentials` in the returned configuration. The operation MUST fail if the configuration does not exist or the client lacks access.
+The operation MUST return configuration details including webhook URL and notification settings. `authentication.credentials` MUST be omitted (absent) in the returned configuration. The operation MUST fail if the configuration does not exist or the client lacks access.
 
 #### 3.1.9. List Push Notification Configs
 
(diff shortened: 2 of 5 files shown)


0xbrainkid:
The write-only credential field with MUST-level redaction on reads is the correct pattern — credentials in a stored config should be treated like passwords: writable at setup, never readable back.

One extension worth adding to the security note: **agent identity binding on stored push notification configs.** Currently, a `GetTaskPushNotificationConfig` response redacts `authentication.credentials` but still returns the full config structure. If the config was created by Agent A and is later read by Agent B (both registered with the same server), Agent B learns the push notification endpoint and authentication scheme Agent A configured — even without the credentials.

For multi-tenant A2A deployments, the config readback should also enforce:
- The reading agent must be the same agent that created the config (identity ownership check)
- Or the reading agent must have explicit delegation from the creating agent

This is analogous to how the `AuthorizedSqlStore` pattern works in llama-stack — access checks on both read and write, with ownership preservation on update. The credential redaction fix closes the most critical leak, but adding agent identity ownership to the config access model closes the structural information leak alongside it.

Worth noting in the security section of the spec: "Config readback MUST redact credentials AND SHOULD enforce that the requesting agent identity matches the agent that created the configuration."

madankumarpichamuthu:
@0xbrainkid great point.. this pr is focused on credential redaction..
is there an issue already created for the point that you have raised?
then we can handle it in a different pr.

madankumarpichamuthu:
Thanks @0xbrainkid — I agree the agent-identity ownership check on config readback is a real structural concern, but it's a distinct access-control feature and out of scope for this PR, which is deliberately limited to credential redaction.

I don't see an existing issue tracking the ownership/delegation-check idea, so it'd be best raised as its own issue (or extension proposal) where the multi-tenant access model can be discussed on its own merits. Happy to link it here once one exists so the two are cross-referenced.

Separately: this PR has now been rebased onto the latest `main` (it was ~68 commits behind), the merge conflicts from the `PushNotificationConfig` → `TaskPushNotificationConfig` rename are resolved, and the wording is standardized on "MUST be omitted (absent)". protolint and markdownlint pass locally against the repo's linter configs.

PR #1789, opened 2026-04-25
feat(spec): publish per-type JSON Schema files for  referencing

### LINKED ISSUE #1781: Publish JSON Schema / OpenAPI bundle for Agent Card and core types
## Request

The A2A specification already notes that you *may publish an OpenAPI v3 + JSON Schema bundle for enhanced tooling* (alongside the canonical protobuf definition). This issue is a request to prioritize that.

## Why this matters

Agent manifests (e.g. `agent.json` files that describe an agent's identity, capabilities, and metadata) need to reference a `$schema` URL to validate their A2A-compatible fields. Without a published JSON Schema for the Agent Card, tooling authors are forced to either:

- Duplicate the type definitions manually (drift risk)
- Reference the TypeScript types or protobuf (not usable in JSON tooling)
- Invent their own schema (interoperability breaks)

A concrete example — an `agent.json` that wants to declare A2A-compatible top-level fields today cannot write:

```json
{
  "$schema": "https://spec.a2aprotocol.ai/schemas/agent-card/v1.json",
  "name": "My Agent",
  "description": "..."
}
```

because no such schema URL exists.

## Proposal

Publish a JSON Schema file (or OpenAPI v3 bundle) for at minimum:

- `AgentCard`
- `Task` and task state types
- `TaskStatusUpdateEvent` and `TaskArtifactUpdateEvent`

The protobuf remains the normative source. The JSON Schema can be generated from it or maintained as a derived artifact, with a clear note that the protobuf is authoritative in case of conflict.

A stable URL pattern like `https://spec.a2aprotocol.ai/schemas/{version}/{type}.json` would allow agent tooling to reference specific versions.

## Context

We are building an AI agent platform that implements A2A for agent discovery and dispatch. We want to use `$schema` in agent manifests to make the provenance of each field explicit — which standard it belongs to — but cannot do so without a published schema URL.

lawcontinue:
Closes #1781

## Problem

Agent manifests (`agent.json`) cannot reference a stable `$schema` URL for A2A types today. Tooling authors must duplicate type definitions manually or invent their own schema (interoperability breaks).

### `scripts/proto_to_json_schema.sh`
- Add optional second argument `split_dir` for per-type output
- Generates `{TypeName}.json` for 14 core types + all `Security*` sub-types
- Each file gets `$id: https://a2a-protocol.org/spec/{TypeName}.json`
- Backward compatible: no second argument = original behavior (bundle only)

### `scripts/build_docs.sh`
- Calls `proto_to_json_schema.sh` with `docs/spec` as split dir during build
- Individual schema files published alongside the bundle

### `docs/spec/index.md`
- Schema reference page listing all 15+ types with stable URLs
- Usage example showing `$schema` in agent manifests
- Regeneration instructions

### `mkdocs.yml`
- Add "JSON Schema Reference" to docs navigation

## Schema URLs (after deploy)

```
https://a2a-protocol.org/spec/a2a.json              (bundle)
https://a2a-protocol.org/spec/AgentCard.json         (per-type)
https://a2a-protocol.org/spec/Task.json
https://a2a-protocol.org/spec/Message.json
...
```

## Usage

```json
{
  "$schema": "https://a2a-protocol.org/spec/AgentCard.json",
  "name": "My Agent",
  "description": "A helpful agent"
}
```

## Notes

- Proto remains the normative source (schemas are derived artifacts)
- The `specification/json/README.md` already documents the build pipeline — this PR extends it


darrelmiller:
Can you help me understand why these schemas are useful? Do you have a reference to what agents.json  is?  Why would tooling authors not use the SDKs provided for A2A?  Json Schema definitions are incomplete definitions of the protocol.

lawcontinue:
Good questions. A couple of concrete use cases:

We have a Rust agent gateway that receives Agent Cards from external agents over HTTP — no official A2A Rust SDK exists, so we generate types from JSON Schema with `typify`. Without a published schema we're hand-maintaining structs that drift from proto every few weeks. Same situation for teams using Kotlin, Elixir, etc.

On the static file side: an `agent.json` served by `.well-known/` isn't running inside any SDK. Editors and CI linters (ajv-cli, VS Code validation) can check it against a `\$schema` URL without importing anything. The SDKs solve runtime consumption; schemas solve static validation and codegen for non-SDK consumers.

Agreed that JSON Schema can't express RPC semantics or streaming — that's why the PR notes proto remains normative. The schemas cover structural validation, which is a different (and cheaper) gate than protocol compliance.

On `agents.json` — typo in the original issue, should be `agent.json` (the agent manifest per spec).

PR #1793, opened 2026-04-27
docs(fix): align push notification auth examples with protocol behavior

### LINKED ISSUE #1594: [Bug]: Clarify the name & usage of the X-A2A-Notification-Token header
### What happened?

This HTTP Header is used in the [Push Notification Setup and Usage](https://a2a-protocol.org/latest/specification/#66-push-notification-setup-and-usage) section without much context. 

From the example, it seems obvious that it corresponds to the `configuration.pushNotificationConfig.token` sent by the user with the message but it might be worth calling out.

I also wonder if this header should be named `X-A2A-Notification-Token ` or `A2A-Notification-Token`.
I am working on updating the other HTTP headers names for the Java SDK (https://github.com/a2aproject/a2a-java/pull/709) and it seem inconsistent that this one continues to use the `X-`.
Could you confirm if the current name is still correct? thanks

### Relevant log output

```shell

```

### Code of Conduct

jmesnil:
The `authentication` field in the SendMessage's configuration has 2 String fields:

* `scheme`
* `credentials`

The request POSTed to the Webhook pass this authentication info in the `Authorization` HTTP header. The `X-A2A-Notification-Token` header is not required (and is not defined in the spec).

This fixes #1594.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -319,7 +319,7 @@ Creates a push notification configuration for a task to receive asynchronous upd
 
 **Inputs:**
 
-{{ proto_to_table("CreateTaskPushNotificationConfig") }}
+{{ proto_to_table("TaskPushNotificationConfig") }}
 
 **Outputs:**
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1640,11 +1640,11 @@ Authorization: Bearer token
     "messageId": "6dbc13b5-bd57-4c2b-b503-24e381b6c8d6"
   },
   "configuration": {
-    "pushNotificationConfig": {
+    "taskPushNotificationConfig": {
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -319,7 +319,7 @@ Creates a push notification configuration for a task to receive asynchronous upd
 
 **Inputs:**
 
-{{ proto_to_table("CreateTaskPushNotificationConfigRequest") }}
+{{ proto_to_table("CreateTaskPushNotificationConfig") }}
```
mikeas1:
```suggestion
{{ proto_to_table("TaskPushNotificationConfig") }}
```

Correct name is TaskPushNotificationConfig

darrelmiller (approved):

msampathkumar (approved):

PR #1810, opened 2026-05-01
feat: add generation field to Task for ordering, long-polling, and optimistic concurrency

### LINKED ISSUE #1381: [Feat]: Method for blocking on Task updates

There are two cases where clients receive non-terminal Tasks in a non-streaming context:

- When using non-blocking `SendMessage`
- When using blocking `SendMessage` and the Task enters an interrupted (`input-required`, `auth-required`) state

Clients must then poll the Task using `GetTask` to observe updates. If the agent and client both support it, clients may also subscribe to the Task and receive a stream of updates.

However, there is no means of performing a blocking `GetTask` that waits for a Task to reach another terminal or interrupted state. This is not an  issue for tasks in `input-required` state, since the client must send a message with the requested input anyway. It can be awkward for `auth-required`; see below.

Viewed another way, we currently have 3 ways of getting updates on a Task when sending a message to an agent:
- Non-blocking (`SendMessage(blocking=False)`)
- Blocking (`SendMessage(blocking=True)`)
- Streaming (`SendStreamingMessage`)

But we only have two ways of getting updates on a Task _without_ sending a message:
- Non-blocking (`GetTask`)
- Streaming (`SubscribeToTask`)

Should we add a blocking case for receiving Task updates?

Two possible approaches to this:

1. Extend `GetTaskRequest` to include a `blocking` field. When a server receives a `GetTask` call with this field set to true, it does not return until the `Task` enters a terminal or interrupted state. The exact same semantics as `SendMessage(blocking=True)` has.
2. Introduce a new RPC: `WaitForTask`. Same semantics as above, just separated to a new protocol method.

I have slight preference for the first, but the benefit of the second is that it clearly separates out functionality that an agent may not implement. 

`SendMessage` with empty `parts` field. We could allow servers to interpret a SendMessage request that specifies a `task_id` but includes no content as the equivalent of a `GetTask` request. The same applies for SendStreamingMessage. We could even use this to completely eliminate `SubscribeToTask` and `GetTask` methods, but that would be very unintuitive.

### Additional context

The primary case where this becomes obvious is when a Task enters `auth-required` state. Authorization can be provided out-of-band, so from a client's perspective the point when a Task begins receiving updates again is not clear. The best thing for a client to do upon receiving a Task in `auth-required` state is to immediately begin polling the Task or subscribe to updates. This where the lack of a blocking get method shows: it feels incongruous to perform a blocking send, but then have to do non-blocking polling on the task.

Example pseudocode:

```python
task = await client.send_message("Please update my calendar to cancel all TSC meetings", blocking=True)
if task.status.state == 'auth-required':
  print(f'Do the auth stuff please: {task.status.message}')
  # This blocks until the task changes state. Without this, clients must perform
  # a polling loop to watch for task state changes.
  task = await client.get_task(task.id, blocking=True)
if task.status.state == 'complete':
  print('I did it!')
else:
  print('Something went terribly wrong')
```

### Code of Conduct

### LINKED ISSUE #1794: [Feat]: Add generation/version number field to Task for event ordering, long-polling, and optimistic concurrency
## Is your feature request related to a problem? Please describe.

Currently, `TaskStatusUpdateEvent` and `TaskArtifactUpdateEvent` carry no inherent sequence information. Clients receiving a stream of these events (or receiving them out-of-order via push notifications) have no reliable way to:

- Determine whether they have missed an event.
- Establish a total ordering of events without relying on wall-clock timestamps (which are unreliable for ordering in distributed systems).
- Efficiently long-poll for changes — clients must either poll unconditionally or rely on coarse-grained timestamps.
- Safely send a follow-up `SendMessage` to a `Task` without risking a lost-update race condition (e.g. two clients both reading generation N and then both sending a message that should only apply once).

## Describe the solution you'd like

Add a `generation` (or `version`) field — a monotonically increasing integer — to the `Task` message. The generation is initialised to `0` (or `1`) when the task is created and **incremented by the server on every state-changing mutation**: a `TaskState` transition or the addition/update of an `Artifact`.

### Schema changes (proto)

```proto
message Task {
  // ... existing fields ...

  // Monotonically increasing generation counter, incremented on every
  // TaskState or Artifact change. Starts at 0 when the task is created.
  // May be used for event ordering, long-polling, and optimistic concurrency.
  int64 generation = 7;
}
```

### 1 — Event ordering / missed-event detection

Include the **post-mutation** `generation` value in both update event types:

```proto
message TaskStatusUpdateEvent {
  // ... existing fields ...
  int64 generation = 5;
}

message TaskArtifactUpdateEvent {
  // ... existing fields ...
  int64 generation = 7;
}
```

A client that receives an event with `generation = N+2` when it last saw `generation = N` knows it missed at least one event and can react accordingly (e.g. re-fetch the full `Task` via `GetTask`).

### 2 — Long-polling support

Extend `GetTaskRequest` with an optional `current_generation` parameter:

```proto
message GetTaskRequest {
  // ... existing fields ...

  // If set, the server SHOULD NOT return until the task's generation
  // is strictly greater than this value, or a server-defined timeout
  // elapses. Enables efficient long-polling without repeated unconditional
  // requests.
  optional int64 current_generation = 4;
}
```

The server returns immediately if `task.generation > current_generation`, otherwise it holds the request open until the generation advances (or the timeout fires, returning the current state). This replaces the need for clients to poll `GetTask` in a tight loop.

### 3 — Optimistic concurrency / lost-update prevention

Extend `SendMessageConfiguration` (used by `SendMessageRequest`) with an optional `if_generation_match` field:

```proto
message SendMessageConfiguration {
  // ... existing fields ...

  // If set, the server MUST reject the request with a suitable error
  // (e.g. HTTP 412 Precondition Failed / gRPC ABORTED) if the task's
  // current generation does not equal this value at the time the message
  // is processed. Allows clients to implement optimistic concurrency
  // and avoid lost-update races.
  optional int64 if_generation_match = N;
}
```

This mirrors the well-established ETag / `If-Match` pattern from HTTP and the `generation` precondition used in Google Cloud Storage and similar APIs.

## Describe alternatives you've considered

- **Wall-clock timestamps** — already present on `TaskStatus`, but unsuitable for reliable ordering in distributed systems due to clock skew.
- **Client-side sequence numbers per stream** — only works for a single streaming session; does not survive reconnects or help non-streaming (push-notification) clients.
- **Versioned resource URLs / ETags** — a pure HTTP-layer ETag could address the lost-update problem but would not help with event ordering or long-polling in a protocol-agnostic way.

## Additional context

The `generation` concept is well established in distributed systems and existing APIs:
- **Kubernetes** uses `resourceVersion` on every resource for exactly these purposes.
- **Google Cloud Storage** exposes a `generation` (object version) and supports `ifGenerationMatch` preconditions on mutations.
- **HTTP ETags + `If-Match`** (RFC 9110) provide the same lost-update semantics at the HTTP layer.

Keeping `generation` as a simple `int64` (rather than an opaque string like `resourceVersion`) makes the ordering and long-polling semantics explicit and easy to implement on both client and server.

Tehsmash:
## Summary

Adds a monotonically increasing `generation` field to `Task` (and related messages) that supports three new capabilities:

### 1. Event ordering / missed-event detection
`TaskStatusUpdateEvent` and `TaskArtifactUpdateEvent` now carry the task's `generation` value after the mutation that produced them. A gap in generation values (e.g. jumping from `3` to `5`) tells a client it missed at least one event and should re-fetch the full task state.

### 2. Long-polling via `GetTaskRequest.current_generation`
If `currentGeneration` is supplied on a `GetTask` call the server **SHOULD** hold the request open until the task's generation advances beyond that value, returning immediately if it already has. This replaces tight polling loops with a single blocking call.

### 3. Optimistic concurrency via `SendMessageConfiguration.if_generation_match`
If `ifGenerationMatch` is set the server **MUST** reject the request with HTTP `412 Precondition Failed` / gRPC `ABORTED` (`TaskGenerationMismatchError`) if the task's generation has changed since the client last read it, preventing lost-update races in multi-client scenarios.

## Proto changes

| Message | New field |
|---|---|
| `Task` | `int64 generation = 7` |
| `TaskStatusUpdateEvent` | `int64 generation = 5` |
| `TaskArtifactUpdateEvent` | `int64 generation = 7` |
| `GetTaskRequest` | `optional int64 current_generation = 4` |
| `SendMessageConfiguration` | `optional int64 if_generation_match = 5` |

## Spec documentation changes

- **New §3.2.7 Task Generation Semantics** — comprehensive description of the `generation` field, event ordering, long-polling behaviour, optimistic concurrency, and backwards compatibility.
- **§3.1.3 Get Task** — cross-reference to §3.2.7 for `currentGeneration` long-polling.
- **§3.2.2 SendMessageConfiguration** — document `ifGenerationMatch` optimistic concurrency behaviour and the new error.
- **§3.3.2 Error Handling** — add `TaskGenerationMismatchError` to the A2A-Specific Errors table.
- **§3.5.1 Overview of Update Mechanisms** — add Long-Polling as a delivery mechanism.
- **§3.5.2 Streaming Event Delivery** — note generation-based event ordering and missed-event detection.
- **§5.4 Error Code Mappings** — add `TaskGenerationMismatchError` (JSON-RPC `-32010` / gRPC `ABORTED` / HTTP `412`).
- **§11.5 Query Parameter Naming** — add `currentGeneration` to the mapping table and example.
- **§11.6 Error Handling** — add HTTP `412` generation mismatch error response example including `ifGenerationMatch` body mapping.

## Backwards compatibility

`generation` defaults to `0` in proto3. Existing clients that do not read or send `generation` continue to interoperate without change. Clients **SHOULD** treat a server-returned `generation` of `0` as "generation unknown" and **MUST NOT** use it for ordering, long-polling, or precondition checks against a server that does not implement this feature.

Fixes #1794
Fixes #1381


specification/a2a.proto
```diff
@@ -181,6 +187,11 @@ message Task {
   // protolint:enable REPEATED_FIELD_NAMES_PLURALIZED
   // A key/value object to store custom metadata about a task.
   google.protobuf.Struct metadata = 6;
+  // Monotonically increasing counter incremented by the server on every
+  // state-changing mutation to this task (TaskState transition or
+  // Artifact addition/update). Starts at 0 when the task is created.
```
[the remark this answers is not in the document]
herczyn: +1

docs/specification.md
```diff
@@ -487,6 +495,51 @@ A key-value map for passing horizontally applicable context or parameters with c
 
 As service parameter names MAY need to co-exist with other parameters defined by the underlying transport protocol or infrastructure, all service parameters defined by this specification will be prefixed with `a2a-`.
 
+#### 3.2.7. Task Generation Semantics
+
+<span id="327-task-generation-semantics"></span>
+
+The `generation` field on [`Task`](#411-task) is a monotonically increasing integer maintained by the server. It is initialised to `0` when a task is created and **MUST** be incremented by the server on every state-changing mutation:
```
herczyn: monotonically -> sequentially. increasing is always monotonic but not always sequential which is important here. in line 511 we mention that this is assumed on the client, but we never outright say that the server should act in that way, i would make is as explicit as possible, something like:

Events produced by the server **MUST** have sequential numbering, meaning each change in the generation of a Task **MUST** by an increase by one and must map 1:1 to a produced Event.

msampathkumar: A what-if question. Along with `generation`, can we have `generationIDs` ? It would be a list of integers.

docs/specification.md
```diff
@@ -487,6 +495,51 @@ A key-value map for passing horizontally applicable context or parameters with c
...
+
+The server returns when the task's `generation` exceeds `3`, or on timeout.
+
+**Optimistic Concurrency:**
+
+The `ifGenerationMatch` field in [`SendMessageConfiguration`](#322-sendmessageconfiguration) provides an optimistic concurrency guard for multi-turn interactions. See that section for details.
+
+**Backwards Compatibility:**
+
+The `generation` field was introduced in version **1.1** of this specification. The field defaults to `0` in the Protocol Buffer encoding (proto3 default for `int64`). Clients that do not read or send `generation` continue to interoperate correctly with servers that support it. Servers that have not yet implemented `generation` will return `0` on all events and responses; clients will detect this via the duplicate-generation rule above and gracefully fall back.
```
msampathkumar: I don't think we have 1.1 but may be we should have :)

Latest Released Version [1.0.0](https://a2a-protocol.org/v1.0.0/specification)

Tehsmash: The current release is 1.0.0 and we're not going back to change that, this change would only be included if we cut a 1.1 because it involves a proto change.

darrelmiller: We should start planning for that.

msampathkumar:
Added a git commit to fix the merge conflict.

For `TaskNotCancelableError` error state, the terminal states are updated.

msampathkumar:
Should this be part of 1.1 or 2.0 ? (vote yes for 1.1)

/vote

spetschulatSFDC (approved):

darrelmiller (approved):

specification/a2a.proto
herczyn, line 292: should we add start/end generation as dicussed?

Tehsmash: Should it be in this PR or a follow up?

muscariello (approved):
LGTM

PR #1848, opened 2026-05-13
docs: add multi-tenancy guide and clarify tenant field semantics

### LINKED ISSUE #1820: [Bug]: missing documentation on `Tenant`
### What happened?

In https://a2a-protocol.org/latest/definitions/ there is a mention of Tenants but there is no official document to learn about Tenants.

aNEW: Multi-Tenancy Support @ https://a2a-protocol.org/latest/whats-new-v1/#push-notification-operations

### Relevant log output

```shell

```

### Code of Conduct

Tehsmash:
## Summary

- Adds a new **Multi-Tenancy** documentation topic (`docs/topics/multi-tenancy.md`) covering how to serve multiple agents behind a single A2A endpoint, with examples of URL-based, header-based, and body-based (`tenant` field) routing strategies.
- Clarifies the `tenant` field in `a2a.proto` — replaces the binding-specific phrase "provided as a path parameter" with accurate, binding-agnostic wording describing it as an opaque routing identifier.
- Adds a normative client rule in spec section 8.3.2: if the selected `AgentInterface` has a non-empty `tenant` field, the client **MUST** include that value in the `tenant` field of every request message body sent to that interface.
- Adds the new topic to the MkDocs navigation.

Fixes #1690
Fixes #1820

### DIFF
--- a/docs/topics/multi-tenancy.md
+++ b/docs/topics/multi-tenancy.md
@@ -9,7 +9,7 @@ when a routing identifier is advertised in an Agent Card.
 ## Overview
 
 A common deployment pattern is to place several agents behind a single host or
-reverse-proxy. From the outside the agents are reachable at the same base URL,
+reverse-proxy. From the outside the agents are reachable at the same domain,
 but each individual agent needs to be distinguished so that requests are
 delivered to the right backend.
 
@@ -21,8 +21,9 @@ Each agent is assigned a distinct URL prefix. The Agent Card for each agent
 advertises its own `url` in `supportedInterfaces`, so clients automatically
 send requests to the correct path.
 
+Agent Card for the "billing" agent:
+
 ```json
-// Agent Card for the "billing" agent
 {
   "name": "Billing Agent",
   "supportedInterfaces": [
@@ -35,8 +36,9 @@ send requests to the correct path.
 }
 ```
 
+Agent Card for the "support" agent:
+
 ```json
-// Agent Card for the "support" agent
 {
   "name": "Support Agent",
   "supportedInterfaces": [
@@ -50,9 +52,8 @@ send requests to the correct path.
 ```
 
 The gateway or reverse-proxy routes `/billing/*` and `/support/*` to the
-appropriate backend without any changes to the A2A protocol messages themselves.
-This is the simplest approach and requires no special client awareness beyond
-reading the Agent Card.
+appropriate backend. This is the simplest approach and requires no special
+client awareness beyond reading the Agent Card.
 
 ### 2. Header-Based Routing
 
@@ -62,7 +63,7 @@ the standard `Authorization` header.
 
 Examples:
 
-- A bearer token whose claims identify the target agent or organisation.
+- A bearer token whose claims identify the target agent or organization.
 - An API key that maps to a particular tenant in the gateway's configuration.
 - A proprietary header such as `X-Agent-ID: billing`.
 
@@ -72,6 +73,13 @@ gateway resolves the target based on the header value. Authentication headers
 are already declared in the Agent Card's `securitySchemes` and `security`
 fields; custom routing headers can be documented as an extension or as
 operator-specific configuration.
+
+!!! note
+    Header-based routing requirements are not discoverable from the Agent Card
+    alone. Clients must obtain the required header values through out-of-band
+    means such as operator documentation or an OAuth token issued during
+    authentication. When routing information needs to be communicated to clients
+    via the Agent Card, prefer the `tenant` field approach described below.
 
 ### 3. Body-Based Routing Using the `tenant` Field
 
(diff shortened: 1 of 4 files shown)


docs/topics/multi-tenancy.md
```diff
@@ -0,0 +1,125 @@
...
+A single A2A endpoint can serve multiple agents or tenants. The A2A protocol
+does not prescribe a specific routing implementation — operators are free to
+choose the approach that best fits their infrastructure. This document describes
+the routing mechanisms the protocol supports and the rules clients must follow
+when a routing identifier is advertised in an Agent Card.
+
+## Overview
+
+A common deployment pattern is to place several agents behind a single host or
+reverse-proxy. From the outside the agents are reachable at the same base URL,
```
herczyn: base URL -> domain

docs/topics/multi-tenancy.md
```diff
@@ -0,0 +1,125 @@
...
+      "url": "https://agents.example.com/support",
+      "protocolBinding": "HTTP+JSON",
+      "protocolVersion": "1.0"
+    }
+  ]
+}
+```
+
+The gateway or reverse-proxy routes `/billing/*` and `/support/*` to the
+appropriate backend without any changes to the A2A protocol messages themselves.
```
herczyn: "without any changes to the A2A protocol messages themselves" let's remove it, it might put the agent name in the tenant field for instance

docs/topics/multi-tenancy.md
```diff
@@ -0,0 +1,125 @@
...
+base URL), the gateway can route on a custom HTTP header or on the contents of
+the standard `Authorization` header.
+
+Examples:
+
+- A bearer token whose claims identify the target agent or organisation.
+- An API key that maps to a particular tenant in the gateway's configuration.
+- A proprietary header such as `X-Agent-ID: billing`.
+
+Header-based routing is transparent to the A2A protocol — the agent's
```
herczyn: this made me pause. are we saying that these agents can only be consumed by a caller that already knows about the header? if it's auth - fine, but if it's something special shouldn't we point to the tenant field instead so that it could be consumed by any a2a client?
and if we're going with extension - why use header instead of metadata?

docs/topics/multi-tenancy.md
```diff
@@ -0,0 +1,125 @@
...
+      "url": "https://agents.example.com/a2a",
+      "protocolBinding": "HTTP+JSON",
+      "protocolVersion": "1.0",
+      "tenant": "billing"
+    }
+  ]
+}
+```
+
+**Client requirement**: If a selected `AgentInterface` entry has a non-empty
```
herczyn: i would even make this stronger, the tenant field MUST be set according to the agent interface definition.
that way we are covering the empty case as well: if the card says empty you should send empty

docs/topics/multi-tenancy.md
Tehsmash: Thats a good point, but I'm envisioning something like a company that puts an `X-Project: abcd1234` header in every request to their services, whether thats A2A or not, to determine which project the request is scoped to. This would just be part of the company's API documentation, but this isn't really discoverable for Agents with just the AgentCard...

You might even have multiple headers i.e. `X-Project:...` and `X-Agent-ID:...` so just pushing to use the tenant field doesn't seem the right fit, unless you do some encoding.
It would also break existing API gateway level routing that is using the headers as it would now be body based routing.

I don't know if there is an easy way to describe this through the AgentCard today unless we also put a "headers" map in the AgentInterface and have an AgentCard per Agent per Project. This is fine as long as the headers aren't dynamic in some way.

specification/a2a.proto
```diff
@@ -462,7 +467,8 @@ message AgentCardSignature {
 
 // A container associating a push notification configuration with a specific task.
 message TaskPushNotificationConfig {
-  // Optional. Tenant ID.
+  // Optional. Opaque routing identifier matching the `tenant` from the
```
herczyn: nit: should be same comment as below "// Optional. Opaque routing identifier. Must match the `tenant` value from"

herczyn (approved):

docs/topics/multi-tenancy.md
Tehsmash: added a NOTE covering the discovery of these headers, if anyone runs into this problem when they start having agents auto discover other agents from other orgs, then perhaps we can solve it then. You could just solve it by putting an API doc link in the AgentCard Description and let the client agent read the docs :D

herczyn: i think once we put it here we take responsibility for it and we should follow with headers in AgentInterface. if we don't want to we should not talk about it here and just stick to the auth routing which can be expressed in the AgentCard.
tenant is a string so it can contain anything, including a map :)

msampathkumar (approved):
LGTM.

mikeas1 (approved):

darrelmiller (approved):

docs/topics/multi-tenancy.md
```diff
@@ -0,0 +1,134 @@
...
+
+When sub-paths are not convenient (for example, when all agents share the same
+base URL), the gateway can route on a custom HTTP header or on the contents of
+the standard `Authorization` header.
+
+Examples:
+
+- A bearer token whose claims identify the target agent or organization.
+- An API key that maps to a particular tenant in the gateway's configuration.
+- A proprietary header such as `X-Agent-ID: billing`.
```
darrelmiller: FYI, X- headers are evil.

docs/topics/multi-tenancy.md
darrelmiller: Based on discussion in the A2A meeting, we are going to focus this paragraph on the use of Auth headers for routing, but we are going to remove the custom headers routing.

docs/topics/multi-tenancy.md
msampathkumar: I too share this idea on X-headers :)

Tehsmash: 🥇

docs/topics/multi-tenancy.md
Tehsmash: Updated, please take a look.

Issue #228, opened 2025-04-18
The semantics of Artifact updates are underspecified in the technical documentation


aickin:
I read through the documentation about artifact updates in the [Streaming Support](https://google.github.io/A2A/#/documentation?id=streaming-support) section, and it feels like this section needs to be clearer about how updates happen. Here are some questions I have that I don't think are currently answered:

1. First off, are names for Artifacts unique in the scope of a Task? Are indexes unique? Is it allowed in the spec to have two Artifacts in a Task with the same name but different indexes? Conversely, is it allowed in the spec to have two Artifacts in a Task with different names but the same index?
2. Assuming that index is unique in the scope of a Task, what do you do when a TaskArtifactUpdateEvent sends a new `name` for the Artifact? Does it overwrite the previous name? Is it an error? Or is it ignored? Similarly, is a new `description` for the Artifact an overwrite/error/ignored?
3. When you send metadata in a TaskArtifactUpdateEvent, does it replace the existing metadata for the Artifact, or is it merged into the existing metadata for the Artifact on a field-by-field basis, or is it an error, or is it ignored?
4. How do `append` and `lastChunk` work in practice? A few examples that I'm unsure about:
    1. Let's say you send one update with `append: false` and `lastChunk: false`, and then send a second update again with `append: false` and `lastChunk: false`. Intuitively, I would think that this wipes out and overwrites the first update. Is that correct, or is this an error?
    2. Now, let's say you send several updates in a row with `append: true` and `lastChunk: false`. Then you send an update with `append: true` and `lastChunk: true`. Finally you send an update with `append: true` and `lastChunk:false`. Is that an error?
    3. Similar to the last one,  let's say you send several updates in a row with `append: true` and `lastChunk: false`. Then you send an update with `append: true` and `lastChunk: true`. Finally you send an update with `append: false` and `lastChunk:false`. Is that an error or does it wipe out all the previous updates?

I know that some of these may seem like nitpicks, but I tend to think a spec should be really clear about error cases and state transitions. I think in it's current state, a reasonable engineer could assume several different ways this artifact update mechanism works.

holtskinner:
The docs have been updated, do they address your concerns? https://google-a2a.github.io/A2A/latest/topics/streaming-and-async/

Issue #64, opened 2025-04-10
Where is the link to the community channel? and can we setup a slack discussion page?


puchengy:
On README, it says "Have questions? Join our community in GitHub discussions.". Can you please properly link the place that we can have discussion? Currently I am not able to raise discussion unless I create an issue ticket.

DJ-os:
Please use https://github.com/google/A2A/discussions/ to start discussions.

PR #1882, opened 2026-05-26
feat: Add new spec for A2A protocol conformance tests

darrelmiller:
Summary

  This PR introduces ACTS (A2A Conformance Test Specification) — a unified YAML format for A2A protocol conformance tests,plus 111 tests covering all testable MUST/SHOULD requirements from the A2A v1.0 spec.

  Problem

  The A2A ecosystem has 4 fragmented conformance testing efforts (a2a-tck, a2a-itk, agntcy/csit, agentbin), each withdifferent test definitions. This makes it impossible to verify that SDKs passing "conformance tests" can actuallyinteroperate.

  Solution

  One canonical test format that all SDKs test against. Language-agnostic YAML declarations + freedom to implement runners in any language = guaranteed interoperability.

  What's Included

  📋 Specification (docs/acts-specification.md)

   - ~2000 lines, 17 sections, inline CDDL grammar (RFC 8610 conformant)
   - Abstract operations (transport-agnostic: JSON-RPC, gRPC, REST)
   - Rich assertion DSL, streaming support, state machine validation
   - Standard JSON report format for dashboards

  ✅ Test Suite (111 tests across 14 files)

   - 72 MUST, 35 SHOULD, 4 MAY tests
   - Core ops, discovery, streaming, errors, multi-turn, auth, push notifications, history, polling, wire format, data types, version negotiation, transport bindings, client parsing
   - Full coverage: All 76 testable spec requirements mapped

  🎨 HTML Viewer (tests/acts/test-viewer.html)

   - Interactive browser with search, filter, syntax highlighting
   - 290 KB standalone file, zero dependencies

  Coverage Validation

  ✅ Spec audit: All 76 testable MUST/SHOULD requirements have tests
  ✅ Gap analysis: All scenarios from 4 existing repos covered
  ✅ CDDL validation: All files validate against RFC 8610 grammar

### DIFF
--- a/tests/acts/push-notifications.acts.yaml
+++ b/tests/acts/push-notifications.acts.yaml
@@ -194,3 +194,261 @@ suites:
                 url: "{{webhookUrl}}"
             expect_error:
               code: UnsupportedOperationError
+
+      - id: PUSH-LIST-001
+        name: "List push configs"
+        level: may
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-complete-task"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-complete-task for push"
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+            expect:
+              status: 200
+              body:
+                id:
+                  type: string
+            capture:
+              configId: "id"
+
+          - id: list-configs
+            operation: list_push_notification_configs
+            params:
+              taskId: "{{create.taskId}}"
+            expect:
+              status: 200
+              body:
+                configs:
+                  type: array
+                  count_gte: 1
+
+      - id: PUSH-ERR-001
+        name: "Get nonexistent push config returns error"
+        level: may
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-complete-task"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-complete-task for push"
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: get-config
+            operation: get_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              id: "00000000-0000-0000-0000-000000000000"
+            expect_error:
+              code: TaskNotFoundError
+
+      - id: PUSH-IDEM-001
+        name: "Delete push config is idempotent"
+        level: may
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-complete-task"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-complete-task for push"
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+            expect:
+              status: 200
+              body:
+                id:
+                  type: string
+            capture:
+              configId: "id"
+
+          - id: delete-config
+            operation: delete_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              id: "{{set-config.configId}}"
+            expect:
+              status: 200
+
+          - id: delete-config-again
+            operation: delete_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              id: "{{set-config.configId}}"
+            expect:
+              status: 200
+
+      - id: PUSH-DELIV-001
+        name: "Push delivery includes auth"
+        description: "Runner MUST set up a webhook endpoint and verify auth credentials are included."
+        level: may
+        tags: [push, delivery, runner-special]
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-long-running"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-long-running start"
+              configuration:
+                returnImmediately: true
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+                authentication:
+                  scheme: "Bearer"
+                  credentials: "test-token"
+            expect:
+              status: 200
+              body:
+                id:
+                  type: string
+
+      - id: PUSH-DELIV-002
+        name: "Push delivery is at-least-once"
+        description: "Runner MUST set up a webhook endpoint and verify at-least-once delivery semantics."
+        level: may
+        tags: [push, delivery, runner-special]
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-long-running"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-long-running start"
+              configuration:
+                returnImmediately: true
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+            expect:
+              status: 200
+              body:
+                id:
+                  type: string
+
+      - id: PUSH-DELIV-003
+        name: "Push payload format correct"
+        description: "Runner MUST set up a webhook endpoint and verify the delivered payload format matches the A2A specification."
+        level: may
+        tags: [push, delivery, runner-special]
+        preconditions:
+          capabilities:
+            pushNotifications: true
+        requires_behaviors: ["tck-long-running"]
+        steps:
+          - id: create
+            operation: send_message
+            params:
+              message:
+                role: ROLE_USER
+                parts:
+                  - text: "tck-long-running start"
+              configuration:
+                returnImmediately: true
+            expect:
+              status: 200
+              body:
+                task:
+                  id:
+                    type: string
+            capture:
+              taskId: "task.id"
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+            expect:
+              status: 200
+              body:
+                id:
+                  type: string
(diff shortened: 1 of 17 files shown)


andysalvo:
ACTS scopes protocol conformance cleanly. One gap worth flagging: two implementations can pass all 111 tests and still produce different cryptographic outputs from the same input — JCS canonicalization, hash derivation, and signature binding sit below the transport layer ACTS validates. We hit this recently when our own JCS serializer diverged on Unicode handling that six other implementations got right; the shared test vectors caught it, not protocol-level tests. A derivation conformance layer would compose well as a separate test class or companion suite.

MoltyCel:
test (edit later)

chopmob-cloud:
We have an adversarial bench live -- 138 profiles across 30 categories, tested against our own A2A agent at `api.algovoi.co.uk`. It sits above what ACTS validates: ACTS checks protocol compliance against MUST/SHOULD requirements, the bench tests agent behaviour under adversarial conditions (prompt injection, coercive payment patterns, wallet draining attempts, malicious tool call sequences). The two layers complement without overlapping.

On @andysalvo's point about JCS and cryptographic outputs: that layer is below protocol-observable behaviour and cannot be reached by protocol assertions -- it needs shared test vectors that implementations run locally and compare byte-for-byte. A derivation conformance suite would compose alongside ACTS as a separate test class rather than an extension of it.

Profile schema is public if it is useful for expressing an adversarial class in the ACTS format.

---

chopmob-cloud:
The gap is real and sits below what ACTS validates. Two implementations can pass all 111 ACTS tests and still diverge on JCS canonical output for the same input — particularly on unicode normalisation, number representation, and key ordering edge cases. ACTS validates the transport and protocol layer; it does not validate the cryptographic substrate underneath.

The 8-implementation cross-validation corpus at \chopmob-cloud/algovoi-jcs-conformance-vectors\ covers exactly this layer: Python (rfc8785), TypeScript (canonicalize), Go (gowebpki/jcs), Rust (serde_json), Java (cyberphone/json-canonicalization), PHP, .NET, Ruby — all producing byte-identical output across the same vector set, including the non-ASCII UTF-8 cases that diverge most frequently in practice.

A complete A2A conformance spec needs both layers: ACTS for protocol, a JCS byte-match suite for the cryptographic substrate. The two are complementary rather than overlapping.

elang2:
Solid work here. I have a few observations after reading through the spec and test structure.

CDDL (RFC 8610) for test declarations makes sense for a multi-language ecosystem. But CDDL is a data definition language, not a behavioral one. You mention "state machine validation" for streaming tests, and I'm curious how temporal ordering constraints are actually expressed. If it's just an ordered sequence of expected events in the YAML, that works, but the spec should call out how runners distinguish "messages arrived in correct order" from "messages are individually well-formed." Those are different guarantees.

The transport abstraction is the right call since A2A is transport-agnostic in theory. In practice though, do any of the tests exercise transport-specific edge cases? Things like HTTP/2 stream resets, gRPC deadline propagation, WebSocket close frames. If not, worth noting that gap explicitly so nobody assumes transport conformance is covered.

One more thing. You state all "testable" requirements are mapped. What's the methodology for deciding something isn't testable? Some MUST NOT requirements are hard to test positively but straightforward to test via negative assertions (send bad input, verify rejection). Listing the excluded requirements would let reviewers push back on that classification where they disagree.

The HTML viewer is a nice touch for adoption. A machine-readable coverage matrix (requirement ID to test ID) would also help SDK maintainers wire this into CI.

astrogilda:
Tried to express a canonicalization corpus in this format and could not, and the reason is structural, not a missing field.

A test requires `steps`, and `step` is a closed union of four kinds. A server step needs an operation from a closed RPC enum and a live system under test. A client step tests the parser, and `expect_parsed` is keyed per field, so it can never assert a whole byte string. A raw step needs a method and a path. An assertion step needs a prior step. None of the four applies a pure function to a local input and compares output bytes, which is the shape every canonicalization vector takes: f(input) == bytes, with no system under test anywhere in it.

Two cases make it concrete. One of our vectors turns on whether an e-acute arrives as the literal character or as its six-character JSON escape, a distinction destroyed the moment a document is parsed, and every step kind parses first. Another turns on `0.000001` against 1e-06, and the numeric operators here are gte, lte, gt and lt, which are value comparisons. Those two are the same number.

The extension is small. A fifth step kind carrying a derivation function, an input and expected bytes, an open registry-backed function name, and one more runner requirement. Around twenty lines of CDDL, and nothing else in the format changes. With it, all 57 of our vectors convert.

Worth saying plainly because it bears on what this proposal is for: across the whole diff, signature, 8785, jcs, canonicaliz and digest return nothing. Positive control, `agentCard` returns 36. So the format meant to unify the conformance efforts currently cannot express an assertion about the bytes a signature is computed over.

This is the gap I would close before the format becomes the canonical one, because none of the four efforts it unifies covers that layer either. I checked all four. Two say so in their own files: a2a-tck tags its four card-signing requirements not-automatable with no executable test referencing any of them, and agentbin publishes a matrix reading not-tested for the same four across five languages. Of the other two, csit's byte-for-byte assertions compare sentinel strings and its A2A fixture sets signatures to none, and a2a-itk returns nothing for 8785, jcs, jws, dsse or ed25519.

Happy to open the CDDL change as its own pull request if that is useful, or to leave it here.

Math1987:
Hi Darrel, 
 I'd like to contribute a focused correction to `SEC-AUTH-001`, `SEC-AUTH-002`, and `SEC-AUTH-003`, building on Jakub's existing ITK ACTS stack.

I exercised [the current corpus](https://github.com/a2aproject/A2A/blob/dbcabfba4f01cc162c965f0fde5f44fc1f1e70e5/tests/acts/auth-security.acts.yaml) through the loader, JSON-RPC dispatcher and runner from [ITK #64](https://github.com/a2aproject/a2a-itk/pull/64), using in-process synthetic HTTP fixtures:

- `SEC-AUTH-001/002` pass when authentication is disabled but validation returns `-32602` for the missing `messageId` in their raw requests. Adding a valid ID makes this control fail, while the fixture's intended 401/403 rejections still pass.
- Both also pass on HTTP 200 with `{"title":"request accepted"}` a malformed SendMessage response providing no authentication-rejection evidence.
- `SEC-AUTH-003` passes when one error says “Task exists but belongs to another user” and the other says “Task not found”, both with code `-32001`. Adding `TaskNotFoundError` alone does not catch this.

These checks explicitly supplied the corpus's `capabilities.authentication` flag, which the current `AgentCapabilities` proto does not define. The fixture/applicability contract needs attention too: credentials must distinguish authenticated-but-insufficient access, and the protected task must demonstrably exist for its owner.

I'd propose valid requests, verified credential roles and positive controls, authentication-specific rejection checks, and a targeted protected/missing-task comparison. The existing raw captures and substitutions are sufficient to detect the demonstrated message leak. I'd start verification with JSON-RPC, while keeping cross-binding behavior explicit. Custom JSON-RPC authentication errors and normalization of variable response fields need an agreed contract; I would not assume a universal auth error number or byte-identical bodies.

I'm aware PROVENANCE already documents the missing `error_type` fields. The contribution would focus on these reproduced false passes and the fixture contract, with corpus changes upstream and verification using the existing ITK implementation.

Would this be a useful bounded follow-up here, coordinated with Jakub's work?

JakubWorek:
> Hi Darrel, I'd like to contribute a focused correction to `SEC-AUTH-001`, `SEC-AUTH-002`, and `SEC-AUTH-003`, building on Jakub's existing ITK ACTS stack.
> 
> I exercised [the current corpus](https://github.com/a2aproject/A2A/blob/dbcabfba4f01cc162c965f0fde5f44fc1f1e70e5/tests/acts/auth-security.acts.yaml) through the loader, JSON-RPC dispatcher and runner from [ITK #64](https://github.com/a2aproject/a2a-itk/pull/64), using in-process synthetic HTTP fixtures:
> 
> * `SEC-AUTH-001/002` pass when authentication is disabled but validation returns `-32602` for the missing `messageId` in their raw requests. Adding a valid ID makes this control fail, while the fixture's intended 401/403 rejections still pass.
> * Both also pass on HTTP 200 with `{"title":"request accepted"}` a malformed SendMessage response providing no authentication-rejection evidence.
> * `SEC-AUTH-003` passes when one error says “Task exists but belongs to another user” and the other says “Task not found”, both with code `-32001`. Adding `TaskNotFoundError` alone does not catch this.
> 
> These checks explicitly supplied the corpus's `capabilities.authentication` flag, which the current `AgentCapabilities` proto does not define. The fixture/applicability contract needs attention too: credentials must distinguish authenticated-but-insufficient access, and the protected task must demonstrably exist for its owner.
> 
> I'd propose valid requests, verified credential roles and positive controls, authentication-specific rejection checks, and a targeted protected/missing-task comparison. The existing raw captures and substitutions are sufficient to detect the demonstrated message leak. I'd start verification with JSON-RPC, while keeping cross-binding behavior explicit. Custom JSON-RPC authentication errors and normalization of variable response fields need an agreed contract; I would not assume a universal auth error number or byte-identical bodies.
> 
> I'm aware PROVENANCE already documents the missing `error_type` fields. The contribution would focus on these reproduced false passes and the fixture contract, with corpus changes upstream and verification using the existing ITK implementation.
> 
> Would this be a useful bounded follow-up here, coordinated with Jakub's work?

From what I've found out during tests with python sdk, not only this tests are not compliant with A2A spec. 
CARD-DISC-002 asserts protocolBinding ∈ {JSONRPC,GRPC,REST}; spec value for that is HTTP+JSON
Part of DM-ART-* tests are using pre 1.0 shapes
Other failing tests on ITK ACTS impl and Python SDK are also mostly caused by the error in the corpus itself (ofc some failures are caused by SDK but they are in minority)

tests/acts/push-notifications.acts.yaml
```diff
@@ -0,0 +1,454 @@
...
+                    type: string
+
+          - id: set-config
+            operation: set_push_notification_config
+            params:
+              taskId: "{{create.taskId}}"
+              pushNotificationConfig:
+                url: "{{webhookUrl}}"
+            expect_error:
+              code: UnsupportedOperationError
```
JakubWorek, line 196: https://a2a-protocol.org/v1.0.0/specification/#317-create-push-notification-config

JakubWorek:
In order to address the issues with correctness of the corpus we propose: https://github.com/a2aproject/A2A/pull/2227

astrogilda:
The false passes you reproduced are why I want vectors that need no live system: an input, the expected bytes, positive controls, reject cases. The canonicalization corpus this thread argued over is at a2aproject/a2a-tck#228 under conformance-vectors/a2a-jcs-v01: 57 RFC 8785 vectors with a Python runner and a Go runner. I'd like the pass and fail counts here from a run against your registry's card canonicalizer and signature check. A vector that disagrees with your bytes is the report.

MoltyCel:
astrogilda —
57/57 against our canonicalizer, 0 FAIL. Corpus at PR #228, branch add-a2a-jcs-v01-conformance-vectors, HEAD a7dfbc82b3ae; all 57 files checked sha256 against MANIFEST.json, no deviation.
Three paths exercised, all on the signing side: the raw canonicalize() our verdict signing and envelope store hang off, the card signing bytes from sign_agent_card(), and the card verify bytes from verify_signed_card(). Both card paths apply the same rule: strip top-level signatures, then canonicalize.
Cross-checked with two libraries, our jcs 0.2.1 and the PR runner's rfc8785 0.1.4. 57/57 each, no per-vector divergence, and byte-identical output across the 47 MUST-ACCEPT cases.
Scope, stated plainly: Python only. The Go oracle did not run here, so the result covers one language family.
On false passes. A green first run says little until the harness is shown to fail, so three mutations against the canonicalizer, each run over all 57:
one output key renamed: 54/57, 3 caught
output replaced with {}: 2/57, 55 caught
exceptions suppressed: rejects 2/10, 8 of 10 caught
The two rejects that survive exception suppression are A2-REJECT-003 and A2-REJECT-004. They assert that the canonical form does not carry the key signatures, and in an implementation that strips signatures before canonicalizing, that holds whether or not anything is thrown.
One portability note: run_python.py does not start under Python 3.9, since bytes | None needs 3.10+. Adding from future import annotations makes it run. Our clone is unmodified at a7dfbc82b3ae; that line went into a local copy.
Worth a look on your side: run A2-REJECT-003/004 against an implementation that keeps signatures through canonicalization. That will show whether they assert the rejection you intended or only the stripping behaviour.

Lars

astrogilda:
Thanks, Lars, for the run, and for the mutations that found this. You were right: both runners passed A2-REJECT-003 and A2-REJECT-004 by checking the fixture, whatever the implementation did.

At 5cf977d on a2aproject/a2a-tck#228 they compare the implementation's signing bytes instead, so keeping `signatures` fails all four A2 vectors, and run_python.py starts on 3.9. I'd propose listing your canonicalizer in the corpus README as a third rail, with your result and commit.

PR #1889, opened 2026-05-28
ci: Minor updates for workflows/automation

holtskinner:
- Fixes issue where `llms.txt` was not being deployed
- Updated `buf` install to use action in `docs`/`linter` for consistency
- Changed Docs requiremnts install to use uv to reduce deploy/test time
- Update docs proto name display with ruff added efficiency


muscariello (approved):
LGTM

PR #1891, opened 2026-05-28
fix: use POST for SubscribeToTask binding

### LINKED ISSUE #1574: [Bug]: a2a.proto and specification divergence on subscribeToTask
### What happened?

According to a2a.proto: subsribeToTask is a GET
`/ Subscribes to task updates for tasks not in a terminal state.
  // Returns `UnsupportedOperationError` if the task is already in a terminal state (completed, failed, canceled, rejected).
    option (google.api.http) = {
      additional_bindings: {
      }
    };
  }`
but in the specifications (https://a2a-protocol.org/latest/specification/#1132-task-operations) it is a POST.
We should clarify this

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #1819: [Bug]: The proto definition of "Subscribes to task" is incorrect
### What happened?

A bug happened!
```
  // Subscribes to task updates for tasks not in a terminal state.
  // Returns `UnsupportedOperationError` if the task is already in a terminal state (completed, failed, canceled, rejected).
    option (google.api.http) = {
      additional_bindings: {
      }
    };
  }
```

The get method is written in the proto file, but the actual method should be the post method.

### Relevant log output

```shell

```

### Code of Conduct

jstar0:
## Summary

- Change the `SubscribeToTask` HTTP annotation from `GET /tasks/{id}:subscribe` to `POST /tasks/{id}:subscribe`.
- Apply the same change to the tenant-scoped binding.
- Add `body: "*"` to both custom POST bindings so the proto passes the API linter.

## Changes

- Updated the default `SubscribeToTask` REST binding to use `post`.
- Updated the tenant-scoped `SubscribeToTask` REST binding to use `post`.
- Added `body: "*"` to both custom POST bindings, matching the API linter requirement for custom POST methods.

## Why

The protocol documentation already lists Subscribe to task as `POST /tasks/{id}:subscribe` in the REST method mapping and task operations sections. The proto annotation still used `get`, so generated REST bindings could disagree with the published specification.

## Scope

This only updates the `SubscribeToTask` HTTP binding in `specification/a2a.proto`. It does not change the request message shape or any other task operation.

Fixes #1819
Fixes: #1574

## Verification

```bash
(cd specification && buf dep update && buf export buf.build/googleapis/googleapis --output=.googleapis)
api-linter --config specification/.api-linter.yaml --output-format github --set-exit-status --proto-path specification --proto-path specification/.googleapis specification/a2a.proto
protoc -I specification -I specification/.googleapis --descriptor_set_out=/tmp/a2a.pb specification/a2a.proto
./scripts/build_docs.sh
```

### DIFF
--- a/specification/a2a.proto
+++ b/specification/a2a.proto
@@ -75,9 +75,11 @@ service A2AService {
   // Returns `UnsupportedOperationError` if the task is already in a terminal state (completed, failed, canceled, rejected).
   rpc SubscribeToTask(SubscribeToTaskRequest) returns (stream StreamResponse) {
     option (google.api.http) = {
-      get: "/tasks/{id=*}:subscribe"
+      post: "/tasks/{id=*}:subscribe"
+      body: "*"
       additional_bindings: {
-        get: "/{tenant}/tasks/{id=*}:subscribe"
+        post: "/{tenant}/tasks/{id=*}:subscribe"
+        body: "*"
       }
     };
   }


yankouskia (approved):

darrelmiller:
Action:  Review all SDKs to see if any actually use GET.  If we can convince ourselves that it doesn't break any existing deployments, we should accept this.

muscariello:
@darrelmiller @herczyn 

JS/Go/Java use POST now 
Python/Rust use POST as canonical method while GET as legacy.

Tehsmash:
Going backwards in the repo this was a bug ('POST' in specification and 'GET' in the proto) since the methods were introduced in the specification in v0.2.6, we just never noticed:

https://github.com/a2aproject/A2A/commit/4cfa452039325faf1bc804219edd6f5bf20825d8#diff-eebad86dc55e949d1d0630c4a2e92fed2a16b19e0ab623719ed429b5d3eee101R1084
https://github.com/a2aproject/A2A/blob/4cfa452039325faf1bc804219edd6f5bf20825d8/specification/grpc/a2a.proto#L71

As all the SDKs seem to have standardised on the written spec 'POST', for v1.0 I think we correct this to 'POST', for v0.3 there isn't much we can do.

ConnorMoss02:
Checked what the SDKs do today. Both clients send POST with no body: `a2a-go` at `a2aclient/rest.go:317-326` (`method: "POST"`, `payload: nil`) and `a2a-js` at `rest_transport.ts:360`. On the server side `a2a-dotnet` registers POST only (`A2AEndpointRouteBuilderExtensions.cs:142`, `MapPost` with no `MapGet`), while `a2a-python` (`rest_routes.py:80-81`) and `a2a-go` (`a2asrv/rest.go:54-55`) accept both.

So POST is the only method every implementation supports today. One thing on this PR specifically: since both clients send an empty body, `post:` without a `body` annotation may match what they actually do better than `body: "*"`.

PR #1898, opened 2026-05-30
docs: add payment evidence frame specification

### LINKED ISSUE #2070: [Feat]: Specify a Payment Evidence Frame for payment lifecycle receipts in task artifacts

When a payment agent returns a receipt as a task artifact, an orchestrator has no documented, transport-agnostic way to do three things without parsing the inner domain format:

- identify what kind of payment receipt the artifact carries
- hold a stable reference identifier that survives re-serialisation and late signing
- detect tampering of the wrapped receipt

The docs specify Artifacts and Parts, but there is no page describing an envelope for payment lifecycle evidence carried inside them. Implementations are inventing their own wrappers, which do not interoperate.

Add a docs/topics page specifying a Payment Evidence Frame: a JSON envelope with a closed claim_type enum covering the payment lifecycle, a deterministic frame_id derived as SHA-256 over a JCS (RFC 8785) canonical preimage that excludes frame_id and signature, and an annotated example conforming to the Artifact and Part schema in a2a.proto.

Excluding the signature from the preimage is the load-bearing detail: it keeps a recorded frame_id valid when a frame is signed after the identifier has already been referenced elsewhere.

### Describe alternatives you have considered

Leaving the envelope to each implementation. That is the current state, and it is what produces the incompatible wrappers described above.

Putting payment receipt fields directly into the Agent Card or a registry record. That places lifecycle evidence in standing identity fields, which is the wrong lifetime for per-transaction data.

### Additional context

This is scoped strictly to payment lifecycle evidence. It is deliberately not a general admission or attenuation frame. The broader question of where verifier-side trust, permit, and receipt artifacts should live relative to A2A is #1769 and is not resolved by this.

Normative references: RFC 8785 (JCS), RFC 9421 (HTTP Message Signatures), and the Internet-Draft draft-hopley-x402-payment-evidence-frame.

### Code of Conduct

chopmob-cloud:
## Summary

This PR adds docs/topics/payment-evidence-frame.md, a standalone specification page for the Payment Evidence Frame (PEF) -- a transport-agnostic JSON envelope for wrapping payment lifecycle receipts in A2A task artifacts.

PEF addresses a practical orchestration problem: when a payment agent returns a receipt as an artifact, the orchestrator needs to identify the receipt type, hold a stable reference ID that survives re-serialisation and late signing, and detect tampering -- all without parsing the inner domain format.

The document covers:

- Overview -- what PEF is and why it exists in A2A
- claim_type taxonomy -- five closed-enum values (payment_admission, payment_settlement, payment_cancellation, payment_refund, composite_verdict)
- Frame fields table -- all required and optional fields with types and semantics
- frame_id derivation -- deterministic SHA-256 over JCS-canonical preimage (frame_id and signature excluded)
- Task artifact usage -- annotated JSON example showing how a payment agent attaches a PEF artifact conforming to the A2A Artifact/Part proto schema
- frame_id stability -- rationale for excluding signature from the preimage so late signing does not invalidate recorded identifiers
- Reference implementations -- Apache 2.0 Python and TypeScript packages
- Normative references -- IETF I-D draft-hopley-x402-payment-evidence-frame, RFC 8785 (JCS), RFC 9421 (HTTP Message Signatures)

The page is also added to the mkdocs nav under Documentation, after Extensions.

IETF I-D: https://datatracker.ietf.org/doc/draft-hopley-x402-payment-evidence-frame/

Closes #2070

Related to #1769

---

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1,7 +1,13 @@
 Abhimanyu
-lfx
-luma
+JCS
+PEF
 Petschulat
 SFDC
 Siwach
+algovoi
+hopley
+jcs
+lfx
+luma
+pef
 siwachabhi
--- a/docs/topics/payment-evidence-frame.md
+++ b/docs/topics/payment-evidence-frame.md
@@ -1,0 +1,174 @@
+# Payment Evidence Frame
+
+## Overview
+
+A **Payment Evidence Frame (PEF)** is a transport-agnostic JSON envelope that
+wraps any payment lifecycle receipt under a named `claim_type`, a deterministic
+`frame_id`, and an optional transport-layer signature.
+
+PEF is defined by the IETF Internet-Draft
+[`draft-hopley-x402-payment-evidence-frame`](https://datatracker.ietf.org/doc/draft-hopley-x402-payment-evidence-frame/).
+This page describes how PEF is used within A2A task artifacts.
+
+---
+
+## Why PEF in A2A
+
+When a payment agent completes a task and returns a receipt as an artifact, the
+orchestrator faces three practical problems:
+
+1. **Type identification** -- the orchestrator must know what kind of receipt it
+   is without parsing the inner format.
+2. **Stable reference** -- the orchestrator needs an identifier that does not
+   change if the receipt is re-serialised or if a signature is appended later.
+3. **Tamper detection** -- the orchestrator must be able to confirm the inner
+   receipt has not been altered without re-parsing it end-to-end.
+
+PEF solves all three at the envelope level, before any inner-format parsing
+occurs. The outer frame carries enough metadata for orchestration logic
+(`claim_type`, `frame_id`, `receipt_hash`), while the inner `receipt` object
+remains the authoritative domain record.
+
+---
+
+## claim_type Taxonomy
+
+The `claim_type` field is a closed string enum. Each value corresponds to a
+distinct payment lifecycle event.
+
+| `claim_type`            | Lifecycle stage                      | Typical inner format                        |
+| :---------------------- | :----------------------------------- | :------------------------------------------ |
+| `payment_admission`     | Compliance screen result             | `compliance-receipt-v1`                     |
+| `payment_settlement`    | Settlement confirmation              | `settlement-attestation-v1`                 |
+| `payment_cancellation`  | Mandate or order termination         | `cancellation-receipt-v1`                   |
+| `payment_refund`        | Post-settlement refund               | `refund-receipt-v1`                         |
+| `composite_verdict`     | Verifier-of-verifier trust conclusion | `composite-trust-query-v1`                 |
+
+An orchestrator that does not recognise a `claim_type` value SHOULD treat the
+frame as opaque and forward it without modification.
+
+---
+
+## Frame Fields
+
+| Field                | Type    | Required | Description |
+| :------------------- | :------ | :------- | :---------- |
+| `pef_version`        | string  | yes      | Specification version. Currently `"1"`. |
+| `claim_type`         | string  | yes      | Closed enum; see taxonomy above. |
+| `receipt_format`     | string  | yes      | Inner receipt format identifier, e.g. `"compliance-receipt-v1"`. |
+| `receipt`            | object  | yes      | The inner receipt verbatim. |
+| `receipt_hash`       | string  | yes      | `sha256:<hex>` of the JCS-canonical form of `receipt`. |
+| `frame_id`           | string  | yes      | `sha256:<hex>` of the JCS-canonical form of the frame preimage (see below). |
+| `frame_provider_did` | string  | yes      | DID URI of the party that issued this frame. |
+| `frame_timestamp_ms` | integer | yes      | Frame creation time as Unix epoch milliseconds. |
+| `canon_version`      | string  | yes      | Canonicalisation algorithm URI. MUST be `"urn:x402:canonicalisation:jcs-rfc8785-v1"`. |
+| `signature`          | string  | no       | RFC 9421 HTTP message signature string, covering the frame preimage. |
+
+All string values are UTF-8. Integer values are JSON numbers with no fractional
+part.
+
+---
+
+## frame_id Derivation
+
+The `frame_id` is computed over the **frame preimage**: the frame object with
+`frame_id` and `signature` fields excluded before serialisation.
+
+Derivation steps:
+
+1. Construct the frame object with all required fields present and `frame_id`
+   and `signature` absent.
+2. Serialise to JSON using JCS canonicalisation (RFC 8785).
+3. Compute SHA-256 over the canonical byte sequence.
+4. Encode as the lowercase hex string prefixed with `sha256:`.
+
+```
+frame_id = "sha256:" + hex( sha256( JCS( frame_without_frame_id_and_signature ) ) )
+```
+
+This derivation ensures that `frame_id` is stable: adding a `signature` field
+after the fact does not change the identifier. Orchestrators may record or index
+`frame_id` before signing has occurred.
+
+---
+
+## Usage as a Task Artifact
+
+A payment agent attaches a PEF as an A2A task artifact with
+`type: "payment_evidence"` and `mimeType: "application/json"`. The `data` field
+carries the complete frame object.
+
+Example artifact for a settlement confirmation:
+
+```json
+{
+  "type": "payment_evidence",
+  "mimeType": "application/json",
+  "data": {
+    "pef_version": "1",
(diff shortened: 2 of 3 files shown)


Poke-nushi:
Thanks for putting this together. PEF looks useful as a payment-specific evidence frame for A2A task artifacts.

I wanted to clarify one boundary before #1769 is treated as fully resolved.

My original question in #1769 was broader than payment lifecycle receipts. It was about where verifier-side trust / permit / admission / attenuation / receipt artifacts for risky external actions should live relative to A2A.

Payment evidence can be one important evidence type. But the broader relying-party admission problem may still involve actor/principal linkage, runtime evidence, task-scoped permits, status/freshness/replay checks, local policy, attenuation semantics, verifier-side admission receipts, and post-execution linkage.

Would it make sense to treat this PR as a payment-specific partial answer to #1769, and either keep the broader verifier-side boundary discussion open or split it into a follow-up placement issue?

I am not suggesting that A2A core should own all of this. A small by-reference metadata surface, with full admission / evidence / receipt artifacts outside A2A core, may be enough.

chopmob-cloud:
Thanks for the framing. That boundary is accurate.

PEF is explicitly payment-specific: payment-lifecycle evidence covering authorisation, settlement, compliance verdict, and linkage back to the initiating task. It is not a general-purpose admission or attenuation frame.

The broader surface you describe in #1769 — actor/principal linkage, runtime evidence, task-scoped permits, status/freshness/replay, local policy, attenuation semantics, verifier-side admission receipts, post-execution linkage — is a distinct and larger problem. This PR does not attempt to resolve it.

Treating this PR as a payment-specific partial answer to #1769 and keeping the broader discussion open (or splitting into a follow-up placement issue) makes sense. The by-reference metadata surface you mention is the right shape: a small anchor in A2A core, with full admission/evidence/receipt artefacts living outside.

On our production side, the payment-specific slice maps to two artefacts: a compliance receipt (allow/refer/deny decision record, written before any side effect) and a settlement attestation (outcome record, written after). Those cover the payment evidence chain. The broader admission/attenuation layer — non-payment risky actions, principal linkage, task-scoped permits — sits outside that surface and is not addressed by PEF.

msampathkumar:
Hi @chopmob-cloud Thank you for the PR. Please join our TSE meeting to present and drive this work.

https://github.com/a2aproject/A2A/blob/main/GOVERNANCE.md#tsc-meetings

chopmob-cloud:
Thank you @msampathkumar, I've registered for the event. I should flag that the current slot clashes with my lecturing timetable, so at minimum there's a partial conflict and I may not be able to join live. In the meantime I'll get all the PEF information ready for the group to evaluate.

chopmob-cloud:
As mentioned, here are the materials for the group to evaluate PEF.

**Evaluation kit (self-contained):** https://github.com/chopmob-cloud/algovoi-pef/releases/tag/v0.1.1

Download, unzip, and run:

```bash
python validate.py
```

Nothing else is required: no pip install and no network. `algovoi-pef` and its only dependency (Trail of Bits' `rfc8785`) are vendored in the kit under `vendor/`, both Apache-2.0. It reproduces every vector's `frame_id` and `receipt_hash` byte for byte (8/8).

Also:

- Reference implementation and conformance vectors: https://github.com/chopmob-cloud/algovoi-pef (`vectors/pef_v1.json`)
- Published packages: [algovoi-pef (PyPI)](https://pypi.org/project/algovoi-pef/), [@algovoi/pef (npm)](https://www.npmjs.com/package/@algovoi/pef)
- IETF Internet-Draft: https://datatracker.ietf.org/doc/draft-hopley-x402-payment-evidence-frame/

The kit also bundles an optional eight language harness that reproduces the same vectors across eight independent JCS implementations (64 of 64).

chopmob-cloud:
**Update: a keystone-bound implementation of the Payment Evidence Frame is now available.**

Beyond the specification in this PR, PEF now ships as an installable package that binds the frame to
the AlgoVoi keystone, the end to end agentic-payment trust chain: identity, authority, policy,
decision, execution, and one verdict over the chain.

`algovoi-pef-keystone` wraps a keystone reference in a frame and pins
it, so the frame's `frame_id` commits to the exact keystone position it carries, and it chains frames
into one `trust_query_ref` capped audit chain. It is open and content addressed, runs on any substrate
version (JCS RFC 8785 only), and is byte for byte identical in Python and TypeScript. The cryptographic
signature (hybrid Falcon-1024 and ML-DSA-65) is the commercial tier.

The construction is reproducible offline from the public conformance corpus, no package import:

```
git clone https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors
python composition/pef_keystone_v1/verify_pef_keystone.py              # frame wraps + pins a keystone fact
python composition/audit_chain_of_frames_v1/verify_audit_chain_of_frames.py   # lifecycle as chained frames, capped
node   composition/pef_keystone_v1/verify_pef_keystone.mjs             # Node == Python
```

- PyPI: https://pypi.org/project/algovoi-pef-keystone/
- Conformance corpus: https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors
- Docs (the Keystone): https://docs.algovoi.co.uk/keystone

Apache-2.0.

---

Poke-nushi:
One small housekeeping point: the PR body still says `Closes #1769`.

Since we agreed above that PEF is a payment-specific partial answer and does
not resolve the broader verifier-side admission / receipt placement question,
could this be changed to `Related to #1769` so merging this PR does not
automatically close the broader issue?

chopmob-cloud:
Done, updated to `Related to #1769`. Agreed that merging this should not close the broader verifier-side admission and receipt placement question.

One pointer, since you were not on the thread when it landed: PEF now has a keystone-bound implementation that pins a frame's `frame_id` to the keystone position it carries, and chains frames into one capped audit chain. It is open, content addressed, and JCS RFC 8785 only. It reproduces offline from the public conformance corpus, no package import:

```
git clone https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors
python composition/pef_keystone_v1/verify_pef_keystone.py
node   composition/pef_keystone_v1/verify_pef_keystone.mjs
```

That is still the payment-specific slice only. The broader admission and attenuation surface you described is unchanged by it.

Silentpartnercoding:
There is one transport boundary in the new page that I do not think is verifiable as written. The embedded `signature` is described as an RFC 9421 signature “covering the frame preimage,” but RFC 9421 verification requires both `Signature` and `Signature-Input` plus the HTTP message components used to build the signature base. After a PEF is carried as an A2A `DataPart`, that original HTTP message may not exist. The referenced x402 binding also profiles HTTP challenge/payment messages, not a detached JSON-frame signature.

Could the A2A text either define a detached signature object and deterministic PEF-specific signing base (with algorithm/key identification and proof bytes), or explicitly scope RFC 9421 verification to the original HTTP transport and treat the A2A-carried frame as unsigned unless the artifact has a separate proof?

A useful negative vector would serialize a signed frame through an A2A artifact with no HTTP headers and require verification to fail rather than inferring `Signature-Input` or a key from `frame_provider_did`. That would keep the stable `frame_id` property while making the portability claim match what a downstream verifier actually receives.

chopmob-cloud:
You are right, and the point holds exactly as stated.

As written, the `signature` row is not verifiable. RFC 9421 verification reconstructs the signature base (Section 2.5) from the covered component values of the carrying HTTP message plus the signature parameters serialised in `Signature-Input` (Section 4.1), then checks the `Signature` field value against that base (Section 3.2). A bare signature string carries none of that, and once the frame travels as an A2A `DataPart` the originating HTTP message is generally gone. The frame body binds into such a signature only through `Content-Digest` (RFC 9530) when `content-digest` is a covered component. The reference implementations already behave accordingly: frame-layer verification is structural and explicitly does not verify `signature`.

I have adopted your second option, since it matches what a downstream verifier actually receives. The page now:

1. Redefines `signature` as opaque evidence preserved from the delivering HTTP exchange, removing the "covering the frame preimage" claim.
2. Adds a Signature Scope section: verification is scoped to the original HTTP transport; a frame re-carried as an A2A `DataPart` MUST be treated as unsigned at the frame layer unless the artifact carries a separate self-contained proof; verifiers MUST NOT reconstruct a `Signature-Input`, infer covered components, or derive key material from `frame_provider_did`.
3. Requires the negative vector you describe: signed frame, A2A carriage, no HTTP context, verification MUST fail rather than succeed.
4. Describes the reference implementations as structural verification only, and adds RFC 9530 to the references.

A detached PEF-specific signing base is worth defining, but that belongs in the Internet-Draft. `frame_id` stability is unchanged.

chopmob-cloud:
Closing this to reduce our open PR footprint. Thanks for the consideration.

PR #1915, opened 2026-06-05
docs: add scoped authorization receipt extension example

### LINKED ISSUE #1909: Agent Card extension example for authority contract metadata
# Agent Card extension example for authority contract metadata

## Context

A2A already provides an extension mechanism and Agent Cards as the discovery
surface for an agent's identity, capabilities, skills, endpoint, and
authentication requirements. The A2A spec also states that servers must
implement authorization checks for A2A protocol operations and scope results to
the caller's authorized access boundaries.

For production agents, it would be useful to make the agent's external
authorization requirements discoverable without embedding sensitive policy
details or credentials in the public Agent Card.

## Proposal

Add a non-normative example extension for authority contract metadata. This
could be a data-only extension for discovery, and optionally a required
extension when the server expects clients to activate an authorization protocol
before sending tasks.

Example Agent Card extension object:

```json
{
  "uri": "https://agentid.dev/extensions/authority-contract/v1",
  "description": "Declares where clients and gateways can discover the agent's authority contract and delegated-task receipt requirements.",
  "required": false,
  "params": {
    "authority_manifest_uri": "https://agent.example.com/.well-known/agentid.json",
    "authorization_contract_uri": "https://agent.example.com/.well-known/agentid-provider-contract.json",
    "receipt_required_for": ["write", "admin", "execute", "financial"],
    "receipt_transport": "message.metadata.agentid_receipt",
    "attestation_uri": "https://agent.example.com/.well-known/vcs.json"
  }
}
```

## Why this helps

This keeps identity/capability discovery separate from runtime authorization
while giving clients and gateways a standard place to find authority metadata.

It also supports common enterprise checks:

- Which actions require approval or just-in-time authority?
- Which delegated tasks require a scoped receipt?
- Where can a verifier fetch the authority contract?
- Where can a verifier fetch identity or security attestations?
- How should an authorization receipt be carried in an A2A request?

## Non-goals

- Do not define a new identity system.
- Do not replace A2A authentication or authorization.
- Do not require public Agent Cards to expose sensitive policy internals.
- Do not require AgentID specifically; AgentID can be one implementation of the
  authority contract URI.

## Related implementation

I have been working on AgentID, which uses a manifest to express action-level
authority: tool, action, resource, job, user, approval, JIT grant, data flow,
delegation, audit, and kill-switch behavior.

The extension above would let A2A clients discover that such a contract exists
and where to retrieve it, without changing the core A2A task model.

Reference: https://github.com/dinpd/AgentID/blob/main/docs/standards-alignment.md

dinpd:
## Summary

- Adds a non-normative scoped authorization receipt extension example to the A2A extensions guide.
- Shows how an Agent Card can advertise receipt transport, binding requirements, verifier metadata, and a `receipt_profile_uri` without changing the core Agent Card schema.
- Notes that domain-specific receipt profiles should define closed outcome vocabularies, while the Agent Card remains generic.

## Context

Resolves #1909.

This follows the discussion in #1909 around separating generic Agent Card discovery metadata from domain-specific receipt profile semantics.

## Validation

- `git diff --check`
- Parsed the JSON examples in `docs/topics/extensions.md`

Full MkDocs validation was not run locally because the local Python environment does not have `mkdocs` installed.

### DIFF
--- a/docs/topics/extensions.md
+++ b/docs/topics/extensions.md
@@ -176,7 +176,7 @@ specification, and the `params` object carries extension-defined metadata:
           ],
           "receipt_verification": {
             "format": "JWS",
-            "canonicalization": "jcs-rfc8785-v1",
+            "canonicalization": "rfc8785-json-canonicalization-v1",
             "algorithms": ["ES256"],
             "jwks_uri": "https://example.com/.well-known/authorization-receipts/jwks.json"
           },
@@ -211,7 +211,7 @@ the receipt issuer. At minimum, this includes the closed outcome vocabulary,
 canonical request digest construction, a single named canonicalization and
 projection that independent issuers and verifiers can reproduce, which fields
 are signed and which are referenced, freshness and replay handling, verifier
-key discovery, and profile versioning. A value such as `jcs-rfc8785-v1` is
+key discovery, and profile versioning. A value such as `rfc8785-json-canonicalization-v1` is
 illustrative; the important property is that the referenced profile pins one
 concrete scheme rather than leaving canonicalization open at verification time.
 Where a verdict basis or policy rationale contains sensitive details, the
--- a/docs/topics/extensions.md
+++ b/docs/topics/extensions.md
@@ -120,6 +120,115 @@ The following is an example of an Agent Card with an extension:
 }
 ```
 
+### Example: Scoped Authorization Receipt Extension
+
+Some enterprise deployments need a client to prove that a runtime policy system
+admitted a high-impact action before an agent processes it. For example,
+transport authentication can show that a client is allowed to reach an A2A
+server, while a scoped authorization receipt can show that a specific
+client-requested action was admitted for a particular task, resource, request
+digest, policy version, and expiry. The A2A server or downstream provider still
+needs to enforce its own business authorization before mutating state.
+
+An extension can advertise those requirements through the Agent Card without
+changing the core Agent Card schema. The extension URI identifies the extension
+specification, and the `params` object carries extension-defined metadata:
+
+```json
+{
+  "name": "Expense Approval Agent",
+  "description": "An agent that can prepare and submit expense approvals.",
+  "version": "1.0.0",
+  "supportedInterfaces": [
+    {
+      "url": "https://example.com/agents/expense-approval",
+      "protocolBinding": "HTTP+JSON",
+      "protocolVersion": "1.0"
+    }
+  ],
+  "capabilities": {
+    "extensions": [
+      {
+        "uri": "https://example.com/ext/scoped-authorization-receipts/v1",
+        "description": "Requires scoped authorization receipts for high-impact actions.",
+        "required": true,
+        "params": {
+          "authority_manifest_uri": "https://example.com/.well-known/authority-manifest.json",
+          "authorization_contract_uri": "https://example.com/contracts/expense-approval/v1",
+          "receipt_required_for": [
+            "expense.submit",
+            "expense.approve"
+          ],
+          "receipt_transport": {
+            "type": "http_header",
+            "name": "Authorization-Receipt"
+          },
+          "receipt_binds": [
+            "agent_card_digest",
+            "task_id",
+            "message_id",
+            "action",
+            "resource",
+            "request_digest",
+            "policy_version",
+            "expires_at",
+            "audience"
+          ],
+          "receipt_verification": {
+            "format": "JWS",
+            "canonicalization": "profile-defined",
+            "algorithms": ["ES256"],
+            "jwks_uri": "https://example.com/.well-known/authorization-receipts/jwks.json"
+          },
+          "receipt_profile_uri": "https://example.com/profiles/payment-compliance-receipt/v1"
+        }
+      }
+    ]
+  },
+  "defaultInputModes": ["application/json"],
+  "defaultOutputModes": ["application/json"],
+  "skills": [
+    {
+      "id": "expense_approval",
+      "name": "Expense approval",
+      "description": "Prepare and submit expense approvals.",
+      "tags": ["expenses", "approvals"]
+    }
+  ]
+}
+```
+
+In this example, `receipt_profile_uri` points to a separate profile that defines
+the receipt's outcome vocabulary and verifier rules. Keeping the profile outside
+the Agent Card lets one extension advertise the generic transport and binding
+requirements while domain-specific profiles define closed categorical outcomes,
+such as a payment-compliance verdict and basis enumeration. Implementations
+should avoid relying on open-ended receipt status strings for decisions that
+must be independently audited.
+
+The referenced profile should define how a verifier reaches the same answer as
+the receipt issuer. At minimum, this includes the closed outcome vocabulary,
+canonical request digest construction, which fields are signed and which are
+referenced, freshness and replay handling, verifier key discovery, and profile
+versioning. Where a verdict basis or policy rationale contains sensitive
+details, the profile can use a categorical value or reference instead of placing
+the full rationale in signed receipt bytes. The `JWS` format and `ES256`
+algorithm above are illustrative; profiles should define their supported
+verification formats and algorithms, and may use other trust or attestation
+mechanisms.
+
+Receipt-consuming agents should reject receipts that are expired, intended for a
+different audience, or bound to a different task, message, action, resource,
+request digest, or policy version before executing the requested action. This
+helps prevent receipt replay and scope drift when a client retries a request or
+changes the requested operation.
+
+This extension only advertises receipt metadata and verification requirements.
+It does not prove that the requested action was executed, that an external
+system accepted the receipt outcome, that a profile's outcome vocabulary is
+normative for all A2A agents, or that private policy basis is exposed in the
+Agent Card.
+
 ## Required Extensions
 
 While extensions generally offer optional functionality, some agents may have
--- a/docs/topics/extensions.md
+++ b/docs/topics/extensions.md
@@ -176,7 +176,7 @@ specification, and the `params` object carries extension-defined metadata:
           ],
           "receipt_verification": {
             "format": "JWS",
-            "canonicalization": "profile-defined",
+            "canonicalization": "jcs-rfc8785-v1",
             "algorithms": ["ES256"],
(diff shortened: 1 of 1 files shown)


chopmob-cloud:
+1 on keeping this non-normative and pushing domain semantics out to `receipt_profile_uri` rather than into the core Agent Card. We run essentially this split in production (multi-chain x402 payment gateway) and it holds up.

**1. Closed vocabulary is the right call.** The AlgoVoi payment-compliance profile uses a small closed outcome set, each value tied to a specific regulatory trigger. That closed set is what makes the receipt independently auditable — open-ended status strings don't survive audit.

**2. Binding needs a canonicalisation rule, not just a field list.** `receipt_binds` says *what* to bind, but replay/scope-drift resistance depends on *how* those fields are serialised before signing. Worth recommending profiles pin a canonical serialisation so two semantically-equal receipts hash identically and verifiers agree.

**3. Basis vs. data minimisation.** The *basis* behind a verdict can be privacy-sensitive. We keep it out of the signed bytes and surface only the categorical outcome, holding detail in operator records. The profile might allow basis-by-reference rather than inline.

**4. Keep `receipt_verification.algorithms` open.** The example pins `ES256`; the `jwks_uri` + JWS transport happily covers other signature schemes, so suggest the algorithm list stay profile-defined.

**One scope note:** this example covers classical JWS/ES256 receipts only — it doesn't address post-quantum signatures or zero-knowledge attestations. Those sit in a separate trust layer in the AlgoVoi deployment and aren't implied by this extension. Flagging so the example isn't read as the full authority/trust surface.

Net: matches what we landed on independently. `authorization_contract_uri` + `receipt_profile_uri` maps cleanly onto an on-chain authority contract + off-chain receipt profile.

Live agent card / A2A explorer if you want to probe a working deployment: https://dash.algovoi.co.uk/a2a


---

chopmob-cloud:
Quick follow-up — we went ahead and wired the extension shape into our live agent card (issuer/verifier side: we *produce* these receipts rather than require them, so `required: false`). It advertises the closed outcome vocabulary, the `EdDSA`/JWS verification metadata with a real `jwks_uri`, and a `receipt_profile_uri` — all populated from production config. Rolling out to `api.algovoi.co.uk/.well-known/agent.json` shortly if you'd like a real instance to probe against.

I hope this helps.


---

dinpd:
Thanks, this is very helpful. The `required: false` producer/verifier case is a useful example because it shows the same extension shape can advertise receipt issuance and verification metadata without forcing every client interaction to present a receipt.

The production config detail also reinforces why the Agent Card should stay generic: it can expose transport, verification metadata, and `receipt_profile_uri`, while the profile defines the closed outcome vocabulary and verifier rules.

Once the live card is available, it would be useful implementation evidence for this extension pattern.

chopmob-cloud:
It's live now — the extension is being served from production:

`https://api.algovoi.co.uk/.well-known/agent.json` → `capabilities.extensions[0]`

You can probe it directly; it passes A2A v0.3 validation with the extension present. The `receipt_verification.jwks_uri`, `receipt_profile_uri`, and `authority_manifest_uri` all resolve, and `receipt_outcomes` shows the closed `ALLOW`/`REFER`/`DENY` vocabulary in practice.

If you need anything else from our side — a different binding, a worked receipt example, or anything to help the example land — just shout.


---

dinpd:
Thanks, this is excellent implementation evidence.

I probed the live card and the extension shape is visible under `capabilities.extensions[0]`: `required: false`, issuer/verifier role, closed `ALLOW | REFER | DENY` outcomes, `JWS`/`EdDSA` verification metadata, and resolving `jwks_uri`, `receipt_profile_uri`, and `authority_manifest_uri`.

That is useful validation for keeping the A2A example non-normative: the same extension shape can describe a real producer/verifier deployment while leaving the profile to define the outcome vocabulary and verifier rules.

chopmob-cloud:
Here's a live, end-to-end-verifiable receipt for the worked example. Anyone can reproduce it — no auth required.

**1. Request a screen** (`POST https://api.algovoi.co.uk/compliance/screen`):
```json
{"recipient_address":"<addr>","network":"algorand:mainnet","amount_microunits":1000000,"asset":"USDC"}
```

**2. The response carries the receipt + its detached signature.** The `compliance_receipt` (the closed-vocabulary payload):
```json
{
  "payer_ref": "sha256:45c8fd4b3a75c567b4fb8933e996f77c56d65ce60ae55568282660bd610107c6",
  "screen_result": "ALLOW",
  "screen_timestamp_ms": 1780689523721,
  "screen_provider_did": "did:key:z6MkgExzvcpvxrghf4Q3285xqSdenhRZHcP6wc5UvY6VVaz5",
  "jurisdiction_flags": ["UK", "EU"],
  "canon_version": "jcs-rfc8785-v1"
}
```
`screen_result` is the closed `ALLOW` / `REFER` / `DENY` vocabulary; `payer_ref` is a salted hash, not the raw address (data minimisation — the *basis* stays out of the signed bytes).

**3. The compact JWS** (`compliance_receipt_jws`, EdDSA / Ed25519):
```
eyJhbGciOiJFZERTQSIsImtpZCI6ImQwNDgxZGY0Y2JiZGE4ZThhYmE4NjcwOTQxOTg4NGVmIiwidHlwIjoiSldUIn0.eyJjYW5vbl92ZXJzaW9uIjoiamNzLXJmYzg3ODUtdjEiLCJqdXJpc2RpY3Rpb25fZmxhZ3MiOlsiVUsiLCJFVSJdLCJwYXllcl9yZWYiOiJzaGEyNTY6NDVjOGZkNGIzYTc1YzU2N2I0ZmI4OTMzZTk5NmY3N2M1NmQ2NWNlNjBhZTU1NTY4MjgyNjYwYmQ2MTAxMDdjNiIsInNjcmVlbl9wcm92aWRlcl9kaWQiOiJkaWQ6a2V5Ono2TWtnRXh6dmNwdnhyZ2hmNFEzMjg1eHFTZGVuaFJaSGNQNndjNVV2WTZWVmF6NSIsInNjcmVlbl9yZXN1bHQiOiJBTExPVyIsInNjcmVlbl90aW1lc3RhbXBfbXMiOjE3ODA2ODk1MjM3MjF9.8KYvvJV0AwR-0cC5PoEgZdn3FqjmTGuSnmUfKzuhgc_3_j-27lD0sofGEEE41MuZmPAYytCEksZzyux8qaBGAA
```

**4. Verify it** against the published JWKS — the `kid` in the JWS header (`d0481df4…`) resolves in `https://api.algovoi.co.uk/.well-known/jwks.json`:
```python
import json, base64, urllib.request
from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PublicKey
b64u = lambda s: base64.urlsafe_b64decode(s + "="*(-len(s)%4))
h, p, sig = JWS.split(".")
kid = json.loads(b64u(h))["kid"]
jwks = json.load(urllib.request.urlopen("https://api.algovoi.co.uk/.well-known/jwks.json"))
key = next(k for k in jwks["keys"] if k["kid"] == kid)
Ed25519PublicKey.from_public_bytes(b64u(key["x"])).verify(b64u(sig), (h+"."+p).encode())
# no exception -> valid
```
This is exactly the `receipt_verification` block advertised on the agent card (`format: JWS`, `algorithms: ["EdDSA"]`, the `jwks_uri`). The action itself is separately bound: `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`, returned alongside as `f59966941c19…`, which is the `request_digest` referenced in `receipt_binds`.

Hope this is a useful concrete reference for the example.


---

rpelevin:
I would keep the Agent Card example generic, but make the profile contract slightly sharper.

The card can advertise that receipt metadata exists. The profile should define how a verifier reaches the same answer. I would make the example say that `receipt_profile_uri` points to rules for:

- closed outcome vocabulary
- canonical request digest construction
- signed vs referenced fields
- freshness and replay handling
- verifier key discovery
- profile versioning

The main thing to avoid is letting the Agent Card imply more than it proves. The card can say where receipt verification metadata lives; it should not imply that every action is authorized, that every interaction requires a receipt, or that the profile's domain vocabulary is part of A2A core.

One small docs addition that would help: a "what this extension does not prove" note. For example:

- it does not prove the action was executed
- it does not prove the external system accepted the outcome
- it does not make the outcome vocabulary normative for all A2A agents
- it does not expose private policy basis in the Agent Card

That would preserve the useful discovery layer while keeping authority and verification in the profile.

chopmob-cloud:
Useful example, and keeping the outcome vocabulary and canonical digest construction in the referenced profile rather than the Agent Card is the right split.

For a concrete, resolvable reference: AlgoVoi runs this same extension in production, under the same name and slug. The agent card advertises it and the profile URI actually dereferences:

- Agent card: https://api.algovoi.co.uk/.well-known/agent.json
- Profile (resolves to the full document): https://api.algovoi.co.uk/ext/scoped-authorization-receipts/v1
- Receipt verifier: `POST https://api.algovoi.co.uk/v1/receipt/verify`

Two production notes that may be worth folding into the guidance:

1. The example leaves `canonicalization` as `profile-defined`. In practice that field is load-bearing: a verifier can only reach the issuer's answer if the profile pins one concrete scheme. Ours pins JCS (RFC 8785) so the signed bytes are reproducible across independent implementations; `profile-defined` should resolve to a single named canonicalization rather than stay open.

2. The closed categorical outcome holds up well (we run `ALLOW` / `REFER` / `DENY`) precisely because the basis stays a categorical value or reference and not free text in the signed bytes, as your final paragraph already notes.

Live profile and verifier above if a working reference is useful.

*AlgoVoi (chopmob-cloud) -- docs.algovoi.co.uk/substrate-2*

dinpd:
Thanks, this is useful production evidence. I pushed a small update that folds in the two guidance points while keeping the Agent Card example generic:

- changed the example from `profile-defined` to a concrete illustrative canonicalization value, `jcs-rfc8785-v1`;
- clarified that the referenced profile should pin one named canonicalization/projection that independent issuers and verifiers can reproduce, rather than leaving canonicalization open at verification time;
- tightened the basis guidance so sensitive rationale stays categorical or by-reference, not free-form text in signed receipt bytes.

Validation:
- `git diff --check`
- parsed the JSON examples in `docs/topics/extensions.md`

PR #1926, opened 2026-06-09
chore: Fix Mermaid diagram syntax in index.md to use markdown properly

SrulyRosenblat:
Fix an issue with markdown on graph


msampathkumar:
Thank you for your work @SrulyRosenblat 

The page is updated now.

PR #1965, opened 2026-06-21
docs(spec): fix broken intra-document anchor links

msrivastav13:
## Summary

Fixes #1945.

Three internal links in `docs/specification.md` pointed to anchors that no header generates, breaking in-document navigation. This corrects them to the actual section-numbered anchors:

| Location | Before | After |
|---|---|---|
| Line 674 (Section 3.1.8 link) | `#76-taskspushnotificationconfigget` | `#318-get-push-notification-config` |
| Line 2573 (StreamResponse) | `#stream-response` | `#323-stream-response` |
| Line 2615 (StreamResponse) | `#stream-response` | `#323-stream-response` |

## Notes

- Verified by checking every `](#...)` link target against the set of slugs generated from the document's headers — after this change, no broken intra-document anchors remain in `specification.md`.
- The reporter's specific `#message` example is already corrected on `main`; the live site still shows it because it serves the published v1.0 spec.

PR #1966, opened 2026-06-21
docs: generate correct numbered anchors for proto type links

### LINKED ISSUE #1945: [Bug]: Invalid HTML anchors in spec
### What happened?

In some places, HTML anchors are incorrect, leading to non-working intra-document links. 

For example, going to [3.2.1. SendMessageRequest](https://a2a-protocol.org/latest/specification/#321-sendmessagerequest), and hovering Message link within the table, you'll see https://a2a-protocol.org/latest/specification/#message, but valid link would be https://a2a-protocol.org/latest/specification/#414-message (`#message` vs `#414-message`).

Image

I can see that the links are broken in many tables, making navigation very difficult.

### Code of Conduct

msrivastav13:
## Summary

Fixes #1945. This is the second of two PRs for that issue — #1965 fixes hand-written anchor links; this one fixes the **generated** links in the proto tables, which is the larger half of the problem the issue describes ("broken in many tables"). Either PR closes the issue; they can be merged independently.

## Problem

The proto table macros in `.mkdocs/macros.py` (`proto_to_table` / `proto_service_to_table`) linked each non-primitive type to a bare lowercased slug:

```python
label = f'[{label}](#{display_name.lower()})'   # Message -> #message
```

But `specification.md` section headers render to **numbered** anchors (e.g. `#### 4.1.4. Message` → `#414-message`). So every type cross-reference inside the generated tables was a dead link — about **45** across the spec, including the `#message` example called out in #1945.

## Fix

`_format_type_for_docs` now resolves each type name against a map built by scanning the spec's numbered headers (`#### 4.1.4. Message` → `#414-message`). Types that have **no** dedicated header (e.g. `SendMessageResponse`, `StreamResponse`, the `*Request` wrappers — 15 of them) render as plain text instead of a broken link.

## Verification

Local `mkdocs build`:
- `'no such anchor'` warnings for `specification.md` dropped from **~45 → 0**
- Spot-checked rendered HTML: `Message` → `#414-message`, `Task` → `#411-task`, `AgentCard` → `#441-agentcard`; undocumented types render as plain `<code>` with no href.


msrivastav13:
Note on CI: the **build_and_deploy** check (`docs.yml`) does not run on this PR because its `paths:` filter watches `docs/**`, `mkdocs.yml`, etc. but not `.mkdocs/**` — so a change to `.mkdocs/macros.py` alone doesn't trigger a docs rebuild. I verified the build locally instead:

```
$ mkdocs build   # via the repo's requirements-docs.txt
INFO - Documentation built in ~4s
# 'no such anchor' warnings for specification.md: ~45 (before) -> 0 (after)
```

Rendered HTML spot-check: `Message` → `#414-message`, `Task` → `#411-task`, `AgentCard` → `#441-agentcard`; undocumented types (`StreamResponse`, `*Request` wrappers) render as plain `<code>` with no `href`.

Separately, maintainers may want to add `.mkdocs/**` to the `docs.yml` path filter so future changes to the rendering macros are covered by CI.

msrivastav13:
## Addressed review feedback

Applied both Gemini suggestions in d3f2dee: `_ANCHOR_MAP` is now initialized to `{}` instead of `None`, so the lookup is a plain `_ANCHOR_MAP.get(display_name)` with no `None` check or `or {}` fallback. Re-verified with a local build — still 0 `no such anchor` warnings.

## Self code-review notes

I also reviewed the change myself for correctness:

- **Duplicate-header collisions:** The header regex matches any numbered single-word header, so I checked for type-name collisions across the spec. The only one is `Streaming` (`### 10.7` vs `### 11.7`), which is a section title, not a proto type — no actual type resolves ambiguously. All real types map to their canonical §4 data-model anchors (`Task` → `#411-task`, `Message` → `#414-message`, `Artifact` → `#417-artifact`, etc.).
- **Slug parity:** `_slugify` mirrors the repo's configured `pymdownx.slugs.slugify` (`case: lower`) — lowercase, strip non-`[a-z0-9 _-]`, spaces/underscores → `-`, collapse repeats. Confirmed against rendered HTML `id`s.
- **Graceful degradation:** If `specification.md` is missing or a type has no header, `_build_anchor_map` returns/omits cleanly and the type renders as plain `<code>` — no exceptions, no dead links.
- **Scope:** Pure rendering-logic change in `.mkdocs/macros.py`; no spec content or proto changes.

msampathkumar (approved):
Tested locally. LGTM.

msampathkumar:
Nice work @msrivastav13. 

This change updates the specification documents. So I have informed TSE to check and add second review.

darrelmiller (approved):

Issue #1864, opened 2026-05-19
[Proposal] New experimental repository: experimental-cpb-websocket


hackeramitkumar:
# WebSocket Custom Protocol Binding for A2A - Proposal

## Abstract

This proposal requests the creation of an `experimental-cpb-websocket` repository under the `a2aproject` GitHub organization to host the **WebSocket** custom protocol binding for A2A.

The binding carries the **existing** A2A method inventory - the same operations, request/response shapes, and server-streaming semantics defined by the core specification and used by the JSON-RPC binding - over a single, long-lived [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) connection. It is layered:

- **Transport:** WebSocket (RFC 6455) provides nothing but an ordered, message-framed, full-duplex byte pipe over an HTTP/1.1 `Upgrade` handshake.
- **Application:** [JSON-RPC 2.0](https://www.jsonrpc.org/specification) - the **same** request, response, and numeric-coded error objects as the A2A JSON-RPC binding - carried one object per WebSocket text frame.

On top of these, the binding adds a small set of **A2A extension frames** (multiplexing by request `id`, a `streamEnd` terminator, and a `cancelStream` control frame) that exist only because one persistent connection now carries many concurrent RPCs.

The binding is named simply **WebSocket** after its transport layer. Multiplexing, streaming termination, and in-band control are a thin A2A layer added on top of standard WebSocket-as-transport, with JSON-RPC 2.0 as the unchanged message format on the wire.

The binding **does not introduce any new operation or streaming semantic to A2A**. It uses the WebSocket connection purely as a *multiplexed RPC transport*: a single connection carries many concurrent request/reply and server-stream RPCs, distinguished by `id`. The connection's full-duplex nature is used only so that a client can issue and cancel independent RPCs at any time - including while server-streaming responses for other requests are still in flight - not to define a new bidirectional-streaming operation.

## Motivation

The core A2A specification defines three standard bindings: HTTP+JSON, HTTP JSON-RPC, and gRPC. The two HTTP/1.1 bindings carry one RPC per HTTP request, and server-streaming RPCs (e.g. `SendStreamingMessage`) are delivered over Server-Sent Events - one long-lived HTTP response per stream. gRPC multiplexes many RPCs and server-streams over one HTTP/2 connection, but requires Protocol Buffer tooling and HTTP/2 end-to-end, and is not natively available in browsers.

[...]

Issue #656, opened 2025-05-26
[Feat]: Support for stream input


yichenj:
### Is your feature request related to a problem? Please describe.

Hello, I notice that the input in A2A communication protocol is always a group of messages, and it works as a ping-pong or ping-stream mode. However, agents can come in many forms. For speech instance, perhaps a bidirectional stream is used in a ASR-like agent, but it cannot  fit well in the A2A system. What do you think about this?

### Describe the solution you'd like

How about introduce a duplex stream like WS or HTTP2/3 internal stream, and add a suffix path(like <BASE_URL>/ws prehaps) to upgrade/downgrade this HTTP stream(depends on WS or HTTP2 stream)

### Describe alternatives you've considered

Maybe stream input should be considered as chucked messages, but this could cause mass HTTP header cost

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

pstephengoogle:
Thank you for this request. Yes, we are looking into the stream in/stream out paradigm as well. It has been lower priority as it is a more complicated environment for the agent developers and creates lots of challenges with a multi-agent network. Will keep this issue open until this effort is taken up

SimplyKnownAsG:
It looks like [this commit from July 29 added streaming](https://github.com/a2aproject/A2A/commit/098fd8f5bfe65880af66f65cc6c0c130ac09a921) as a possible gRPC extension.

> **gRPC**: May leverage gRPC-specific features like bidirectional streaming, metadata, or custom status codes.

Any update for incorporating other bidirectional streaming transport options?

HungYangChang:
Hi @pstephengoogle 

I would like to follow up this
Could the A2A protocol itself be extended to support bidirectional streaming rather than maintaining two parallel streams? 

cc @yichenj

pstephengoogle:
This is proposed in https://github.com/a2aproject/A2A/pull/1120 

It has/is being discussed in TSC meetings. Feel free to comment on that PR

Tehsmash:
Thanks for raising this. We're consolidating closely related work into a single tracking issue so the discussion and design stay in one place: **https://github.com/a2aproject/A2A/issues/1995**. Your report is captured as a checklist item there. We plan to close this issue in favor of the tracking issue on or after **2026-07-02** — please follow it and continue the conversation there. If a specific requirement was lost or mis-summarized, comment before then and we'll fold it in (or reopen). Nothing is being dropped — just centralized.

Tehsmash:
[...]

Issue #1942, opened 2026-06-16
[TSC] Proposal: backlog consolidation & v1.1 scope alignment


geneknit:
## What this issue is asking of the TSC

Approval to take **23 open issues** and either **consolidate them** (16 issues → 5 tracking issues) or **close them as duplicates / off-topic** (6 issues); one extension proposal (#1796) is **retained and labeled** rather than closed. Consolidations are executed with a one-week comment window.

**This vote covers the *process and the mapping*, not the technical resolution of any individual issue** — each tracking issue keeps its own design discussion open.

## Rationale

- Several v1.1-relevant clusters are really *one* piece of work split across many issues (stream resumption, idempotency, push-config, skill selection, auth discovery). Consolidating gives contributors one place to follow each topic and affords the project a legible foundation to a v1.1 scope.
- A handful of issues are exact duplicates or off-topic (vendor notices, community submissions) and should be closed with a clear, courteous redirect.

## The plan

1. **Create** 5 umbrella "tracking" issues (drafts in the appendix).
2. **Notice** — post a consolidation comment on each consolidated issue announcing a **comment window (default: 1 week)**. Note: this does not apply to the "safe-now "closes (2 verified duplicates, 5 off-topic redirects).
3. **Close** — after the window, close the consolidated children, each referencing its umbrella.

**Guardrails:**
- Notice-before-close with a comment window — honors our "adequate notice and time for comment" norm; nothing is silently closed.
- Every close references the umbrella/original, so GitHub backlinks preserve the trail.

Draft consolidation comment each affected issue will receive:

> Thanks for raising this. We're consolidating closely related work into a single tracking issue so the discussion and design stay in one place: **[tracking issue]**. Your report is captured as a checklist item there. We plan to close this issue in favor of the tracking issue on or after **[window date]** — please follow it and continue the conversation there. If a specific requirement was lost or mis-summarized, comment before then and we'll fold it in (or reopen). Nothing is being dropped — just centralized.

## Mapping for consolidation
_Note: this mapping was the original proposal, an extended consolidation with a total of 8 umbrella epics is detailed in the thread ad os 30 June 2026_
| New tracking issue | Consolidates |
|---|---|
| **Reliable stream resumption & event ordering** | #1794 (anchor), #1410, #1635, #999 |
| **Idempotency & safe retries** | #928 (anchor), #1893 |
| **Push-notification config — security & semantics** | #1681 (anchor), #1250, #1483 |
[...]

PR #1975, opened 2026-06-23
feat: add bidirectional live sessions (SendLiveMessage)

herczyn:
## Description

A2A v1.0 supports server→client streaming only (`SendStreamingMessage`): a client cannot stream additional input on the same call, and there is no full-duplex media channel for real-time audio/video. This PR adds a bidirectional **live session** so a client and an agent can exchange continuous media and multiple turns over a single connection.

The change is **additive and capability-gated** — existing methods and types are unchanged, so v1.0 clients/servers are unaffected. In particular, the existing `StreamResponse` is *embedded* in the new `LiveResponse` rather than modified, so `SendStreamingMessage` / `SubscribeToTask` are untouched.

### What's added (`specification/a2a.proto`)

- **`SendLiveMessage(stream LiveRequest) returns (stream LiveResponse)`** — a bidirectional streaming RPC (requires a bidi-capable transport, e.g. gRPC).
- **`LiveRequest` / `LiveResponse`** — two-arm wrappers that *compose* existing types:
  - `LiveRequest = oneof { SendMessageRequest message | MediaFrame frame }`
  - `LiveResponse = oneof { StreamResponse event | MediaFrame frame }`
- **`MediaFrame`** — a media-plane frame (not a turn; no `message_id`/`role`) carrying audio/video/binary in `Part.raw`, with `start`/`delta`/`stop`/`pause`/`interrupt` (barge-in) events and per-modality sequencing; plus **`ModalitySpec`** and the **`Directionality`** enum (SDP-style directions).
- **`AgentCapabilities.bidi_streaming`** (live/duplex marker) and **`AgentCapabilities.live_session`** (`LiveSessionCapabilities`, per-modality media support).
- **`SendMessageConfiguration.live_session`** (`LiveSessionOffer`) — the media offer carried on the opening message.

### Design notes

- The stream opens with a `SendMessageRequest` (existing contract); a `message` frame is semantically identical to a standalone `SendMessageRequest`, so this does **not** change the semantics of sending a message to a running task.
- Media negotiation is **capability-based** (Agent Card + offer in `SendMessageConfiguration`) — no separate signaling/control message family: acceptance is a successful stream, rejection is a standard error, teardown is the transport close.
- The session binds to the `Task` from the opening message; transports come from the existing `supported_interfaces`.

### Scope / follow-ups

This PR is the protocol (proto) definition. A WebSocket binding (for browser/edge clients that can't use gRPC bidi) and the prose update to `docs/specification.md` are intended as separate follow-ups, to keep this change focused and shippable incrementally.

### Validation

- `buf lint` passes locally (no new findings).
- Additive-only (new field numbers, new messages, new RPC, new enum) — no breaking changes.

Related to #656, #1549, #1864.

### DIFF
--- a/specification/a2a.proto
+++ b/specification/a2a.proto
@@ -5,6 +5,7 @@ package lf.a2a.v1;
 import "google/api/annotations.proto";
 import "google/api/client.proto";
 import "google/api/field_behavior.proto";
+import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/struct.proto";
 import "google/protobuf/timestamp.proto";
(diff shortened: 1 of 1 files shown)


Tehsmash:
This seems very similar to https://github.com/a2aproject/A2A/pull/1549/changes, which was originally based on a PR/proposal from @mikeas1  https://github.com/a2aproject/A2A/pull/1120 

I think we need to consolidate into a single epic issue with design discussion instead of a bunch of PRs, because otherwise we end up missing things which have already been discussed in other places, for example 1549 handles the reconnection to a stream but it doesn't look like this one covers that yet.

We should look at how this relates to Tasks, there are a number of issues open on multi-turn for tasks, I've proposed a consolidated issue for this here: https://github.com/a2aproject/A2A/issues/1942#issuecomment-4778642008 I think decisions made on this will influence how/where we hook BiDi into the protocol.

PR #1997, opened 2026-06-26
fix(proto): correct gRPC URL example in AgentInterface

### LINKED ISSUE #1388: [Bug]: Wrong example of AgentInferface URL for gRpc
### What happened?

The current description of the AgentInterface URL mentions that for grpc the URL would look like: "https://grpc.example.com/a2a"
https://github.com/a2aproject/A2A/blob/f9a8f5b85d5b07824c52d55d63f7d71ccc6303c5/specification/grpc/a2a.proto#L392-L395

However, gRpc URLs should not include the `https` part and the path

### Relevant log output

```shell

```

### Code of Conduct

Tehsmash:
## Summary

Fixes the incorrect gRPC URL example in the `AgentInterface` message of `a2a.proto`.

**Problem:** The example showed `"https://grpc.example.com/a2a"` for a gRPC interface, but gRPC URLs do not use an `https://` scheme or a path. The correct format is `hostname:port`.

**Fix:**
- Updated the example URL from `"https://grpc.example.com/a2a"` to `"grpc.example.com:443"`
- Clarified the field comment to note that the HTTPS URL requirement applies to HTTP-based transports only, while gRPC uses `hostname:port` format

Fixes #1388


chopmob-cloud:
Verified this fix independently while tracing the same example. The reference SDK confirms the corrected format: a2a-go's gRPC transport passes `AgentInterface.url` straight into `grpc.NewClient(iface.URL, ...)` (`a2agrpc/v1/client.go`), and a gRPC dial target is `host:port`, so the current `https://grpc.example.com/a2a` example would fail as written while `grpc.example.com:443` works. Scoping the HTTPS wording to HTTP-based transports is the right accompanying clarification. Hope this helps it land.

msampathkumar (approved):

darrelmiller (approved):

PR #2046, opened 2026-07-12
docs(spec): fix Agent Card security requirement sample

### LINKED ISSUE #2044: [Bug]: Sample Agent Card in specification.md uses the v0.3 `security` field; the proto defines `security_requirements` (and a different shape)
### What happened?

While implementing a v1.0 server and checking my Agent Card against the specification, I found that the sample Agent Card in the prose contradicts the normative `.proto` on the security requirements field — in two ways at once.

[`docs/specification.md` L2166](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/specification.md#L2166) shows:

```json
"security": [{ "google": ["openid", "profile", "email"] }],
```

But [`specification/a2a.proto` L383](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/specification/a2a.proto#L383) declares the field on `AgentCard` as:

```proto
repeated SecurityRequirement security_requirements = 9;
```

and `SecurityRequirement` is a message wrapping a map, not a bare map:

```proto
message SecurityRequirement {
  // A map of security schemes to the required scopes.
  map<string, StringList> schemes = 1;
}
```

So the sample deviates on **the field name** (`security` vs `securityRequirements`) and on **the value shape** (the `schemes` wrapper is missing). Under ProtoJSON, a conformant card should read:

```json
"securityRequirements": [ { "schemes": { "google": ["openid", "profile", "email"] } } ],
```

which is what the official Go SDK (`a2a-go`) emits for a card carrying a security requirement.

Notably, the *rest* of the same sample is correct ProtoJSON — the neighbouring `securitySchemes` entry uses the wrapped oneof form (`"openIdConnectSecurityScheme": { ... }`). That makes `security` look like a leftover from 0.3, where it was the correct spelling, rather than an intentional alias.

**Impact.** `docs/specification.md` is the document most people read; the sample card is the thing they copy. Anyone writing a client or server from it will look for a `security` field that no conformant implementation sends, and will build the wrong value shape. It fails silently — auth requirements simply appear to be absent.

**Expected:** the sample matches the normative proto.

It appears to be a single occurrence (`grep -rn '"security":' docs/` returns only this line), so the fix should be a one-line change to the sample.

This issue was opened with AI assistance.

### Relevant log output

```shell
# on main @ 2183794bfb9b67af4aee1be0a0ef726050642873

$ grep -rn '"security":' docs/
docs/specification.md:2166:  "security": [{ "google": ["openid", "profile", "email"] }],

$ grep -n 'security_requirements' specification/a2a.proto
383:  repeated SecurityRequirement security_requirements = 9;
451:  repeated SecurityRequirement security_requirements = 8;
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hxaxd:
# Description

The sample Agent Card still used the v0.3 `security` field even though the v1.0 specification makes `a2a.proto` authoritative and adopts strict ProtoJSON serialization. Readers copying the sample could therefore send a field that conforming v1.0 implementations do not recognize, making agent-level authentication requirements appear absent.

PR #1160 introduced the stale line while restructuring the specification around a transport-independent Protocol Buffers model. A later edit migrated the neighboring `securitySchemes` example, but this security-requirement line retained its v0.3 spelling and shape.

This change replaces that line with the complete v1.0 ProtoJSON representation:

```json
"securityRequirements": [{ "schemes": { "google": { "list": ["openid", "profile", "email"] } } }]
```

- [x] Make your Pull Request title follow the Conventional Commits specification and repository rules (e.g., `docs(spec):` for `docs/specification.md`, `docs:` for other docs, and `feat:` / `fix:` reserved for `a2a.proto`). See [CONTRIBUTING.md](https://github.com/a2aproject/A2A/blob/main/CONTRIBUTING.md#conventional-commits) for details.
- [x] Ensure the tests and linter pass (Run `bash scripts/format.sh` from the repository root to format)

Fixes #2044

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -2163,7 +2163,7 @@ Clients verifying Agent Card signatures **MUST**:
       }
     }
   },
-  "security": [{ "google": ["openid", "profile", "email"] }],
+  "securityRequirements": [{ "schemes": { "google": { "list": ["openid", "profile", "email"] } } }],
   "defaultInputModes": ["application/json", "text/plain"],
   "defaultOutputModes": ["application/json", "image/png"],
   "skills": [


chopmob-cloud:
Confirmed this is the conformant shape, verified against the normative proto and against a2a-go's generated types. The sample in this PR is the one to take.

The proto makes `StringList` a message rather than a bare list:

```proto
message StringList {
  repeated string list = 1;
}

message SecurityRequirement {
  map<string, StringList> schemes = 1;
}
```

Because `schemes` is a map whose values are a message type, ProtoJSON serializes each value as that message's JSON object, so the scopes sit inside a `list` field. That is exactly what this PR writes:

```json
"securityRequirements": [{ "schemes": { "google": { "list": ["openid", "profile", "email"] } } }]
```

a2a-go's generated types agree. In `a2apb/v1/a2av1.pb.go`, `SecurityRequirement.Schemes` is `map[string]*StringList`, and `StringList` carries `List []string` with the json name `list`. The map value is a message pointer, so the `list` wrapper is emitted whether the card is marshalled with protojson or with encoding/json.

One thing worth flagging for whoever merges this, because it is the reason this fix is easy to get wrong. The shape suggested in #2044 drops the wrapper:

```json
"securityRequirements": [{ "schemes": { "google": ["openid", "profile", "email"] } }]
```

That is not valid ProtoJSON for `map<string, StringList>`, so correcting the sample to that form would replace one non-conformant sample with another. The related claim in the issue, that a2a-go emits the bare array, does not match a2a-go's generated types above.

The field-name half of the issue is right. `security` is the v0.3 spelling, which a2a-go's own compat layer confirms by mapping `Security` on the v0.3 card to `SecurityRequirements` on v1, and the proto declares `security_requirements` on `AgentCard`.

It is worth stating why that half is the dangerous one. A card written from the current sample carries its auth requirement under a name no v1 consumer reads. A strict ProtoJSON parser rejects the unknown field outright, but a lenient consumer simply does not see it, so an agent that requires authentication reads as one that declares none. That is the fail-open direction, and it is why the replacement sample is worth getting exactly right rather than approximately right. This PR fixes both the name and the value shape correctly.

hxaxd:
@holtskinner Hi Holt — apologies for the ping. This is a small correction to a specification example; the current branch is mergeable and the current checks are green. It has been waiting for review for just over a week. Could you please take a look when you have a moment? Thank you!

darrelmiller (approved):

PR #2056, opened 2026-07-14
docs: align v1 migration guide with proto

### LINKED ISSUE #2045: [Bug]: whats-new-v1.md documents streaming event names and Task/PushNotificationConfig fields that do not exist in the proto
### What happened?

`docs/whats-new-v1.md` is the document a v0.3 implementer reads first when migrating. Several of the v1.0 changes it announces do not match [`specification/a2a.proto`](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/specification/a2a.proto). Code written from this guide compiles, runs, and silently does the wrong thing.

Three distinct claims, all checked against `main` @ `2183794bfb9b67af4aee1be0a0ef726050642873`:

---

**1. Streaming event member names are wrong (`taskStatusUpdate` / `taskArtifactUpdate`).**

The guide states the new discrimination pattern explicitly ([L471](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L471)):

> ✅ **NEW PATTERN:** Event type determined by JSON member name (`taskStatusUpdate` or `taskArtifactUpdate`)

and shows migration code to match ([L668-671](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L668-L671)):

```js
if ("taskStatusUpdate" in event) {
  handleStatusUpdate(event.taskStatusUpdate);
} else if ("taskArtifactUpdate" in event) {
```

The proto names the oneof members `status_update` and `artifact_update` — i.e. `statusUpdate` and `artifactUpdate` under ProtoJSON:

```proto
message StreamResponse {
  oneof payload {
    Task task = 1;
    Message message = 2;
    TaskStatusUpdateEvent status_update = 3;
    TaskArtifactUpdateEvent artifact_update = 4;
  }
}
```

A client written from the guide matches on member names a conformant server never sends, so **every streaming update is silently dropped** — no error, just a stream that appears to produce nothing.

(The same passages also show a `"kind"` discriminator at [L447](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L447) and [L480](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L480), inside snippets whose surrounding prose says `kind` was *removed* in v1.0.)

---

**2. `Task.createdAt` / `Task.lastModified` do not exist.**

[L86](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L86):

> **✅ NEW:** `createdAt` and `lastModified` timestamp fields added to Task object

with sample code at [L869-870](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L869-L870):

```js
const taskAge = Date.now() - new Date(task.createdAt).getTime();
const timeSinceUpdate = Date.now() - new Date(task.lastModified).getTime();
```

`message Task` has six fields, none of them a timestamp:

```proto
message Task {
  string id = 1 [(google.api.field_behavior) = REQUIRED];
  string context_id = 2;
  TaskStatus status = 3 [(google.api.field_behavior) = REQUIRED];
  repeated Artifact artifacts = 4;
  repeated Message history = 5;
}
```

`new Date(undefined)` yields `Invalid Date`, so that sample produces `NaN`.

---

**3. `createdAt` on the push notification config does not exist either.**

[L158](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L158) and [L433](https://github.com/a2aproject/A2A/blob/2183794bfb9b67af4aee1be0a0ef726050642873/docs/whats-new-v1.md#L433) announce a `createdAt` timestamp on `PushNotificationConfig`. The corresponding message is `TaskPushNotificationConfig`, and it carries `tenant`, `id`, `task_id`, `url`, `token`, `authentication` — no timestamp.

In fact `grep -n 'created_at\|last_modified' specification/a2a.proto` returns **nothing at all**: there is no such field anywhere in the schema.

---

**Impact.** These are not stylistic divergences; they are an API surface the guide invents. Claim 1 in particular fails silently and is the kind of thing a migrating implementer only discovers by diffing against a real server.

**Expected:** the migration guide matches the normative proto, or states clearly which sections are aspirational.

Filed after implementing a v1.0 server against the proto and finding that the guide disagreed with the wire. **Prepared with AI assistance**; every claim above was verified by hand against the files at the pinned commit.

### Relevant log output

```shell
# on main @ 2183794bfb9b67af4aee1be0a0ef726050642873

$ grep -n 'taskStatusUpdate\|taskArtifactUpdate' docs/whats-new-v1.md
447:  "kind": "taskStatusUpdate",
459:  "taskStatusUpdate": {
471:- ✅ **NEW PATTERN:** Event type determined by JSON member name (`taskStatusUpdate` or `taskArtifactUpdate`)
480:  "kind": "taskArtifactUpdate",
491:  "taskArtifactUpdate": {
668:if ("taskStatusUpdate" in event) {

$ grep -n 'status_update\|artifact_update' specification/a2a.proto
    TaskStatusUpdateEvent status_update = 3;
    TaskArtifactUpdateEvent artifact_update = 4;

$ grep -n 'created_at\|last_modified' specification/a2a.proto
# (no output — the fields do not exist)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hxaxd:
# Description

The v1.0 migration guide still described stream wrappers and timestamp fields that do not exist in the normative `a2a.proto`. A client copied from the guide would check `taskStatusUpdate` / `taskArtifactUpdate`, while strict ProtoJSON emits `statusUpdate` / `artifactUpdate`, so conforming stream updates could be silently ignored. The guide also recommended `Task.createdAt`, `Task.lastModified`, and push-notification configuration fields that the proto cannot serialize.

This change aligns the v1.0 JSON and TypeScript examples with the canonical ProtoJSON member names, removes the nonexistent timestamp and `configId` claims, preserves the valid `TaskStatus.timestamp` guidance, and leaves the intentional v0.3 `kind` examples unchanged.

- [x] Make your Pull Request title follow the Conventional Commits specification and repository rules
- [x] Ensure the tests and linter pass

Fixes #2045 🦕

### DIFF
--- a/docs/whats-new-v1.md
+++ b/docs/whats-new-v1.md
@@ -184,7 +184,7 @@ The v1.0 release focuses on four major themes:
 
 - **✅ BREAKING:** All IDs are now simple literals
 - **✅ BREAKING:** Operations that previously used compound IDs now separate parent and resource ID
-    - Example: `tasks/{taskId}/pushNotificationConfigs/{configId}` → separate `task_id` and `config_id` fields
+    - Example: `tasks/{taskId}/pushNotificationConfigs/{configId}` → separate `task_id` and `id` fields
 - **✅ BENEFIT:** Simpler to implement - IDs map directly to database keys
 
 #### HTTP URL Path Simplification (#1269)
@@ -437,7 +437,7 @@ if ("text" in part) { ... }        // v1.0
 
 ```json
 {
-  "kind": "taskStatusUpdate",
+  "kind": "status-update",
   "taskId": "...",
   "contextId": "...",
   "status": {...},
@@ -470,7 +470,7 @@ if ("text" in part) { ... }        // v1.0
 
 ```json
 {
-  "kind": "taskArtifactUpdate",
+  "kind": "artifact-update",
   "taskId": "...",
   "contextId": "...",
   "artifact": {...}
@@ -648,9 +648,9 @@ Stream events changed from kind-based to wrapper-based discrimination:
 **Before (v0.3.0):**
 
 ```typescript
-if (event.kind === "taskStatusUpdate") {
+if (event.kind === "status-update") {
   handleStatusUpdate(event);
-} else if (event.kind === "taskArtifactUpdate") {
+} else if (event.kind === "artifact-update") {
   handleArtifactUpdate(event);
 }
 ```
--- a/docs/whats-new-v1.md
+++ b/docs/whats-new-v1.md
@@ -83,7 +83,6 @@ The v1.0 release focuses on four major themes:
 **v1.0 Changes:**
 
 - **✅ RENAMED:** Operation now **GetTask**
-- **✅ NEW:** `createdAt` and `lastModified` timestamp fields added to Task object
 - **✅ CLARIFIED:** More precise specification of history inclusion behavior
 - **✅ NEW:** Task object now includes `extensions[]` array in messages and artifacts
 - **✅ CLARIFIED:** Authentication/authorization scoping - servers MUST only return tasks visible to caller
@@ -155,7 +154,6 @@ The v1.0 release focuses on four major themes:
 **v1.0 Changes:**
 
 - **✅ RENAMED:** Operations now **CreateTaskPushNotificationConfig**, **GetTaskPushNotificationConfig**, **ListTaskPushNotificationConfigs**, **DeleteTaskPushNotificationConfig**
-- **✅ NEW:** `createdAt` timestamp field added to PushNotificationConfig
 - **✅ CLARIFIED:** Push notification payloads now use StreamResponse format
 - **✅ BREAKING:** model changed for all methods, with TaskPushNotificationConfig flattened
 
@@ -427,11 +425,6 @@ if ("text" in part) { ... }        // v1.0
 
 ### PushNotificationConfig Object
 
-**Added Fields:**
-
-- ✅ `configId`: Unique identifier for the configuration
-- ✅ `createdAt`: Timestamp - Configuration creation time
-
 **Modified Fields:**
 
 - ✅ `authentication`: Enhanced PushNotificationAuthenticationInfo structure
@@ -456,7 +449,7 @@ if ("text" in part) { ... }        // v1.0
 
 ```json
 {
-  "taskStatusUpdate": {
+  "statusUpdate": {
     "taskId": "...",
     "contextId": "...",
     "status": {...}
@@ -468,7 +461,7 @@ if ("text" in part) { ... }        // v1.0
 
 - ⛔ **REMOVED:** `kind` discriminator
 - ⛔ **REMOVED:** `final` boolean field (stream closure indicates completion instead)
-- ✅ **NEW PATTERN:** Event type determined by JSON member name (`taskStatusUpdate` or `taskArtifactUpdate`)
+- ✅ **NEW PATTERN:** Event type determined by JSON member name (`statusUpdate` or `artifactUpdate`)
 - ✅ **CLARIFIED:** Terminal state indicated by protocol-specific stream closure mechanism
 
 **TaskArtifactUpdateEvent:**
@@ -488,7 +481,7 @@ if ("text" in part) { ... }        // v1.0
 
 ```json
 {
-  "taskArtifactUpdate": {
+  "artifactUpdate": {
     "taskId": "...",
     "contextId": "...",
     "artifact": {...},
@@ -500,7 +493,7 @@ if ("text" in part) { ... }        // v1.0
 **Changes:**
 
 - ⛔ **REMOVED:** `kind` discriminator
-- ✅ **NEW PATTERN:** Wrapped in `taskArtifactUpdate` object
+- ✅ **NEW PATTERN:** Wrapped in `artifactUpdate` object
 - ✅ **NEW:** `index` field indicates artifact position in task's artifacts array
 
 ### OAuth 2.0 Security Updates (#1303)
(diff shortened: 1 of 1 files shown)


chopmob-cloud:
Verified this against the normative sources. The PR is correct on everything it changes and worth landing. There is one thing it leaves behind that belongs in the same fix, because it fails in exactly the same way.

## The changes here are right

Against `specification/a2a.proto` on current main:

* `StreamResponse` names its oneof members `status_update` and `artifact_update`, so ProtoJSON emits `statusUpdate` and `artifactUpdate`. The renames in this PR match.
* `message Task` carries `id`, `context_id`, `status`, `artifacts`, `history`, `metadata`. No timestamps, so removing `createdAt` and `lastModified` is correct.
* `TaskPushNotificationConfig` carries `tenant`, `id`, `task_id`, `url`, `token`, `authentication`. No timestamp there either, so removing that `createdAt` is correct.
* `TaskStatus.timestamp` does exist (`google.protobuf.Timestamp timestamp = 3`), so re-pointing the ISO 8601 line at it rather than deleting the line outright is the right call.

The PR is also right to leave the `kind` snippets in place, which is worth saying plainly since the issue raises them:

> (The same passages also show a `"kind"` discriminator at L447 and L480, inside snippets whose surrounding prose says `kind` was *removed* in v1.0.)

Those two blocks are labelled **v0.3.0:**, and the `Changes:` list noting the removed `kind` discriminator sits below both of them, describing the move to v1.0. So a `kind` field is correct there, and removing it would have been the wrong fix.

## What it leaves behind: those v0.3 examples carry the wrong `kind` values

The v0.3 side of the examples is wrong on its own terms, and this PR keeps it wrong.

v0.3.0 never used `taskStatusUpdate` or `taskArtifactUpdate` as `kind` values. From `types/src/types.ts` at tag `v0.3.0`, which is the source `specification/json/a2a.json` is generated from (`specification/json/README.md`: "DO NOT EDIT `a2a.json` directly. Make all edits in `types/src/types.ts`"). Excerpted, with the intervening fields elided:

```ts
export interface TaskStatusUpdateEvent {          // L519
  ...
  /** The type of this event, used as a discriminator. Always 'status-update'. */
  readonly kind: "status-update";                 // L525
  ...
}

export interface TaskArtifactUpdateEvent {        // L542
  ...
  /** The type of this event, used as a discriminator. Always 'artifact-update'. */
  readonly kind: "artifact-update";               // L548
  ...
}
```

The generated schema at the same tag carries those through as `"const": "status-update"` and `"const": "artifact-update"`.

Worth adding that this is not a value that went stale. It was never correct. I checked every released tag rather than sampling: the constants are `status-update` and `artifact-update` in each of v0.2.0 through v0.3.0, and the string `taskStatusUpdate` occurs zero times in the published schema of any released version, v0.1.0 included.

The guide uses the wrong values in four places, all still present at this PR's head:

* L440, `"kind": "taskStatusUpdate"` in the v0.3.0 JSON example
* L473, `"kind": "taskArtifactUpdate"` in the v0.3.0 JSON example
* L651 and L653, `event.kind === "taskStatusUpdate"` and `event.kind === "taskArtifactUpdate"` in the **Before (v0.3.0)** TypeScript snippet

The correct values are `status-update` and `artifact-update`.

I checked every `kind` value in the document against the v0.3.0 schema rather than only these, to be sure the fix is scoped correctly. The Part discriminators (`text`, `file`, `data`) are all correct as written. These four are the only wrong ones.

This matters for the same reason the v1.0 half did, and it fails the same way. Any v0.3 branch written or kept from that snippet tests `event.kind` against a string no conformant v0.3 server has ever sent, so the branch never fires and the events are silently dropped. No error, just a stream that appears to produce nothing.

The sharpest case is the one a migration guide reliably produces: a client that has to speak both versions during the transition. That code is written straight from this document, with the Before snippet as the v0.3 arm and the After snippet as the v1.0 arm. Before this PR, both arms were wrong. After it, the v1.0 arm is right and the v0.3 arm still is not.

## One smaller residual

L187 still gives the compound-ID example `tasks/{taskId}/pushNotificationConfigs/{configId}` and says it maps to separate `task_id` and `config_id` fields.

There is no `config_id` in the proto. `TaskPushNotificationConfig` names it `id` (field 2), alongside `task_id` (field 3). That is the same `configId` claim this PR already removes from the PushNotificationConfig field list, so it is worth taking out in the same pass for consistency.

With those four `kind` values corrected and that line adjusted, the guide reads clean against both the v0.3.0 schema and the v1.0 proto.

hxaxd:
Hi @darrelmiller - this is a small update aligning the v1 migration guide with the current proto, all checks are green, and I'd greatly appreciate your review when you have a moment. Thank you!

hxaxd:
Hi @darrelmiller - this is a small update aligning the v1 migration guide with the current proto, all checks are green, and I'd greatly appreciate your review when you have a moment. Thank you!

keraron (approved):
LGTM :)

hxaxd:
Hi @keraron — you approved this PR a week ago and all checks are green (mergeStateStatus: CLEAN), but it looks like it slipped through and hasn't been merged. Could you merge it when you get a chance? Thanks!

PR #2068, opened 2026-07-18
docs(spec): change SubscribeToTask HTTP method from POST to GET

### LINKED ISSUE #2063: SubscribeToTask HTTP method mismatch between proto and spec
There is a mismatch between the .proto file and the A2A specification regarding the HTTP method for `SubscribeToTask`.

The protobuf definition uses GET:
```
Protocol Buffers
  option (google.api.http) = {
    additional_bindings: {
    }
  };
}
```
However, the [A2A Specification](https://a2a-protocol.org/latest/specification/#113-url-patterns-and-http-methods) indicates it should be a POST request.

kuangmi-bit:
The proto definition (a2a.proto) uses GET with gRPC transcoding annotation for SubscribeToTask, but the specification document had POST in two places.

## Changes

- section 7.1 URL Reference table: `POST /tasks/{id}:subscribe` → `GET /tasks/{id}:subscribe`
- section 11.3.2 URL Reference list: same fix

## Evidence

Proto uses GET:
```proto
  option (google.api.http) = {
  };
}
```

GET is also the semantically correct method — SubscribeToTask is a read-only streaming operation (SSE), not a state-changing request.

Closes #2063

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1166,7 +1166,7 @@ When an agent supports multiple protocols, all supported protocols **MUST**:
 | Get task                        | `GetTask`                          | `GetTask`                          | `GET /tasks/{id}`                                       |
 | List tasks                      | `ListTasks`                        | `ListTasks`                        | `GET /tasks`                                            |
 | Cancel task                     | `CancelTask`                       | `CancelTask`                       | `POST /tasks/{id}:cancel`                               |
-| Subscribe to task               | `SubscribeToTask`                  | `SubscribeToTask`                  | `POST /tasks/{id}:subscribe`                            |
+| Subscribe to task               | `SubscribeToTask`                  | `SubscribeToTask`                  | `GET /tasks/{id}:subscribe`                             |
 | Create push notification config | `CreateTaskPushNotificationConfig` | `CreateTaskPushNotificationConfig` | `POST /tasks/{id}/pushNotificationConfigs`              |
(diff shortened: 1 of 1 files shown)


jstar0:
Linking the earlier opposite-direction work so the two proposals are evaluated
together: #1574 is the triaged `TSC Review` issue, and #1891 changes the proto
binding to POST. The discussion on #1891 records that JavaScript, Go, and Java
use POST, while Python and Rust use POST canonically with GET retained as
legacy; it also traces the POST form in the written specification back to
v0.2.6. #1891 has an approval, but the protocol direction still appears to need
a maintainer/TSC decision. It would be worth resolving GET versus POST there
before either change lands.

kuangmi-bit:
@jstar0 apologies — I missed your 07-19 comment until today and should have responded weeks ago. Thanks for the pointer to #1574 and #1891; I've now read both in full.

Agreed this needs the direction decided before either change lands, and to be clear about what #2068 is and isn't: it is purely a **consistency fix** — the proto (the normative reference) declares GET via the gRPC transcoding annotation, and the written spec contradicted it in two spots. It takes no position on which binding is *correct*; it only removes an internal contradiction.

#1891 makes the stronger claim, and the evidence you traced there is compelling: JS/Go/Java already bind POST, Python/Rust treat POST as canonical with GET as legacy, and the POST form in the written spec dates back to v0.2.6. If the maintainer/TSC decision lands on POST, our spec-side GET alignment becomes moot and I'll close #2068 to keep the queue clean. If the decision is to keep GET, this PR stands as the consistency fix.

Happy to rebase, adjust, or close on the outcome — no strong feelings on my side; the direction call belongs to the TSC process.

PR #2077, opened 2026-07-23
docs: fix hybrid agent grammar

scosemicolon:
Fixes a small grammar issue in the Hybrid Agents paragraph on the Life of a Task page.

This keeps the existing explanation intact and only changes "then generate a task" to "then generates a task" so the sentence reads correctly.


msampathkumar (approved):
LGTM

PR #2083, opened 2026-07-29
docs(spec): fix missing contextId in streaming example

juanferrub:
## What changed
- add `contextId` to the streaming task example in section 6.2
- keep the SSE example aligned with the documented event/task shape

## Why
The streaming example omitted `contextId` even though the spec requires it for task/event continuity. This makes the example easy to copy incorrectly.

## Root cause
The prose and field definitions were updated, but the SSE example in `docs/specification.md` was left behind.

## Validation
- ran `bash scripts/sort_spelling.sh`
- ran `npx markdownlint-cli 'docs/**/*.md' --config '.github/linters/.markdownlint.json' --fix`
- ran `git diff --check`

Fixes #1683


msampathkumar (approved):
Valid request as per https://a2a-protocol.org/latest/specification/#411-task

sugandhrakha (approved):
LGTM.

Issue #1991, opened 2026-06-25
[Epic] Coherent Task History — gaps in semantics, querying, and observability


Tehsmash:
## Coherent Task History — gaps in semantics, querying, and observability

The A2A protocol has a `Task.history` field and a `historyLength` query parameter, but as the protocol has matured a cluster of related gaps and ambiguities have surfaced in separate issues. This epic tracks them together so they can be addressed as a coherent unit.

### 1. What belongs in `history`?

`Task.history` currently holds only `Message` objects. There is no way to represent the interleaving of messages and artifacts in chronological order.

- **#982** — [Feat]: Integrate artifacts into task `history`

### 2. Messages lack timestamps

`Message` objects carry no timestamp. Clients that reconstruct a conversation log (e.g. a chat UI), or that reconnect after a disconnect, have no reliable way to order or date-stamp turns. `Task.status.timestamp` reflects only the latest status change, not the history.

- **#301** — Feature Proposal: Provide Timestamped Task History via `statusHistory`

### 3. History query controls are too blunt

The existing `historyLength` integer is all-or-nothing for most real scenarios: either you get everything (bandwidth cost grows with each turn) or you miss context. There is no way to ask for messages *after a known message ID*, or to page through status transitions.

- **#274** — How to Use History in Task/Send
- **#1183** — [Feat]: Add `historyMode` to control returned history in `MessageSendConfiguration` and `TaskQueryParams`

### 4. No streaming event when `history` changes

The protocol defines `TaskStatusUpdateEvent` (state transition) and `TaskArtifactUpdateEvent` (artifact produced), but has no corresponding event for a message being appended to `Task.history`. Observers that are not the sender have no way to learn that the conversational history changed — particularly when a follow-up message arrives without triggering a state transition.

- **#1253** — [Question]: Is it valid to create a `TaskStatusUpdateEvent` with the same status but update the message?
- **#1635** — [Protocol Question] Data loss during SSE reconnection — should resubscribe replay missed events?
- **#1811** — [Feat]: Add `TaskMessageUpdateEvent` to notify observers when messages are added to task history

### 5. Injecting conversation history as input

There is no standard way to send a pre-existing message history (e.g. from a chat completions-style API) into `SendMessage`. Clients bridging stateless LLM APIs through A2A, or passing context across tasks in the same context, currently have no protocol-level mechanism to do so.

- **#602** — [Feat]: Include list of messages for the agent to initiate a task in `MessageSendParams` instead of only one
[...]

Issue #1992, opened 2026-06-25
[Epic] Multi-turn interaction gaps — state acceptance rules, interruptibility, and task lifecycle


Tehsmash:
## Multi-turn interaction gaps — state acceptance rules, interruptibility, and task lifecycle

Section 3.4 of the A2A specification defines the foundations of multi-turn interaction: `contextId` and `taskId` semantics, the `INPUT_REQUIRED` pattern, follow-up message references, and context inheritance. However, several practical scenarios are left unspecified. This epic tracks the gaps.

### 1. No normative rules for which states accept a follow-up message

§3.4.3 states that clients *can* send additional messages with a `taskId` reference, but the spec does not define:

- Which task states MUST accept a follow-up message
- Which states MUST reject one, and with which error
- Whether sending to a `WORKING` or `SUBMITTED` task is valid or an error

Two open proposals take opposing positions on this, and the tension needs a TSC decision:

- **#1027** proposes that follow-up messages MUST only be accepted in `INPUT_REQUIRED`. Sending to any other non-terminal state (e.g. `WORKING`) MUST return an `UnsupportedOperationError` ("Task not accepting input"). This makes the protocol deterministic.
- **#1022** proposes the opposite: agents should be *able* to accept messages while actively `WORKING` (mid-execution steering, as in long agentic loops), advertised via a new `guidable` capability flag on the Agent Card.

Without a normative answer here, client and agent implementations diverge.

- **#1022** — [Feat]: Support being able to send messages to actively running tasks
- **#1027** — [Feat]: Clarify multi-turn message acceptance requirements for `message/send`

### 2. No mechanism to re-engage a task in a terminal state

Once a task reaches `COMPLETED`, `FAILED`, or `CANCELED`, the spec provides no way to continue within the same task context. Conversational agents frequently need to handle follow-ups to just-completed actions ("actually, move that meeting to 16:00"). Today clients must start a fresh task, losing direct task linkage.

- **#1095** — [Feat]: Task Revival Proposal

### 3. No agent-initiated handoff from synchronous to asynchronous delivery

Clients must choose their delivery strategy (`message/send`, `message/stream`, push) before an agent has enough information to know how long a task will take. There is no mechanism for an agent to signal mid-stream: *"this will take much longer than expected — please switch to async polling or push."*

- **#1960** — [Feat]: Agent-Initiated Transition from Synchronous to Asynchronous Task Delivery

### 4. Cancellation of in-progress tasks is underspecified

[...]

PR #2084, opened 2026-07-30
docs: add ADR-002 for coherent task history (timeline + artifacts)

herczyn:
## Summary

Proposes **ADR-002: Coherent Task History (Timeline + Artifacts)** to capture the design decision behind #1991 before it is finalized, and to record the "additive v1.1 vs. event-sourced v2" fork explicitly.

The ADR follows `adrs/adr-template.md` and proposes an **additive, backward-compatible** model for v1.1:

- A role-tagged **`timeline`** collection (client + agent messages + progress updates), ordered by **`generation`** (from #1794/#1810) so it stays reconstructable in bidirectional/multi-turn tasks.
- A separate **`artifacts`** collection annotated with **`start_generation`/`end_generation`**, so large/streamed artifacts interleave with the timeline without flooding it with per-chunk events.
- Optional **`elicitations`** collection (`waiting | blocked | resolved`) to model input-required additively, without changing existing task states.
- Two forward-compatible hooks toward a possible v2 unification: **client-written (bidirectional) artifacts** and an origin/**`role`** signal on `Artifact`.
- **`TaskStatus.message`** soft-deprecated (kept populated with current SDK logic through v1.x for existing consumers; removed in 2.0).

The full **event-sourced / unify `Message`+`Artifact`** direction is recorded as the motivation for a **future v2 ADR** rather than adopted now, to keep v1.1 shippable.

Status is **Proposed** and Decision Makers is the **TSC** — opening this so the framing and considered options can be reviewed/adjusted before promotion to Accepted.

Related: #1991, #1794, #1810

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -327,6 +327,7 @@ pyupgrade
 qwq
 rcm
 regen
+replayable
 repomapr
 reportgen
 reposted
--- a/adrs/adr-002-coherent-task-history.md
+++ b/adrs/adr-002-coherent-task-history.md
@@ -1,0 +1,237 @@
+# ADR-002: Coherent Task History (Timeline + Artifacts)
+
+**Status:** Proposed
+
+**Date:** 2026-07-30
+
+**Decision Makers:** Technical Steering Committee (TSC)
+
+**Technical Story:** [#1991 Coherent Task History](https://github.com/a2aproject/A2A/issues/1991) (v1.1-candidate); builds on [#1794](https://github.com/a2aproject/A2A/issues/1794) / [#1810](https://github.com/a2aproject/A2A/pull/1810) (`generation` ordering primitive)
+
+## Context
+
+A2A exposes what happened during a task through several disconnected places. The
+latest agent status message lives in `TaskStatus.message`; client-sent messages,
+prior agent messages, progress updates, and artifacts are surfaced separately,
+with no single ordered view. A consumer cannot reliably reconstruct the sequence
+of an interaction, and `TaskStatus.message` gets overloaded as a lossy, de-facto
+history channel.
+
+This matters most for bidirectional / multi-turn interaction. When a task carries
+many client and agent messages (including streamed input and output), the
+fragmented representation is no longer reconstructable — and the ordering numbers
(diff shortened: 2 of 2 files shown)


ToddSegal (approved):

vinoo999:
Sorry, meant to get this down earlier...

I generally support the core a `timeline` of events ordered by `generation`, and getting off the `TaskStatus.message` overload. My hesitation is that the ADR goes beyond history into a few things I don't think it should be deciding here - and I either disagree or feel they are underspecified as part of this ADR:

1. Artifacts as input question. Adding `role` to `Artifact` and making it client-writable only makes sense if there are client operations that join it and/or Message/Artifact is not to be unified in the future. It may enable bidi but I think this part can be deferred to that bidi conversation, no?

2. Deprecating `TaskStatus.message` needs replacement events and feel this part is underspecified as part of the ADR (can defer to PR otherwise). That field rn carries both progress notifications and elicitation text. This ADR gives us the read side (timeline, elicitations) but not the write side/the event that actually delivers a progress update that doesn't change state. Today that's a same-state `TaskStatusUpdateEvent`. Is it now a `role:agent` timeline entry via #1811? And what carries an elicitation + its `requestMessageId`? Through v1.1 we'd be double-writing to `TaskStatus.message` *and* the timeline which may be okay but need semantics on what goes where and how. i.e. is there a `TaskStatusUpdateEvent` that puts something on the Task? For a client in the StreamMessage workflow, what are they able to consume?

3. In the ADR, I'm a little confused on where `generation`, `startGeneration`, `endGeneration` will go: on the raw `Artifact` or on each event? For Artifacts that span multiple Tasks and/or multiple context ids we do not get interleaving and a context level, but that's possibly okay to be out of scope and can be deferred to #1992. Assuming #1810 is merged, I assume this ADR is saying that `generation` is on each Event, `currentGeneration` on Task and start and end on Artifact to provide appropriate sequencing in the Task? 

Nit: we'll need to describe how `historyLength` is treated and what `includeArtifacts` may mean for the payload...

I'm non-binding but voting against as bundled.

darrelmiller:
I'm supportive in spirit of this ADR.  I think there may be a few details that we can iron out when doing the PR on the spec.
I largely agree with @vinoo999  that narrowing the scope might be helpful

PR #2088, opened 2026-07-30
Design Doc : Hardware-Rooted Remote Attestation Extension

amolr:
This extension lets two agents each prove to the other that they are genuine,
unmodified code running inside a verified Trusted Execution Environment (TEE),
with the proof **bound to the live channel** and to the agent's **published
Agent Card identity**. It is built on the IETF RATS architecture
([RFC 9334](https://www.rfc-editor.org/rfc/rfc9334)) and reuses existing
verifiers (Microsoft Azure Attestation, Google Confidential Space, Intel Trust
Authority). It introduces **no new verifier** and **no core-protocol change**.

The extension is **opt-in** and **data-only** in the sense that it composes with
the existing A2A trust stack — Agent Card JWS signatures (§8.4), identity, and
authorization efforts — rather than forking them. An agent that ignores the
extension still interoperates.

### DIFF
--- a/docs/topics/hardware-attestation.md
+++ b/docs/topics/hardware-attestation.md
@@ -5,11 +5,7 @@
     extension. It describes an end-to-end flow, security considerations, and an
     example transcript so reviewers have something concrete to evaluate. It does
     **not** add core types or enums, and it does **not** change the core
-    protocol. The normative specification and reference implementation are
-    expected to live in a separate `experimental-ext-*` repository once the
-    proposal secures maintainer sponsorship, per the
-    [Extension and Binding Governance](extension-and-binding-governance.md)
-    process.
+    protocol. 
 
     Tracking issue:
     [#2078](https://github.com/a2aproject/A2A/issues/2078).
--- a/docs/topics/hardware-attestation.md
+++ b/docs/topics/hardware-attestation.md
@@ -1,0 +1,595 @@
+# Hardware-Rooted Remote Attestation Extension (Design Draft)
+
+!!! warning "Non-normative design draft"
+    This document is a **non-normative design draft** for a proposed A2A
+    extension. It describes an end-to-end flow, security considerations, and an
+    example transcript so reviewers have something concrete to evaluate. It does
+    **not** add core types or enums, and it does **not** change the core
+    protocol. The normative specification and reference implementation are
+    expected to live in a separate `experimental-ext-*` repository once the
+    proposal secures maintainer sponsorship, per the
+    [Extension and Binding Governance](extension-and-binding-governance.md)
+    process.
+
+    Tracking issue:
+    [#2078](https://github.com/a2aproject/A2A/issues/2078).
+
+## Summary
+
+This extension lets two agents each prove to the other that they are genuine,
+unmodified code running inside a verified Trusted Execution Environment (TEE),
+with the proof **bound to the live channel** and to the agent's **published
+Agent Card identity**. It is built on the IETF RATS architecture
+([RFC 9334](https://www.rfc-editor.org/rfc/rfc9334)) and reuses existing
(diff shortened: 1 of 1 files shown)


Silentpartnercoding:
I reviewed the draft around checks 3, 5, and 6. Two claims currently look stronger than the transcript can prove.

First, binding `identity_key_hash` into `report_data` proves that the measured TEE included those public-key bytes in its quote. It does not by itself prove possession of the corresponding private key, so check 5 cannot yet say the TEE “proved it holds” that key. At minimum I would add an `identityProof` signature or key-confirmation step over the canonical binding object plus the Attestation Result digest, verified against the Agent Card key before any resource release. If a platform profile cannot prove that the signing operation/key custody is inside the measured TEE, the portable claim should be the narrower “quote-bound key identifier,” not “in-TEE key custody.”

Second, check 6 allows the Passport verifier to return a yes/no verdict while also saying the relying party compares measurements with its own `referenceValues`. Those cannot both be true when the signed result omits the measurements and the identity of the appraisal policy/reference set. RFC 9334 separates the Verifier's appraisal of Evidence from the Relying Party's appraisal of Attestation Results; the result has to carry enough signed material for that second appraisal. I think the extension needs two explicit modes:

1. **Claims mode:** the signed result carries the required measurement/TEE claims and the relying party evaluates its local reference values.
2. **Delegated-appraisal mode:** the relying party explicitly trusts the verifier's appraisal and requires signed `appraisalPolicyDigest` and `referenceValuesDigest` values that match its local allow-list.

A bare `pass` from an allowed verifier should not be enough when it was produced under a different or unknown policy.

Two negative vectors would pin these boundaries cleanly:

- a valid quote binds the Agent Card public-key hash, but the peer cannot complete the live proof-of-possession transcript ⇒ reject and release nothing;
- a correctly signed `pass` comes from an allowed verifier but names an unexpected appraisal-policy/reference-values digest ⇒ reject and release nothing.

These additions preserve the draft's Passport topology and fail-closed flow; they make the identity and local-policy claims match what the relying party can actually verify.

imran-siddique (changes requested):
The RATS separation and fail-closed intent are good, but I don't think the current data-only transcript can yet guarantee the two properties it claims.

1. **The channel binding assumes a stable, exporter-accessible TLS connection that A2A does not guarantee.** The challenge and result are carried in application messages, but a client, HTTP/2 pool, retry, load balancer, or L7 terminator may put them on different TLS connections. In many deployments the agent process cannot access the external TLS exporter at all. The proposed HPKE fallback is only mentioned, not defined; merely carrying a binding value inside HPKE does not establish which A2A peer, request, and attested key the inner channel belongs to.

   Please define a binding profile with an exact exporter label/context and connection-lifetime requirements, plus explicit retry/reconnect behavior. For deployments with TLS termination, specify an authenticated inner-channel transcript that binds the challenge, Agent Card identity, attested key, both ephemeral keys, and A2A request/task identity. A proof generated for connection X and returned or retried on connection Y must be a negative test and fail closed.

2. **`Message.metadata` is too late to enforce “process nothing before attestation” unless the protocol defines a pre-task gate.** Receiving an `AttestationChallenge` through the normal message path may already create a task, invoke routing/middleware, or expose message content to the untrusted workload. An opt-in data object ignored by an implementation also cannot guarantee the stated rejection semantics or prevent downgrade before protected data is processed.

   Please define which existing operation carries a non-sensitive challenge without starting normal task execution, and a normative state machine such as `unattested -> challenged -> attested -> task-eligible`, including behavior for streaming, retries, redirects, token expiry, and re-attestation. If existing A2A operations cannot provide that gate, this needs the normative hook currently left as an open question rather than claiming the extension is fully data-only.

I agree with the existing review on proof of possession for the Agent Card key and on separating claims mode from delegated appraisal; those are also required. With these four boundaries specified, the design will have a much more implementable security contract.

Issue #71, opened 2025-04-10
Authentication and Authorization RFC?


robomotic:
Hi there,
I have been working on a similar protocol already but this makes more sense to have a larger support base.
I would like to propose an extension to the protocol to allow agents to authenticate with each other and and also authorization based on a delegation model.
Should I write an RFC of some sort but where?
Cheers.

one-kash:
to maintain the same access for user as is it outside the agentic framework, I assume authentication and authorization will be done at the user level and not use some kind of service user.

robomotic:
I think we have to be careful to avoid using the GCP terminology as "service account" but work on first principles.
I am going to write down a practical example about a simple multi party interaction.
The balance is to give the user controls over the possible actions and only be called when an exception occurs that requires higher privilege.
Anyway the best will be to mockup a practical example. Give me a day to rewrite it with the A2A protocol.
Cheers.

DJ-os:
The repo is new and we are woking on a process to review and accept RFCs. 
In the meantime it would be helpful to know what they are proposing, if you can elaborate and explain your proposal in a doc with some practical examples as you have mentioned above.

robomotic:
Hi there,
apologies for the delays - we are in full holiday here - you can find some practical examples possible issues we will encounter within the A2A protocol.
https://github.com/robomotic/MultiAgentCommLink/blob/main/README.md
Thanks for your patience.

holtskinner:
Auth support added in 0.2

PR #2099, opened 2026-08-03
docs(spec): require verifier-side trust root for Agent Card signature verification

### LINKED ISSUE #2096: spec(§8.4.3): Agent Card signature verification permits signer-controlled trust root via jku
§8.4 defines Agent Card signing over an RFC 8785 JCS payload — a solid mechanism. But the verification procedure in §8.4.3 selects the trust root from signer-controlled input, which undermines the authenticity guarantee the signature is meant to provide.

**Where.** §8.4.2 lists `jku` as a protected-header field the signer populates. §8.4.3 step 2 then instructs the verifier to:

> "Retrieve the public key using the `kid` and `jku` (or from a trusted key store)"

**The problem.** `jku` travels inside the artifact's own signer-controlled protected header. A verifier that follows step 2 literally fetches the verification key from a URL the signer chose. An attacker therefore signs a forged card with a freshly generated key, sets `jku` to their own JWKS, and the card verifies — the signer has nominated its own trust root (CWE-863, trust-boundary violation). The "(or from a trusted key store)" alternative is offered with equal weight and is not required, so the unsafe path is spec-conformant. The §8.4.3 Security Considerations cover HTTPS retrieval, key expiry/revocation, and multi-signature rotation, but not trust-root selection.

**Already validated downstream.** a2aproject/a2a-go#368 adopted a normative `KeyResolver` constraint for exactly this (commit `f6229c312`): *"The trust root MUST be selected by verifier-side policy. Implementations MUST NOT use the signer-supplied `jku` to select or fetch the trust root… `kid` MAY be used to select among keys enrolled out-of-band through a verifier-controlled path… The same constraint applies to `x5u`."* The spec should carry the same constraint so every SDK is safe by default, not just this one.

**Proposed normative text for §8.4.3:**

> Key discovery MUST be controlled by verifier-side policy. A verifier MUST NOT use a signer-supplied `jku` (or `x5u`, if later added) from the protected header to select or establish the trust root. `kid` MAY be used to select among keys already enrolled through a verifier-controlled path (a configured JWKS endpoint, DANE/TLSA anchor per #2043, or a trusted key store). A verifier that resolves keys by URL MUST constrain the target to a verifier-side allowlist.

Optionally, reclassify `jku` in §8.4.2 as an informational hint rather than a retrieval instruction. This is complementary to the identity/anchoring proposals in #1672 and #2043 — it fixes the verification procedure those mechanisms all rely on.

sanskarjaiswal2001:
## Summary
- §8.4.3 step 2 instructed verifiers to retrieve the signing key using the signer-supplied `jku` header, letting the signer nominate its own trust root (CWE-863). A verifier following this literally accepts a forged Agent Card signed with an attacker-generated key, since the attacker also controls the `jku` URL pointing to it.
- Constrains key discovery to verifier-side policy (configured JWKS endpoint, trusted key store, or other verifier-controlled path); `kid` is only used to select among keys already enrolled that way.
- Reclassifies `jku` in §8.4.2 as an informational hint, not a retrieval instruction.
- Matches the `KeyResolver` constraint already adopted in `a2a-go` (referenced in the issue).
- Adds a general Trust Anchor Selection rule in §7.1 (Authentication and Authorization) so extensions that carry their own key-discovery metadata (e.g. #1628's `trust.signals[]`) have a normative clause to cite, rather than re-deriving §8.4.3's constraint. §8.4.3's Security Considerations now point at §7.1 instead of restating it.

Fixes #2096

## Test plan
- Doc-only change; no build/test suite applies


chopmob-cloud:
Verified against current `main` (the diff applies cleanly to `docs/specification.md` §8.4.2 and §8.4.3; PR head `7a40718`). This closes #2096: the unsafe path it reported is removed and the safe requirement is now normative. Walking the three changed sites:

1. §8.4.2 now labels `jku` an informational hint and points forward to §8.4.3, so a reader can no longer treat the header URL as a retrieval instruction.

2. §8.4.3 step 2 previously read "Retrieve the public key using the `kid` and `jku` (or from a trusted key store)", which let a verifier fetch the verification key from a URL the signer chose. The revised step 2 requires "verifier-side policy (a configured JWKS endpoint, a trusted key store, or another verifier-controlled path); `kid` MAY be used to select among keys enrolled through that path." That removes the signer-nominated-trust-root path.

3. The added Security Considerations bullet makes it a `MUST NOT`: a verifier must not use a signer-supplied `jku` (or similar signer-controlled URL such as `x5u`) to select or establish the trust root, and a verifier that resolves by URL MUST constrain the target to a verifier-side allowlist. That is the CWE-863 fix stated as a binding requirement rather than guidance.

So the forgery in #2096 (attacker signs a card with a freshly generated key, sets `jku` to their own JWKS, card verifies) is no longer spec-conformant. Good to see it land as a `MUST NOT` and not only as a reworded step.

One substantive point for the signing domain, offered as a follow-up rather than a blocker: the fix closes trust-root selection by URL, but `kid` remains signer-supplied and now selects "among keys enrolled through that path." If a verifier's trusted key store holds keys for more than one issuer, `kid` alone selecting the key does not bind the resolved key to the identity the card claims. An attacker who is themselves an enrolled issuer (their own key legitimately in the store) can sign a card impersonating another issuer, set `kid` to their own enrolled key, and it verifies, because the text requires only that the key be verifier-enrolled, not that `kid` map to the card's asserted identity. Worth a sentence in §8.4.3 that the verifier must bind the selected key to the expected issuer identity, not just confirm the key is enrolled. This is the "authentic wrapper around a wrong decision" case: the signature is valid, but over the wrong trust binding.

For grounding, this matches what we enforce in our own RFC 9421 / Agent Card verification path. The per-message verifier requires the caller to pin an expected identity before the key is resolved: `verify_a2a_request` takes `expected_keyid` or an `authorize_keyid` predicate and authorises the `keyid` before resolution, and raises rather than trusting an arbitrary signer-chosen `keyid` (default `require_keyid_authorization=True`). Key resolution itself is verifier-side: inline `did:key` (self-certifying, no network) is preferred, a caller-supplied cache / trusted key store is consulted first, and URL resolution is opt-in and SSRF-guarded. We do not follow a separate signer-supplied `jku`, which is the same posture this PR now writes into the spec.

Thanks for taking this through to normative text; it fixes the verification procedure that the identity and anchoring proposals all rely on.

imran-siddique:
Verified this independently at `867fe682330bba82fb67f3dc0de6730014edb9c1` after filing the same finding on #2096 without knowing this PR existed. It is correct and it is more complete than the issue thread asked for. Detail below so a maintainer has something to merge on rather than a fresh read.

**§8.4.3 step 2.** The defect was that two key-resolution paths were offered with equal weight and only one is safe, so a verifier implementing the sentence in reading order takes the unsafe one, and a conformance suite cannot fail it because the unsafe path is conformant as written. The replacement is single-path and keeps `kid` in its correct role, selecting among keys already enrolled through verifier-side policy. Nothing is left for an implementer to interpret.

**§8.4.2.** Reclassifying `jku` at the definition site, not only at the procedure site, is the part I would have missed. A reader meets `jku` in the MAY-include list first, and until this change nothing there suggested it was anything other than a retrieval instruction.

**Security Considerations.** The new bullet is placed above the HTTPS-retrieval bullet rather than appended. The existing four considerations all concern a key that has already been chosen, so leading with trust-root selection is the ordering that matters. Carrying `x5u` prospectively and requiring a verifier-side allowlist for any URL-resolving resolver both go beyond the reported issue and are right.

**Consistency with the reference implementation.** `a2a-go#368` already behaves this way: `AgentIDKeyResolver` ignores a signer-supplied `jku`, asserted from both directions by `TestAgentIDKeyResolver_IgnoresSignerJKU` (a decoy endpoint receives zero requests) and `TestAgentIDKeyResolver_UnreachableJKU_StillResolves` (resolution succeeds against a dead address). That PR has been open since 9 July. Merging this makes the specification agree with the implementation rather than the other way round.

**What this does not do**, so it is on the record and not treated as a gap in the PR: the rule lives inside §8.4.3, so the extension layer has nothing general to cite. #1628's `trust.signals[]` specifies fetching `provider.jwks` from the signal and verifying with the selected key, which is this defect with a different field name. That wants a separate change and should not hold this one.

Checks are green and the branch is mergeable. Thirty days on a three-line change that removes a CWE-863 from the spec text is the wrong state to be in.

imran-siddique:
Confirming the follow-up commit closes the gap I recorded above. Re-verified at `20d38d00`, not at the head I reviewed on 3 September.

The clause lands as **Trust Anchor Selection** in §7.1 Protocol Security, ahead of §7.2, which is the right place for it. A reader arrives at §7 long before reaching Agent Card signing, and the rule was never specific to Agent Cards. Both cross references resolve against the head: `#71-protocol-security` matches `### 7.1. Protocol Security` and `#843-signature-verification` matches `#### 8.4.3. Signature Verification`. §8.4.3's new Security Considerations bullet cites §7.1 rather than restating it, so there is one normative statement and one pointer instead of two texts that can drift apart at the next edit.

I wrote above that the extension layer had nothing general to cite and that it wanted a separate change. That is no longer true, and folding it in here was better than the sequencing I suggested.

One consequence is worth stating plainly now that it has a citation rather than an argument behind it. #1628's shared verification model, step 1, is "Fetch provider JWKS via `provider.jwks`", where `provider.jwks` is carried inside the signal being verified. Under §7.1 as written here that is not conformant. I raised the same thing on #1628 on 12 and 28 August as a design objection; with this merged it stops being my objection and becomes a MUST NOT with a section number. I am posting a note there so it does not arrive as a surprise.

Nothing here changes my read of the three original sites. Green, mergeable, and 36 days open on a five line change that removes a CWE-863 from normative text.

imran-siddique:
Not a fourth request for a review. There are three verifications on this patch now, at three different heads, from people who did not write it, and I do not think review attention is what it is short of.

Something structural that has not been said on the thread: this PR is in no queue.

The repo routes decisions through GitVote. `.gitvote.yml` at the root defines one profile, `default`, with `allowed_voters.teams: [a2a-tsc]`, a 51 percent pass threshold and `close_on_passing: true`. The `automation:` block that would open votes by file pattern is commented out, so a vote exists here only when a TSC member opens one by hand.

This PR carries no labels at all. For comparison, #2129 carries `gitvote` and `gitvote/open`, and #2114 carries `wait-for-a2a-tsc`. Not every specification edit needs a vote: #2165 changed `docs/specification.md` and merged unlabelled on 28 August, so either route is available here. What is missing is either one. 37 days open, mergeable, all eight checks green, zero reviews recorded, and nothing marking it as awaiting a decision by anybody.

So the step that moves this is a label or a `/vote` from a TSC member, not more evidence. For whoever picks it up, the state as of head `20d38d00`: the §8.4.3 step 2 rewrite removes the signer-nominated trust root, §8.4.2 reclassifies `jku` at the definition site as well as the procedure site, and the general Trust Anchor Selection rule now sits in §7.1 with §8.4.3 citing it rather than restating it. The a2a-go reference implementation already behaves this way, so merging closes a contradiction between the spec and its own reference implementation rather than opening one.

kuangmi-bit:
Implementation-side data point for the clauses this patch makes normative — not a fourth request for review.

`a2aproject/a2a-go` #368 (Agent Card JWS signing/verification) implements the verifier-side trust-root rule, at head `ea003f9`:

| clause | implemented behaviour | evidence |
|---|---|---|
| §7.1 Trust Anchor Selection | the trust root is verifier-side policy; a signer-supplied `jku` never selects or fetches it | `a2acrypto/a2acrypto.go:24-38` (KeyResolver contract), `a2acrypto/resolver_agentid.go:60-67` (`ResolveKey(kid, jku)` ignores `jku`; the comment cites CWE-863 / #2096) |
| §8.4.3 step 2 (safe key resolution) | `jku` is carried through verification for information only | `a2acrypto/verify.go:51,64` |
| §8.4.1 (signing side) | `jku` is emitted in the protected header, informational | `a2acrypto/sign.go:44` |

The canonicalization layer underneath that signature independently passes a third-party JCS edge-case corpus 10/10 byte-identical, on both decode paths (`a2aproject/a2a#2219`, corpus pinned at `sha256 a8a1a1a8…386033`).

cc @a2aproject/a2a-tsc — per imran's note above, what this patch is short of is a TSC member's label or `/vote`, not another technical verification. Adding the above as the implementation-conformance half of that decision.

sanskarjaiswal2001:
Hey @msampathkumar , sorry for the direct ping. I noticed you've been triaging a lot of the PRs here lately and wanted to ask for help with this one.

It's been open since early August. Checks are green and there are no conflicts, but it has no labels or reviewers, so I think it slipped through without landing in anyone's queue. A few people have verified the change on the thread, and the a2a-go implementation already behaves this way (a2aproject/a2a-go#368), so as far as I can tell it just needs someone to pick it up.

Could you point it in the right direction? A wait-for-a2a-tsc label, a vote, or whatever makes sense. And if there's anything I should change first, just let me know. Thanks a lot!

eriknewton:
Supporting this. A verifier that fetches its trust root from a header the signer controls is not verifying. With a verifier-side root, the 8.4 example can only be checked against a published key, which is the gap #2249 reports, and the jcs_edge_v1 vectors in #2219 give an implementer a way to test the canonical-form half against it.

PR #2129, opened 2026-08-12
feat: implement coherent task history — timeline + artifact ordering (ADR-002)

### LINKED ISSUE #1991: [Epic] Coherent Task History — gaps in semantics, querying, and observability

herczyn:
Fixes #1991

## Summary

Implements the coherent task history model from **ADR-002** (#2084, tracking #1991) in the proto (`specification/a2a.proto`) and the normative spec prose (`docs/specification.md`), on top of the merged `generation` field (#1810). **Additive and backward compatible**, and deliberately **introduces no new wire event**: the timeline reuses the existing `TaskStatusUpdateEvent` for delivery.

> Elicitations were **split out** of this PR into #2149 following review feedback, so this change is now scoped to the timeline alone.

## Design (why it's non-breaking)

The `timeline` is a restructuring of the **Task snapshot**, not a new event on the wire. `TimelineEntry` is a `oneof`:

- `Message message` — client messages recorded in the task.
- `TaskStatus status` — agent status progression (state + the agent's `message`), delivered by the **existing** `TaskStatusUpdateEvent`.

This matters because a new member of the stream event union is **not** safely ignorable today: a2a-go, a2a-java and a2a-rs all hard-error on an unrecognised event (java's `StreamingEventKind` is a `sealed interface`), a2a-python raises on strict `ParseDict`, and a2a-js / python-gRPC silently drop it. Reusing an existing event avoids that entirely.

`TaskStatus.message` is **retained** (not deprecated): the "latest status message" pollution problem is solved by the timeline retaining the *full ordered* sequence of statuses rather than just the head.

## What's included

1. **`TimelineEntry`** `{ oneof { Message message; TaskStatus status } int64 generation }` + `repeated TimelineEntry timeline` on `Task`. Widens `generation` so appending a timeline entry increments it (previously only `TaskState` transitions and artifact add/update counted, so a progress message that doesn't change state wouldn't advance it).
2. **Artifact interleaving** — `start_generation` / `end_generation` on `Artifact`, so artifacts are placed relative to timeline entries without per-chunk events flooding the timeline.
3. **`history` deprecation** — the timeline supersedes it, so `Task.history` and the three `history_length` params are soft-deprecated and `timeline_length` added. Deprecated fields stay populated/honored through v1.x, removed in 2.0.
4. **Spec prose** — Task Timeline Semantics (§3.2.8), artifact interleaving, the deprecation, the widened generation mutation set, and a `TimelineEntry` object section.

## Notes for reviewers

- **Fail-open is normative**: clients **MUST** skip a `TimelineEntry` whose `oneof` kind they don't recognise (its `generation` still counts), so future entry kinds are additive.
- **Timeline generations are monotonic but not contiguous** — artifact updates also consume generation numbers, so gaps between entries are expected.
- Live delivery of *client-message* timeline entries to third-party subscribers is deferred to the bidirectional work (a new method, where both parties are 1.1 by construction).
- `Artifact.role` was dropped from this PR and deferred to bidi, per review.

Related: #2084 (ADR-002), #2149 (elicitations), #1810, #1794

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -505,9 +505,9 @@ The `generation` field on [`Task`](#411-task) is a sequentially increasing integ
 
 - Appending a [`TimelineEntry`](#418-timelineentry): an agent status update (any [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), including a progress message that does not change `Task.status.state`), or a client message added to the task's `timeline`.
 - An Artifact addition or update (any change to `Task.artifacts`).
-- An [`Elicitation`](#419-elicitation) state change (any change to `Task.elicitations`).
+- An [`Elicitation`](#419-elicitation) creation or state change (any change to `Task.elicitations`).
 
-Mutations delivered to subscribers as streaming events — agent status updates ([`TaskStatusUpdateEvent`](#421-taskstatusupdateevent)) and artifact updates ([`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent)) — **MUST** map 1:1 to exactly one emitted event whose `generation` value reflects the task's generation **after** the mutation. Other mutations (client-message timeline entries and elicitation state changes) also advance `generation` but are not necessarily delivered to every subscriber as a discrete event; clients reconcile these via [Get Task](#313-get-task) (see Event Ordering below).
+Mutations delivered to subscribers as streaming events — agent status updates ([`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), which also carry any [`Elicitation`](#419-elicitation) delta) and artifact updates ([`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent)) — **MUST** map 1:1 to exactly one emitted event whose `generation` value reflects the task's generation **after** the mutation. Client-message timeline entries also advance `generation` but are not necessarily delivered to every subscriber as a discrete event; clients reconcile these via [Get Task](#313-get-task) (see Event Ordering below).
 
 **Event Ordering and Missed-Event Detection:**
 
@@ -566,9 +566,23 @@ Because agent status entries are delivered by the existing [`TaskStatusUpdateEve
 
 <span id="329-elicitation-semantics"></span>
 
-An [`Elicitation`](#419-elicitation) is an explicit record of a request for client input. The `elicitations` field on [`Task`](#411-task) lets agents model input-required flows additively, without overloading the [`TaskState`](#413-taskstate) enum. Each elicitation has an [`ElicitationState`](#4110-elicitationstate) of `ELICITATION_STATE_WAITING`, `ELICITATION_STATE_BLOCKED`, or `ELICITATION_STATE_RESOLVED`, and MAY reference the agent [`Message`](#414-message) that requested the input.
+An [`Elicitation`](#419-elicitation) is an explicit record of a request for client input. The `elicitations` field on [`Task`](#411-task) lets agents model input-required flows additively, without overloading the [`TaskState`](#413-taskstate) enum. Each elicitation has an [`ElicitationState`](#4110-elicitationstate) of `ELICITATION_STATE_WAITING`, `ELICITATION_STATE_BLOCKED`, or `ELICITATION_STATE_RESOLVED`, and MAY reference the agent [`Message`](#414-message) that requested the input via `requestMessageId`.
 
-A task is considered to require input while any elicitation is in the `WAITING` or `BLOCKED` state. Servers MAY continue to set `Task.status.state` to `TASK_STATE_INPUT_REQUIRED` for backward compatibility; clients that understand `elicitations` **SHOULD** derive input-required status from the collection. An `Elicitation` state change is a generation-advancing mutation (see [Task Generation Semantics](#327-task-generation-semantics)).
+A task is considered to require input while any elicitation is in the `WAITING` or `BLOCKED` state. Servers MAY continue to set `Task.status.state` to `TASK_STATE_INPUT_REQUIRED` for backward compatibility; clients that understand `elicitations` **SHOULD** derive input-required status from the collection.
+
+**Creation and identity:**
+
+Elicitations are created and state-changed by the agent. An elicitation is normally created **together with the status update that requests the input**: the requesting agent message travels in that [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent)'s `status.message`, and the elicitation's `requestMessageId` points back to it.
+
+Each elicitation is identified by `elicitationId`, which is unique within the task. Elicitations are **never removed**; they only transition between states (`WAITING` and `BLOCKED` are open states, `RESOLVED` is terminal). The collection is therefore monotonic and can be merged by `elicitationId`.
+
+**Delivery:**
+
+Elicitation creations and state changes are delivered to streaming clients via the `elicitations` field on [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent) — a delta containing only the affected elicitations. No new streaming event type is introduced. A server MAY attach an elicitation delta to a status update it is already emitting, or emit a status update specifically to carry an elicitation change; such an elicitation-only update need not add a new [timeline](#328-task-timeline-semantics) entry. Clients **SHOULD** merge each delta into their view of `Task.elicitations` by `elicitationId`. The full current set is always available via [Get Task](#313-get-task).
+
+An `Elicitation` create or state change is a generation-advancing mutation and is delivered as part of the status update that carries it (see [Task Generation Semantics](#327-task-generation-semantics)).
+
+How a client indicates *which* elicitation a subsequent input satisfies is defined by the bidirectional interaction model and is out of scope for this version; agents resolve elicitations from interaction context.
 
 ### 3.3. Operation Semantics
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -503,11 +503,10 @@ As service parameter names MAY need to co-exist with other parameters defined by
 
 The `generation` field on [`Task`](#411-task) is a sequentially increasing integer maintained by the server. It is initialised to `1` when a task is created and **MUST** be incremented by exactly `1` by the server on every state-changing mutation:
 
-- Appending a [`TimelineEntry`](#418-timelineentry): an agent status update (any [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), including a progress message that does not change `Task.status.state`), or a client message added to the task's `timeline`.
+- Appending a [`TimelineEntry`](#418-timelineentry): an agent status update (any [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent) — including a progress message that does not change `Task.status.state`, or one whose `status` carries an [`Elicitation`](#419-elicitation) delta), or a client message added to the task's `timeline`.
 - An Artifact addition or update (any change to `Task.artifacts`).
-- An [`Elicitation`](#419-elicitation) creation or state change (any change to `Task.elicitations`).
 
-Mutations delivered to subscribers as streaming events — agent status updates ([`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), which also carry any [`Elicitation`](#419-elicitation) delta) and artifact updates ([`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent)) — **MUST** map 1:1 to exactly one emitted event whose `generation` value reflects the task's generation **after** the mutation. Client-message timeline entries also advance `generation` but are not necessarily delivered to every subscriber as a discrete event; clients reconcile these via [Get Task](#313-get-task) (see Event Ordering below).
+Mutations delivered to subscribers as streaming events — agent status updates ([`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), whose `status` also carries any [`Elicitation`](#419-elicitation) delta) and artifact updates ([`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent)) — **MUST** map 1:1 to exactly one emitted event whose `generation` value reflects the task's generation **after** the mutation. Client-message timeline entries also advance `generation` but are not necessarily delivered to every subscriber as a discrete event; clients reconcile these via [Get Task](#313-get-task) (see Event Ordering below).
 
 **Event Ordering and Missed-Event Detection:**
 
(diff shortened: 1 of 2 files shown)


specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
...
+  // Optional. The task `generation` at which this artifact was completed (its
+  // last chunk). Unset while the artifact is still streaming; for a single-shot
+  // artifact it equals `start_generation`.
+  int64 end_generation = 8;
+  // Optional. The originator of this artifact. Artifacts are agent outputs by
+  // default (`ROLE_AGENT`); a client-written (bidirectional) artifact is
+  // `ROLE_USER`. This preserves the input/output distinction as client-written
+  // artifacts are introduced. The mechanics of writing artifacts are specified
+  // separately (bidirectional streaming). See ADR-002.
+  Role role = 9;
```
vinoo999: As mentioned in https://github.com/a2aproject/A2A/pull/2084#issuecomment-5251467670 I'm opposed to role being added as part of this ADR. This is out of scope of what I think this accomplishes. Please defer this to bidi

specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
   google.protobuf.Struct metadata = 5;
   // The URIs of extensions that are present or contributed to this Artifact.
   repeated string extensions = 6;
+  // Optional. The task `generation` at which this artifact first appeared.
+  // Together with `end_generation`, this lets consumers interleave artifacts
```
vinoo999: 335-337 are fluff

specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
...
+  // The request has been satisfied or withdrawn.
+  ELICITATION_STATE_RESOLVED = 3;
+}
+
+// `Elicitation` is an explicit, additive record of a request for client input.
+// It lets consumers model input-required flows without overloading `TaskState`.
+// See ADR-002.
+message Elicitation {
+  // Unique identifier (e.g. UUID) for this elicitation within the task.
+  string elicitation_id = 1 [(google.api.field_behavior) = REQUIRED];
```
vinoo999: not sure about this. elicitation ids should be concrete Message Ids imo that are emitted. This should be structurally linked to a `message_id` that has been made by the agent. Optionally can include both but an elicitation functionally will be a `Message` no? so why invent new semantics when we have an identifier? i.e. use something like `referenceMessageId` supported in `SendMessage` for when `SendMessage` done to a Task in progress/input required state.

specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
...
+// occurred. Agent status entries are delivered on the wire by the existing
+// `TaskStatusUpdateEvent`, so the timeline needs no new event type. The `entry`
+// is a `oneof` so additional entry kinds can be added later; clients MUST fail
+// open and skip an entry whose `entry` kind they do not recognise (its
+// `generation` is still accounted for). See ADR-002.
+message TimelineEntry {
+  // The content of this timeline entry.
+  oneof entry {
+    // A standalone message recorded in the timeline, used for client input
+    // (`ROLE_USER`). Agent messages are carried by `status` entries.
```
vinoo999, line 341: why are agent messages carried differently than user messages?

docs/specification.md
```diff
@@ -501,14 +503,14 @@ As service parameter names MAY need to co-exist with other parameters defined by
...
 - An Artifact addition or update (any change to `Task.artifacts`).
 
-Each mutation **MUST** map 1:1 to exactly one emitted event. The `generation` value carried in [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent) and [`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent) **MUST** reflect the task's generation **after** the mutation that produced the event.
+Mutations delivered to subscribers as streaming events — agent status updates ([`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), whose `status` also carries any [`Elicitation`](#419-elicitation) delta) and artifact updates ([`TaskArtifactUpdateEvent`](#422-taskartifactupdateevent)) — **MUST** map 1:1 to exactly one emitted event whose `generation` value reflects the task's generation **after** the mutation. Client-message timeline entries also advance `generation` but are not necessarily delivered to every subscriber as a discrete event; clients reconcile these via [Get Task](#313-get-task) (see Event Ordering below).
```
vinoo999: To clarify, it seems that Artifact updates are not found in timeline entries correct but do increment generation? This would not be 1:1 both from a client side mapping and artifact mapping.

specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
...
+  // without a change to the `TaskState` enum.
+  ELICITATION_STATE_BLOCKED = 2;
+  // The request has been satisfied or withdrawn.
+  ELICITATION_STATE_RESOLVED = 3;
+}
+
+// `Elicitation` is an explicit, additive record of a request for client input.
+// It lets consumers model input-required flows without overloading `TaskState`.
+// See ADR-002.
+message Elicitation {
```
vinoo999: I'm actually unclear if this is now needed if we are putting the full TaskStatus in the TimelineEntry. You can filter on TimelineEntries with task status `input_required` to determine a list of elicitations? It is helpful to determine separation if multiple elicitations are requested but is that strictly required at the moment? Could be modeled as an extension for adoption otherwise

specification/a2a.proto
```diff
@@ -182,17 +188,35 @@ message Task {
...
   google.protobuf.Struct metadata = 6;
-  // Sequentially increasing counter incremented by exactly 1 by the server
-  // on every state-changing mutation to this task (TaskState transition or
-  // Artifact addition/update), with each mutation mapping 1:1 to one event.
+  // Sequentially increasing counter incremented by exactly 1 by the server on
+  // every state-changing mutation to this task, with each mutation mapping 1:1
+  // to one event. Mutations include: a status update recorded as a
+  // `TimelineEntry` (any `TaskStatusUpdateEvent`, including a progress message
+  // that does not change `TaskState`); an `Artifact` addition or update; and an
+  // `Elicitation` state change.
```
vinoo999: including client messages?

specification/a2a.proto
```diff
@@ -302,6 +331,74 @@ message Artifact {
   google.protobuf.Struct metadata = 5;
   // The URIs of extensions that are present or contributed to this Artifact.
   repeated string extensions = 6;
+  // Optional. The task `generation` at which this artifact first appeared.
+  // Together with `end_generation`, this lets consumers interleave artifacts
+  // with the `timeline` without the timeline carrying per-chunk stream events.
+  // See ADR-002.
+  int64 start_generation = 7;
```
vinoo999, line 322: looking at this in more detail I'm not sure if generation belongs on an Artifact. This might be a separate problem but artifacts that are modified/shared across Tasks are a sematnic that is not modeled. I think today, a new artifact is created potentially initialized with the same content as previous. Artifact becomes purely task scoped as opposed to context scoped. Not necessarily blocking here but adding start/end generation firmly puts Artifact at being task scoped

specification/a2a.proto
vinoo999: additional semantic clarity required: is elicitation agent facing? if so, how does agent 1 (client) submit a SendMessage to agent 2 (server) with an elicitation response? I think we can cut this altogether and utilize timeline entry message ids as responses that have elicitation semantics as opposed to building in this new type

vinoo999 (commented):
I think we should remove `Role` from artifacts and `Elicitation` altogether. Seeking clarification on `Artifact` generation firmly tying it to a Task and how `generation` increments from client side invocations. Curious on alternatives to `TimelineEntry` of `Message` and if all we are really missing from `History` is the idea that it can be repeated `Message` | `Event`. Artifact updates are missing from Timeline Entry which may be desirable but will lead to a non-1:1 relation in entry to generation increments.

docs/specification.md
```diff
@@ -540,6 +542,43 @@ The `ifGenerationMatch` field in [`SendMessageConfiguration`](#322-sendmessageco
...
+
+Because agent status entries are delivered by the existing [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), introducing the timeline requires **no new streaming event type**, and clients that do not understand the `timeline` field are unaffected on the wire. Live delivery of client-message timeline entries to other subscribers (for example, in bidirectional interactions) is out of scope for this version; such entries are observed via [Get Task](#313-get-task). The `oneof` in [`TimelineEntry`](#418-timelineentry) is an extension point: additional entry kinds MAY be added in future minor versions. Clients **MUST** fail open — a client that encounters a `TimelineEntry` whose `entry` is a kind it does not recognise **MUST** skip that entry rather than rejecting the timeline or aborting the stream. The entry's `generation` is still accounted for (it does not count as a gap), so ordering and missed-event detection continue to work across unknown entries.
```
Tehsmash, line 558: I'm not sure that the client message timeline event should be out of scope here, its certainly one of the issues covered by our epic. Introducing a new event for this TaskMessageUpdateEvent or something shouldn't breaks backwards compatibility as clients which receive an unknown event should ignore it (but perhaps this is a assumption too far)

specification/a2a.proto
herczyn: "an elicitation functionally will be a Message no?" not necessarily 1:1. an agent can send a message and ask for multiple data points and represent them as separate elicitations. then can also be answered separately and some of them might be blocking while others aren't. i was considering adding elicitation_id to sendmessage but i think this should be on the agent to figure out, not the caller.

specification/a2a.proto
herczyn: because currently agent ones come with task status updates and possibly elicitations and user messages don't. when we get to bidi maybe we can flatten it allowing updates from clients and pure messages from server, but for now the comms are not symmetrical.

docs/specification.md
herczyn: correct. generation bump == event sent to client != timeline entry

specification/a2a.proto
herczyn: input_required means agent is no longer waiting (is blocked). this allows you to add elicitations and continue working and potentially get a response in parallel. also one input_required might mean more elicitations, up to the agent.
how does agent 1 (client) submit a SendMessage to agent 2 (server) with an elicitation response? however they want. this allows for a clear state transitions as well which just using timeline doesn't give us

specification/a2a.proto
herczyn: it is task scoped already, no?

specification/a2a.proto
herczyn: makes sense

specification/a2a.proto
herczyn: right now client message is only at generation 0->1 change, move this discussion to bidi?

docs/specification.md
```diff
@@ -540,6 +542,43 @@ The `ifGenerationMatch` field in [`SendMessageConfiguration`](#322-sendmessageco
...
 The `generation` field was introduced in version **1.1** of this specification. The field defaults to `0` in the Protocol Buffer encoding (proto3 default for `int64`). Clients that do not read or send `generation` continue to interoperate correctly with servers that support it. Servers that have not yet implemented `generation` will return `0` on all events and responses; since a correctly-implemented server always returns `generation ≥ 1`, clients will detect a `0` value as a signal that the server does not support the field and gracefully fall back.
 
+#### 3.2.8. Task Timeline Semantics
+
+<span id="328-task-timeline-semantics"></span>
+
+The [`timeline`](#418-timelineentry) field on [`Task`](#411-task) is the coherent, generation-ordered record of the interaction on the wire. Each [`TimelineEntry`](#418-timelineentry) is a `oneof` carrying either:
+
+- a client [`Message`](#414-message) — client input recorded in the task; or
+- an agent [`TaskStatus`](#412-taskstatus) — the agent's state and `message` at a point in the interaction, delivered to subscribers by the existing [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent).
```
Tehsmash, line 552: Are we dropping the ability for an Agent to add a Message separately from a TaskStatus update?

I thought it would be nice to include this, for backwards compat we can map it so for older clients Task.status.message gets updated, but in the timeline it appears simply as a Message? Then we could add a new simple string field on TaskStatus for a human readable status update message.

Mapping would look something like:

Agent Message added to timeline -> SDK also updates TaskStatus.message, TaskStatusUpdateEvent && TimelineAdditionEvent triggered (newer clients can ignore/drop TaskStatusUpdateEvent)
 
TaskStatus changed (and/or TaskStatus.displayMessage updated) added to timeline -> TaskStatus.state and TaskStatus.message updated with displayMessage && TaskStatusUpdateEvent and TimelineAdditionEvent triggered.

So new clients will see the timeline of Message Role user, TaskStatus, Message Role Agent.
Older clients will see history + TaskStatus updating the way it used to.

docs/specification.md
```diff
@@ -540,6 +542,43 @@ The `ifGenerationMatch` field in [`SendMessageConfiguration`](#322-sendmessageco
...
+
+#### 3.2.9. Elicitation Semantics
+
+<span id="329-elicitation-semantics"></span>
+
+An [`Elicitation`](#419-elicitation) is an explicit record of a request for client input. It lets agents model input-required flows additively, without overloading the [`TaskState`](#413-taskstate) enum. Each elicitation has an [`ElicitationState`](#4110-elicitationstate) of `ELICITATION_STATE_WAITING`, `ELICITATION_STATE_BLOCKED`, or `ELICITATION_STATE_RESOLVED`.
+
+The `elicitations` field on [`Task`](#411-task) is the authoritative, current set of the task's elicitations. A task is considered to require input while any elicitation is in the `WAITING` or `BLOCKED` state. Servers MAY continue to set `Task.status.state` to `TASK_STATE_INPUT_REQUIRED` for backward compatibility; clients that understand `elicitations` **SHOULD** derive input-required status from the collection.
```
Tehsmash: personally I kind of feel like if a Client is requesting A2A-Version: 1.0 then we should return INPUT_REQUIRED if there are any elicitations and the TaskStatus.state = WORKING. 

If Client requests A2A-Version: 1.1 or newer it should see "WORKING" + elicitations list. 

This seems like an easy transformation to make between an internal storage datamodel on v1.1 and the old v1.0 protocol. 

This also makes it easy for the Agent implementations as they no longer need to know or worry about INPUT_REQUIRED as a state, they only need to know about elicitations/ v1.1 data model and the version transformation layer will handle the rest.

specification/a2a.proto
```diff
@@ -182,17 +188,35 @@ message Task {
...
+  // progression: the accumulated `TaskStatus` history. Each entry corresponds to
+  // a `TaskStatusUpdateEvent` already delivered on the wire, so no new event
+  // type is required to stream it. See ADR-002.
+  repeated TimelineEntry timeline = 8;
+  // protolint:enable REPEATED_FIELD_NAMES_PLURALIZED
+  // Optional. Outstanding and resolved requests for client input. The task is
+  // considered to require input while any elicitation is in the `WAITING` or
+  // `BLOCKED` state; this derives `INPUT_REQUIRED` additively without changing
+  // the `TaskState` enum. See ADR-002.
+  repeated Elicitation elicitations = 9;
```
Tehsmash: Perhaps these belong in TaskStatus?

docs/specification.md
herczyn: it is breaking, at least in some of the sdks - when an event comes that is not recognized the code can fail unfortunately. we should file issues it to fix it in sdks, but i don't think we can expect it to be rolled out by 1.1.

we could guard it by version but i don't like it if it's not necessary - it complicates the sdk code very much, introduces requirements on folks building their own implementations (or new sdks) and makes everything much harder to reason about (do we add 1.1 as a separate version to all ITK tests for each SDK?)

docs/specification.md
herczyn: i think we need to drop it to be be backwards compatible. we can add it as part of the bidi stream - it's a new method so we can assume both parties are on 1.1

docs/specification.md
herczyn: i was thinking input_required if at least one elicitiation is BLOCKED, but i would not have this as part of the spec but how sdks implement it. it should not be a requirement on the protocol level, just a default sdk implementation

specification/a2a.proto
herczyn: the are there but only the latest delta. here we keep the full set.

spetschulatSFDC (approved):

PR #2149, opened 2026-08-18
feat: add elicitations for additive input-required (ADR-002)

herczyn:
## Summary

Split out of #2129 so it can be reviewed on its own merits, per review feedback that the timeline change was bundled with contested extras.

Adds an explicit, additive way to model requests for client input, without overloading the `TaskState` enum.

- **`Elicitation`** `{ elicitationId, state, metadata }` and **`ElicitationState`** (`WAITING` / `BLOCKED` / `RESOLVED`).
- **`Task.elicitations`** — the authoritative, current set.
- **`TaskStatus.elicitations`** — a *delta* of what was created/changed at that status. Because it rides the status, it is delivered by the existing `TaskStatusUpdateEvent` and captured in the `timeline`, so walking the timeline reconstructs elicitation history. **No new streaming event type.**
- **`SendMessageRequest.elicitationId`** — an optional client *hint* for which elicitation a message answers.

## Why a separate type rather than reusing message ids

A single agent message may raise **several** elicitations — asking for multiple data points that can be answered separately, where some block progress and others don't. That's not 1:1 with a message, and it's what lets an agent keep working while a non-blocking request is outstanding.

## `INPUT_REQUIRED` mapping

The protocol does **not** mandate a mapping onto `TaskState`, and it is **not** version-conditioned. The expected default (typically an SDK default, so agent authors don't manage the state by hand) is: `TASK_STATE_INPUT_REQUIRED` while at least one elicitation is `BLOCKED`; `WAITING`-only leaves the state as-is (e.g. `WORKING`). Existing clients reading `Task.status.state` keep working unchanged; clients reading `elicitations` additionally see non-blocking requests.

## Responding to an elicitation

Resolution stays the agent's responsibility — it decides from content and context what has been satisfied. The `elicitationId` on `SendMessageRequest` is a hint only:

- clients are never required to set it;
- servers MAY use it, or ignore it and infer from context;
- servers **MUST NOT** reject a message solely because it is absent, unknown, or refers to an already-`RESOLVED` elicitation.

## Verification

`buf breaking` (FILE) clean; api-linter `problems: []`; protolint clean; markdownlint clean; new spec tables render via the `proto_to_table` macro. Fully additive — no fields renumbered, no types removed.

Related: #1991, #2084 (ADR-002), #2129

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -505,9 +505,9 @@ The `generation` field on [`Task`](#411-task) is a sequentially increasing integ
 
 - Appending a [`TimelineEntry`](#418-timelineentry): an agent status update (any [`TaskStatusUpdateEvent`](#421-taskstatusupdateevent), including a progress message that does not change `Task.status.state`), or a client message added to the task's `timeline`.
 - An Artifact addition or update (any change to `Task.artifacts`).
-- An [`Elicitation`](#419-elicitation) state change (any change to `Task.elicitations`).
+- An [`Elicitation`](#419-elicitation) creation or state change (any change to `Task.elicitations`).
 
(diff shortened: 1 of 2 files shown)


darrelmiller:
We need to make a call as to whether we need this in the v1.1 spec.

herczyn:
agreed in TSC meeting this should rather be handled by an extension for now

PR #2156, opened 2026-08-19
docs(spec): add HTTP base path deployment guidance

### LINKED ISSUE #2029: [Docs]: Add deployment best practices for A2A HTTP base path (e.g. /a2a)
While implementing a Spring-based A2A server integration, I ran into a question regarding HTTP endpoint organization in production deployments.

The specification defines only relative HTTP paths (e.g. /message:send, /tasks/{id}) and leaves the base URL fully implementation-defined. This provides flexibility, but also leaves ambiguity for framework authors and production setups where A2A endpoints coexist with business REST APIs, admin endpoints, and observability interfaces.

Because of this, implementers need to make their own decision whether to expose A2A endpoints at the root level or under a dedicated base path.

Example:

/message:send
/tasks/{id}

vs

/a2a/message:send
/a2a/tasks/{id}

From an operational perspective, using a dedicated base path (e.g. /a2a) simplifies:
- API Gateway routing
- security configuration (Spring Security / OAuth / mTLS boundaries)
- reverse proxy setup
- rate limiting and traffic separation
- logging and observability
- clear separation between business APIs and A2A protocol endpoints

I am not proposing any change to the protocol itself and not suggesting to make any base path mandatory.

Instead, I would like to contribute a small documentation addition describing deployment best practices for HTTP endpoints, to help implementers make consistent production decisions.

A possible recommendation could be:

"A2A endpoints MAY be exposed under a dedicated base path such as /a2a to simplify routing, security configuration, and operational management."

If this makes sense, I would be happy to contribute a PR. I’d also appreciate guidance on where such documentation would fit best (specification vs separate deployment/best-practices section).

madankumarpichamuthu:
## Description

Closes #2029.

The HTTP+JSON/REST binding (§11.3) defines only relative URL patterns (`/message:send`, `/tasks/{id}`, …) and leaves the base URL implementation-defined. This leaves framework authors and production deployments unsure whether to root A2A endpoints at the host root or under a dedicated prefix, especially where A2A endpoints coexist with business REST APIs, admin endpoints, and observability interfaces.

This adds a non-normative **§11.3.5 Base Path** subsection that:

- States implementations **MAY** serve endpoints at the root or under a dedicated base path such as `/a2a`.
- Lists the operational benefits (gateway/proxy routing, middleware attachment, rate limiting, observability partitioning).
- Clarifies that a base path is an operational/routing convenience — **not** a security or identity boundary — and that declared transport security requirements apply regardless of prefix. (This addresses the "operational isolation, not a protocol-level security boundary" framing raised in the issue discussion.)
- Notes that the advertised `AgentInterface.url` **MUST** include the base path when one is used.

No protocol change and no mandatory base path — documentation only.

## Checklist

- [x] Conventional Commit title (`docs(spec):` for `specification.md`).
- [x] Linter passes.

Fixes #2029

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -2813,6 +2813,32 @@ A2A-Extensions: https://example.com/extensions/geolocation/v1,https://standards.
 
 - `GET /extendedAgentCard` - Get authenticated extended Agent Card
 
+#### 11.3.5. Base Path
+
+The URL patterns above are defined relative to a base URL that is
+implementation-defined. The specification does not mandate where A2A endpoints
+are rooted, so implementations **MAY** serve them at the root of a host or
+under a dedicated base path.
+
+Exposing A2A endpoints under a dedicated base path such as `/a2a` (for example,
+`POST /a2a/message:send` and `GET /a2a/tasks/{id}`) can simplify operational
+concerns when A2A endpoints coexist with other APIs on the same host, including:
+
+- API gateway and reverse-proxy routing
+- attaching authentication and authorization middleware to a single prefix
+- rate limiting and traffic separation
+- logging and observability partitioning
+
+A base path is an operational and routing convenience only. It is not a
+security or identity boundary: implementations **MUST NOT** treat the presence
+of a particular prefix as authentication or authorization for a request, and
+**MUST** apply the transport's declared security requirements regardless of the
+base path in use.
+
+When a base path is used, the fully-qualified endpoint URL advertised in the
+Agent Card's [`AgentInterface`](#446-agentinterface) `url` **MUST** include that
+base path so clients resolve the correct location.
+
 ### 11.4. Request/Response Format
 
 All requests and responses use JSON objects structurally equivalent to the Protocol Buffer definitions.
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -2813,31 +2813,45 @@ A2A-Extensions: https://example.com/extensions/geolocation/v1,https://standards.
 
 - `GET /extendedAgentCard` - Get authenticated extended Agent Card
 
-#### 11.3.5. Base Path
+#### 11.3.5. Path Prefix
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -2815,6 +2815,32 @@ A2A-Extensions: https://example.com/extensions/geolocation/v1,https://standards.
 
 - `GET /extendedAgentCard` - Get authenticated extended Agent Card
 
+#### 11.3.5. Base Path
+
+The URL patterns above are defined relative to a base URL that is
```
handrews: I'm new to A2A so my apologies if I am missing something, but I would avoid the term "base URL" here because the host+base path is not a [base URI in the RFC3986 sense](https://www.rfc-editor.org/rfc/rfc3986.html#section-5.1).  If it were, the "base" path would always be [truncated](https://www.rfc-editor.org/rfc/rfc3986.html#section-5.2.2) before the A2A path was appended as the A2A path begins with `/`.

It would be more accurate to talk about prefixes and suffixes as the intent is clearly to concatenate the A2A path to the prefix.  You'll also want to forbid prefixes from ending in `/` to avoid the need to handle annoying path normalization involving `//`.

Again, if I'm missing something obvious please feel free to just say so and maybe point to where I can read up on what I missed.

madankumarpichamuthu: Thanks, you're right and I've updated the PR to fix this.

- Renamed the section to **Path Prefix** and reframed it around concatenating the operation path onto a prefix, rather than "base URL" / base-URI resolution.
- Added a note calling out that the host+prefix is **not** an RFC 3986 base URI, since resolving an operation path beginning with `/` against a base URI would replace the prefix rather than preserve it.
- Added a **MUST NOT** end-with-`/` rule on the prefix to avoid `//` in the resulting path.

Let me know if the wording still reads ambiguously anywhere.

PR #2166, opened 2026-08-22
docs: fail docs build on unresolved proto references

### LINKED ISSUE #2151: Docs build succeeds when a proto table macro cannot resolve, so the error ships to the site
The three proto table macros return their failures as page content rather than raising, so an unresolved type reference renders into the published spec and the build still exits 0. Two are on the site right now:

- §4.3.1 — `Error: Message PushNotificationConfig not found.`
- §10.4.7 — `Error: Message CreateTaskPushNotificationConfigRequest not found.`

https://a2a-protocol.org/latest/specification/

The proto defines `TaskPushNotificationConfig`; there is no `PushNotificationConfig` message. `CreateTaskPushNotificationConfigRequest` is only an RPC method name, never a defined type.

#1676 reported the first in March and #1981 fixes both, but neither surfaced on its own, because a broken reference is not a build failure.

To find them from a clean checkout:

```python
import re
spec = open('docs/specification.md').read()
proto = open('specification/a2a.proto').read()
refs = set(re.findall(r'proto_(?:enum_)?to_table\(\s*"([^"]+)"', spec))
types = set(re.findall(r'^\s*(?:message|enum)\s+(\w+)\s*\{', proto, re.M))
print(sorted(refs - types))
```

Two of the three macros also wrap the whole body in `except Exception`, so any error, not only a missing type, becomes page text.

I have a branch that raises instead, and sets `on_error_fail` on the macros plugin. Without that flag mkdocs-macros catches the exception, renders it into the page and still exits 0, so raising alone does not close this. With it, the build fails naming the file and the type.

Verified locally against `16ba526`:

| | build | §4.3.1 |
|---|---|---|
| main today | exits 0 | error string rendered |
| raise only | exits 0 | exception rendered into the page |
| raise + `on_error_fail` | exits 100, names file and type | build refuses to ship |
| guard + #1981's two lines | exits 0 | real `TaskPushNotificationConfig` table |

Happy to open it once #1981 lands, or now if you would rather review them together. Kept it off `docs/specification.md` so it stays a `ci:` change rather than a `docs(spec):` one.

sachinsharma3191:
## Summary

Fixes #2151

- Changed `proto_to_table`, `proto_enum_to_table`, and `proto_service_to_table` macros to raise `LookupError` instead of silently returning error strings when a proto type is not found
- Added `on_error_fail: true` to the macros plugin config in `mkdocs.yml`
- Unresolved references now break the docs build instead of rendering as "Error: not found" in published docs

## Test plan

- [x] Missing proto message/enum/service raises `LookupError` during build
- [x] `on_error_fail: true` causes `mkdocs build` to exit non-zero on macro errors
- [x] Valid proto references render correctly (no behavior change for happy path)

### DIFF
--- a/.mkdocs/macros.py
+++ b/.mkdocs/macros.py
@@ -62,15 +62,14 @@ def define_env(env):
         message_name: str, proto_file: str = 'specification/a2a.proto'
     ) -> str:
         """Parses a .proto file and renders a message table."""
-        try:
-            elements = _parse_proto(proto_file)
-        except FileNotFoundError as e:
-            return f'**Error:** {e}'
+        elements = _parse_proto(proto_file)
 
         # Find the specific message object
         target_message = _find_type(elements, message_name, Message)
         if not target_message:
-            return f'**Error:** Message `{message_name}` not found.'
+            raise LookupError(
+                f'Message `{message_name}` not found in {proto_file}'
+            )
 
         # Extract data
         rows = []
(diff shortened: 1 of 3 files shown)


keraron (approved):

sugandhrakha (approved):
LGTM

sachinsharma3191:
@keraron @sugandhrakha Thanks for the approvals! I've merged the latest `main` into this branch and resolved the conflict in `docs/specification.md`, keeping `main`'s updated push-notification section. All proto references in the docs still resolve and all checks pass. Could one of you please merge this when you get a chance?

PR #2172, opened 2026-08-23
docs(spec): fix HTTP method for SubscribeToTask REST binding

### LINKED ISSUE #1574: [Bug]: a2a.proto and specification divergence on subscribeToTask

sanskarjaiswal2001:
## Summary
- The Method Mapping Reference table (§5.3) listed `SubscribeToTask`'s REST endpoint as `POST /tasks/{id}:subscribe`, but the proto definition (a2a.proto) binds it to `GET`.
- `SubscribeToTask` is a read/subscribe operation (non-mutating), consistent with `GetTask` and `ListTasks`, both of which are `GET` in the same table — unlike `CancelTask`, which mutates state and is correctly `POST`.
- This aligns the doc table with the proto source of truth.

Fixes #1574

## Test plan
- [x] Confirmed via `grep` that `a2a.proto` rpc `SubscribeToTask` uses `option (google.api.http) = { get: ... }`
- [x] No other spec.md sections state the REST method for this RPC (only the §5.3 table)

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1166,7 +1166,7 @@ When an agent supports multiple protocols, all supported protocols **MUST**:
 | Get task                        | `GetTask`                          | `GetTask`                          | `GET /tasks/{id}`                                       |
 | List tasks                      | `ListTasks`                        | `ListTasks`                        | `GET /tasks`                                            |
 | Cancel task                     | `CancelTask`                       | `CancelTask`                       | `POST /tasks/{id}:cancel`                               |
-| Subscribe to task               | `SubscribeToTask`                  | `SubscribeToTask`                  | `POST /tasks/{id}:subscribe`                            |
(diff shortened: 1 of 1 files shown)


sanskarjaiswal2001:
Closing — duplicate direction of #2068, and discussion on #1891/#1574 is trending toward fixing the proto to POST instead (JS/Go/Java already use POST; Python/Rust treat POST as canonical). Deferring to that consensus.

PR #1981, opened 2026-06-24
docs(spec): Rename PushNotificationConfig to TaskPushNotificationConfig

Iwaniukooo11:
This PR resolves two macro rendering errors in `specification.md` where the old, removed `PushNotificationConfig` and `CreateTaskPushNotificationConfigRequest` messages were being called. 

Since these were flattened into `TaskPushNotificationConfig` in v1.0, this PR aligns the documentation and cross-page links (in `streaming-and-async.md`) to use `TaskPushNotificationConfig` consistently.

4.3.1 after update:

10.4.7 after update:

Fixes: #1676

PR #2188, opened 2026-08-28
ci: fail the docs build when a proto macro cannot resolve

### LINKED ISSUE #2151: Docs build succeeds when a proto table macro cannot resolve, so the error ships to the site

Ectsang:
# Description

All three proto table macros return their failures as page content instead of raising, so a reference to a type that does not exist renders into the page and the build still exits 0. That is how `Error: Message PushNotificationConfig not found.` reached the published spec. #1981 fixed the two live references today, so nothing is broken right now.

Two changes:

- `.mkdocs/macros.py` raises `ProtoReferenceError` naming the file and the type instead of returning an error string, and drops the two `except Exception` wrappers that turned any error into page text.
- `mkdocs.yml` sets `on_error_fail: true`. Without it mkdocs-macros catches the exception, renders it into the page and still exits 0, so raising alone does not close this.

Built at `f63dbb4`:

| | build |
|---|---|
| with this PR | exits 0 |
| one reference renamed to a type that does not exist | exits 100, names file and type |
| raise only, without `on_error_fail` | exits 0, traceback rendered into the page |

- [x] Follow the `CONTRIBUTING` Guide.
- [x] Make your Pull Request title follow the Conventional Commits specification and repository rules.
- [x] Ensure the tests and linter pass.

Fixes #2151 🦕

### DIFF
--- a/.mkdocs/macros.py
+++ b/.mkdocs/macros.py
@@ -40,6 +40,21 @@ TYPE_MAP = {
     'google.protobuf.Empty': 'empty',
 }
 
+
+class ProtoReferenceError(Exception):
+    """A doc macro referenced a proto type that does not exist.
+
+    Raised rather than returned as page text. These macros used to return
+    strings like ``**Error:** Message `X` not found.`` when a lookup failed,
+    which rendered the error into the published specification while the docs
+    build still reported success. Two such strings were live on
+    https://a2a-protocol.org/latest/specification/ (sections 4.3.1 and
+    10.4.7); the first was reported in #1676 and went unnoticed for months
+    because nothing failed. Raising makes an unresolved reference a build
+    error, so it cannot ship.
+    """
+
+
 # -----------------------------------------------------------------------------
 # Main Macros
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Ectsang:
Still green and still merges clean against `main` after the seven commits since this opened. The sabotage case in the description is the part worth a second of attention: with `on_error_fail` and the raise, a reference to a type that does not exist exits 100 and names the file and the type; with the raise alone it exits 0 and renders the traceback into the page. Both changes are needed to close it.

PR #2227, opened 2026-09-11
refactor: update ACTS test suite specifications to standardize error formats, include header inspection, and refine transport requirements

JakubWorek:
# Description

ACTS corpus from #1882 PR required some major adjustments.

#### Corrected against the normative A2A sources

Each of these disagreed with `docs/specification.md` or `specification/a2a.proto`.

- **§6.2 error tables rebuilt from A2A §5.4 and §9.5.** The nine A2A-specific errors now match §5.4 across all three columns. `VersionNotSupportedError` moves `-32006` → `-32009`; `InvalidAgentResponseError` (`-32006`), `ExtendedAgentCardNotConfiguredError` (`-32007`) and `ExtensionSupportRequiredError` (`-32008`) are added; the invented `StreamingNotSupportedError` and `ExtendedCardNotSupportedError` are removed. Several HTTP mappings were wrong (`409`, `405`, `415`, `406`, `501` → `400`).
- **The five standard JSON-RPC errors are split into their own table**, because A2A §9.5 defines them for the JSON-RPC binding only. A test naming one MUST now carry a `transport` filter.
- **Error type must not be inferred from status.** HTTP and gRPC statuses are not injective — A2A §11.6 has several error types sharing one status — so runners MUST identify an error by JSON-RPC `error.code` or by `google.rpc.ErrorInfo.reason`.
- **§4.1 operation table realigned with A2A §5.3**: `subscribe_to_task` is `POST` not `GET`; the push operations are `CreateTaskPushNotificationConfig` and siblings on `/pushNotificationConfigs`, not `PushNotificationConfig` on `/pushNotifications`; `get_extended_agent_card` was missing entirely.
- **`Part` is flat.** The nested `file: {url, raw, mediaType, name}` wrapper is replaced by sibling `raw` / `url` / `filename` / `mediaType` / `metadata`, matching the proto. Same correction in §10's client-test payload grammar.
- **Push-config request params are flat**, because the request message *is* a `TaskPushNotificationConfig` — `url`, `token` and `authentication` sit beside `taskId` rather than nested under a `pushNotificationConfig` wrapper.
- **§7.1 state transitions de-contradicted.** The old text listed valid transitions *and* said the rule was not a linear progression. It is now an exhaustive list of **illegal** transitions, with an explicit statement that a runner MUST NOT reject a transition for skipping an intermediate state (`SUBMITTED → COMPLETED` and `SUBMITTED → REJECTED` are both legal).
- **§16 file layout** now describes the real `tests/acts/` tree and the `.acts.yaml` extension, replacing a `v1.0/` / `v0.3/` structure that never existed.

#### Grammar additions that legalize shapes the corpus already used

The CDDL rejected constructs the corpus needed. Usage counts are unchanged by this commit — only the grammar moved.

- **`items`** (5 sites) — assert on array elements in the same map that constrains the array, which §5.2's nesting rules cannot express. JSON Schema semantics: a map applies to every element, a list positionally.
- **`one_of` over assertions**, not just scalars — this is what expresses an alternation between two response shapes.
- **`expect_stream` on a raw step** (12 sites) — raw SSE framing is a legitimate thing to test; the grammar was incomplete, not the tests.
- **`error_type` is now optional** — some requirements mandate that an operation fail without mandating which error, and such a test asserts only on `message` or `details`.
- **`details` is an assertion**, not a map of assertions — it addresses the error details *array*.

#### New capabilities

- **`expect.headers`** — the only way to verify response metadata that never reaches the body: content types, caching directives, authentication challenges. Header names are case-insensitive. A test using it MUST exclude `grpc` by `transport` filter, since gRPC carries metadata rather than an HTTP header block, and a runner MUST NOT treat such an assertion as satisfied on a binding that cannot produce the value.
- **`expect_webhook` — a new step kind (§7.2).** A push notification is a request the **SUT** makes, to a URL the runner supplied. It appears in no response, so none of the other assertion blocks can reach it, and a test that registers a configuration and stops has shown that the configuration API works and nothing about delivery. `expect_webhook` asserts on count, payload, delivery headers and the notification token, polling until `min_count` arrives or `timeout_ms` elapses. It carries no `expect` / `expect_error` / `expect_stream` / `capture` / `repeat`: those address a reply to a request the runner made, and this step makes none.
- **`expect_stream.streams` and `disconnect_after` (§7.3).** Some requirements are about the connections rather than the events on them: that an agent serves two subscribers to one task at once, and that it keeps serving one when the other goes away. Neither is observable from a single stream read to its end, however many assertions are made about the events it delivered. `streams` opens one stream per entry, concurrently, all from the step's single request — a runner MUST NOT rebuild the request per stream, or it is observing several different tasks. `disconnect_after: N` reads N events and MUST **actually close the connection**, not merely stop reading: an abandoned response is closed eventually by whatever reclaims it, by which time the stream has ended and the SUT has observed no mid-stream disconnect at all.
- **`preconditions.authentication`** — see below.
- **§12.2 runner-provided variables** — the corpus references `{{insufficientAuthToken}}`, `{{otherUserTaskId}}` and `{{webhookUrl}}`, which no document defined, so any third-party runner failed on unresolved variables. All three are now specified in a table tying each to its `runner_requirement`. The *valid* credential is deliberately not a variable: it would let an author write a raw step whose meaning depends on runner configuration rather than on the text of the test.

#### A declaration that cannot drift from the feature

§3.2 already let a test declare `runner_requirements`. Nothing made it do so, and the two drifted apart in the worst possible direction: three tests demanded a capability **in `description` only** — prose no runner reads — and every runner quietly executed something weaker and reported a pass.

So the requirement is now explicit and checkable. A test whose steps open more than one stream MUST declare `concurrent_streams`; one using `disconnect_after` MUST declare `stream_disconnect`; one using `expect_webhook` MUST declare `webhook_endpoint`; one naming a §12.2 variable MUST declare its requirement. A runner lacking any of them MUST skip.

§7.3 adds five narrower rules with the same purpose — stopping a deliberate cut from making a stream's assertions unfailable: `max_count` MUST NOT appear beside `disconnect_after` (the runner stops first, so the limit cannot be exceeded), `min_count` MUST NOT exceed it, `timeout_ms` bounds the step rather than one stream of it, and `capture` MUST NOT appear on a step opening several streams, since which stream a value came from would be decided by arrival order. A runner MUST report which stream a failure came from.

#### Clarifications with no grammar change

- **§4.4** — a raw request is sent exactly as written; the runner MUST NOT add the `A2A-Version` header or generate a `messageId`. A raw body carrying a `message` must therefore spell out its own.
- **§6.1** — `status` on gRPC is the canonical transcoding of the RPC status (`OK` → 200), which is what lets a status assertion stay binding-agnostic.
- **§7** — stream events may be addressed either through the `StreamResponse` discriminator or directly on the event body, and runners MUST accept both the `statusUpdate` and `status_update` spellings.
- **§10** — a `wire_payload` carrying a JSON-RPC envelope has its `id` replaced by the runner with the id the client actually issued; a fixed id can never be the one the client is waiting for.
- **§11.2** — `tck-auth-required` documented.

#### Mechanical conformance to the corrected grammar

- **13 × `code:` → `error_type:`** in `expect_error`. Every site held an error *name*, never a numeric code.
- **18 push-operation renames** — `set_push_notification_config` → 12 × `create_push_config`, plus `get`/`list`/`delete` to their short forms, which are the names §4.1's enum actually defines.
- **11 `messageId` additions** to raw bodies, now that §4.4 forbids the runner from generating one.
- **12 push-param flattenings** and **12 `taskId`/`contextId` relocations** into `message`, matching the corrected request shapes.
- **Nested `file:` parts flattened** to the proto's shape (`bytes` → `raw`, `name` → `filename`).

#### Semantic corrections

- **`REST-PD-001`** asserted an RFC 7807 problem-details body; A2A §11.6 mandates `google.rpc.Status`. It failed against every conformant implementation, including the reference SDKs.
- **`VER-NEG-002`** asserted a *successful* v1.0 response to a request with no `A2A-Version` header. A2A §3.6.2 makes an absent header mean 0.3, so a 1.0 interface must refuse it — now `-32009` or `-32601`.
- **`VER-NEG-001`** asserts `-32009`, per the corrected error table.
- **`CARD-CACHE-001`, `JSONRPC-CT-001`, `REST-CT-001`** were named for header checks but asserted something unrelated, because the grammar had no `headers` key. They now assert the header they are named for, and declare `runner_requirements: [header_inspection]` — the field's first use anywhere in the corpus.
- **`HTTP+JSON`** replaces `REST` as the binding name where the spec spells it that way.
- **`CORE-CTX-001`** rewritten as a raw step, being bimodal.
- **Four `AgentSkill` objects in `client-parsing`** gained the `description` and `tags` that `a2a.proto:441-443` marks REQUIRED. A §10 payload tests a client's correctness, so a card the spec forbids tests its tolerance instead: a strict parser rejects it before reaching the `capabilities` those tests assert on, and fails for a reason the test is not about.

#### Tests that passed without testing anything

Nine tests, in three families, described a requirement and asserted something weaker. All of them were green on all three transports against every reference SDK, which is how they survived.

- **Five push-delivery tests asserted nothing about delivery.** `PUSH-DELIV-001/002/003` and `SEC-PUSH-001/002` registered a push configuration, checked it came back, and stopped — a server that stores a config and pushes anonymously, or never pushes at all, passed every one. They now assert on the delivered request through `expect_webhook`. Two of them (REQ-SEC-018, REQ-SEC-019) check `Authorization`, because that is what A2A §4.3.3 makes a MUST: *"the agent MUST include authentication credentials in the request headers as specified in the `PushNotificationConfig.authentication` field"*. Deliberately **not** `X-A2A-Notification-Token`: A2A defines `PushNotificationConfig.token` but no header to carry it, so requiring one would fail a conformant agent over a convention the SDKs implement differently.
- **`webhookUrl` was a document literal** pointing at `https://example.com/...`, so every configuration named an address nobody answers. It is runner-provided by nature — the URL the runner listens on is not a property of the SUT — and moves to the §12.2 table.
- **`STREAM-MULTI-001`, `STREAM-MULTI-002` and `STREAM-RESUB-001`** demanded two concurrent streams and a mid-stream disconnect in `description` only. Their steps opened **one** undisturbed stream; `min_count: 2` meant two *events*, not two streams; `STREAM-MULTI-002` was byte-identical to `STREAM-MULTI-001` and closed nothing; `STREAM-RESUB-001` contained no disconnect, making it a duplicate of `STREAM-SUB-001`. All three now use `streams` / `disconnect_after` and declare their requirements. They also move onto a `tck-long-running` task, which is what gives the concurrency a subject — two `send_streaming_message` calls are two different tasks, so the "same events" the first is named for could never have been compared.
- **`CORE-ERR-009` was missing the half it is named for.** *"Unauthorized access returns same error as missing"* fetched two absent ids, so it asked only what a server does with an id nobody owns — which every server answers the same way, and which says nothing about disclosure. It now fetches `{{otherUserTaskId}}` alongside the absent id and requires both to name `TaskNotFoundError`.
- **Positive controls** added to `SEC-EXTCARD-001/002/004` and `REST-STATUS-001`. Each asserted only that a request is refused, so a server refusing that endpoint outright — or serving no such route at all — passed a MUST. Each now shows the same operation succeeding first. This is the SHOULD codified in §3.3 in the previous commit, applied to the tests that still lacked one.

#### Authentication

- **Five tests re-gated.** `SEC-AUTH-001/002/003/004/006` (4 must, 1 should) gated on `preconditions.capabilities.authentication` — and `AgentCapabilities` has exactly four members, none of them `authentication`. They were therefore **unrunnable against any agent on any binding**, and no SUT configuration could change that. They now use the new top-level `preconditions.authentication`, read off the card's `securitySchemes` **and** `securityRequirements`, presence only.
- **Why a precondition and not an unconditional assertion.** A2A conditions the server's obligation on *"its declared authentication requirements"*, so an agent declaring none is not violating anything by serving an unauthenticated request. The one exception is `Get Extended Agent Card`, which A2A mandates unconditionally — so `SEC-EXTCARD-001/002/004` deliberately do **not** carry the gate, and gating them would let an agent escape a MUST by declaring nothing.
- **Positive controls added** to `SEC-AUTH-001/002/004/006`. Each now proves the same operation on the same endpoint *succeeds* with a credential before asserting it is rejected without one. Without that, the tests pass against a server that refuses the request for any unrelated reason. Codified as a SHOULD in §3.3.
- **`SEC-AUTH-003` was vacuous** — both steps asserted only `message: {type: string}`, with no cross-step comparison, so any two error strings passed a MUST. Both steps now require `TaskNotFoundError`, which is the substance of the requirement: reporting an inaccessible task as *not found* is what prevents the disclosure, and naming the error after the authorization decision is itself the leak.
- **`SEC-AUTH-002`, `SEC-AUTH-003` and `SEC-EXTCARD-002` declare `runner_requirements: [auth_credentials]`**, being the tests that name a §12.2 variable.


tests/acts/auth-security.acts.yaml
```diff
@@ -33,30 +49,42 @@ suites:
...
       - id: SEC-AUTH-002
         name: "Server returns authorization error for insufficient permissions"
         description: "Runner MUST attempt a request with valid but insufficient credentials and verify rejection."
         spec_ref: "REQ-ERR-003"
         level: must
         preconditions:
-          capabilities:
-            authentication: true
+          authentication: true
         tags: [auth, runner-special]
```
aeoess, line 70:
```suggestion
        tags: [auth, runner-special]
        runner_requirements: [auth_credentials]
```

tests/acts/auth-security.acts.yaml
```diff
@@ -71,42 +99,49 @@ suites:
...
       - id: SEC-AUTH-003
         name: "Server does not reveal unauthorized resource existence"
         description: "Runner MUST verify that getting a task belonging to another user returns the same error as getting a nonexistent task."
         spec_ref: "REQ-ERR-004"
         level: must
         preconditions:
-          capabilities:
-            authentication: true
+          authentication: true
         tags: [auth, security, runner-special]
```
aeoess, line 121:
```suggestion
        tags: [auth, security, runner-special]
        runner_requirements: [auth_credentials]
```

aeoess (commented):
§12.2 now ties `insufficientAuthToken` and `otherUserTaskId` to the `auth_credentials` runner requirement, but the three tests that reference them do not declare it:

* `SEC-AUTH-002` → `{{insufficientAuthToken}}`
* `SEC-AUTH-003` → `{{otherUserTaskId}}`
* `SEC-EXTCARD-002` → `{{insufficientAuthToken}}`

Without the declaration, a runner that cannot supply the fixture reaches the unresolved-variable failure path instead of the skip behavior §12.2 defines.

The first two are in this diff, so I left suggestions there. `SEC-EXTCARD-002` is outside the changed lines and needs the same addition after its `tags` line:

```
runner_requirements: [auth_credentials]
```

tests/acts/auth-security.acts.yaml
JakubWorek: agree, added

tests/acts/auth-security.acts.yaml
JakubWorek: same as above

PR #2246, opened 2026-09-18
docs: pin the JCS nesting bound with a jcs_depth_v1 vector corpus

astrogilda:
`jcs_depth_v1` goes in beside the jcs_edge_v1 appendix from #2219. It holds 8 vectors: 4 accept at a bound of 128 open containers, 4 reject one level past it, 4 pair invariants.

jcs_edge_v1 pins the output a canonicalizer must produce, and this pins the input it must refuse, which RFC 8785 leaves open. So 2 implementations agreeing on its 10 vectors can still disagree over whether a deep artifact canonicalizes at all. Canonicalization runs before signature verification, and I read that as putting the recursive walk in reach of anyone holding an artifact.

That bound is the rule in-toto/attestation#570 proposes, in a pull request that has not merged. The 2 boundary cases come from agent-evidence-vectors, which runs with `uvx agent-evidence-vectors`, ahead of the 22 September TSC slot.


eriknewton:
Thanks, Sankalp. The empty-container boundary pair is a useful addition to the discussion. For the TSC, I would keep the distinction explicit: #2219 checks canonical output bytes; #2246 proposes an input-admission rule with a 128-container limit. Could the README label that limit as a proposed profile choice, rather than a requirement of RFC 8785 or an already-adopted A2A rule? The provenance file already makes the status of the in-toto proposal clear. It would also help to link the second serializer and its pinned run results so others can reproduce the independent cross-check. I haven’t run these vectors against Concordia yet.

astrogilda:
Erik, thank you for the label catch: the README now calls 128 a proposed profile choice, required by neither RFC 8785 nor A2A. The second-serializer line is withdrawn, because it named nothing a reader could repeat.

@kuangmi-bit, thank you for the run that replaced it, and for finding that jcs-depth-104's refusal came from the decoder's cap. The corpus does not depend on the number, so whichever bound the TSC settles, the counting rule from [agent-evidence-vectors](https://github.com/probityai/agent-evidence-vectors) carries over unchanged.

